PR #3, opened 2016-10-06
Adds docstrings everywhere

lukpueh:
A first iteration of the attempt of having useful and style guideline conformant docstrings

### DIFF
--- a/toto/log.py
+++ b/toto/log.py
@@ -1,19 +1,45 @@
+"""
+<Program Name>
+  log.py
+
+<Author>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+
+<Started>
+  Oct 4, 2016
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Simple wrapper of Python's logger module.
+
+Todo:
+  - Rethink log levels and function names
+  - Think about using logging formats and different logging scopes
+"""
+
 import logging
 from simple_settings import settings
 
 logging.basicConfig(level=settings.LOG_LEVEL, format='%(message)s')
 
 def doing(msg):
+  """Logs things that are currently being done """
   logging.info("DOING:   %s" % str(msg))
 
 def warning(msg):
+  """Logs things the user should be warned about """
   logging.warn("WARNING: %s" % str(msg))
 
 def passing(msg):
+  """Logs verification routines that pass """
   logging.critical("PASSING: %s" % str(msg))
 
 def failing(msg):
+  """Logs verification routines that fail """
   logging.critical("FAILING: %s" % str(msg))
 
 def error(msg):
+  """Logs program failures """
   logging.error("EXCEPTION: %s" % str(msg))
--- a/toto/log.py
+++ b/toto/log.py
@@ -13,10 +13,6 @@
 
 <Purpose>
   Simple wrapper of Python's logger module.
-
-Todo:
-  - Rethink log levels and function names
-  - Think about using logging formats and different logging scopes
 """
 
 import logging
--- a/toto/models/common.py
+++ b/toto/models/common.py
@@ -1,3 +1,31 @@
+"""
+<Program Name>
+  common.py
+
+<Author>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+  Santiago Torres <santiago@nyu.edu>
+
+<Started>
+  Sep 23, 2016
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Provides base classes for various classes in the model.
+
+<Classes>
+  Metablock:
+      pretty printed canonical JSON representation and dump
+
+  Signable:
+      sign self, store signature to self and verify signatures
+
+  ComparableHashDict: (helper class)
+      compare contained dictionary of hashes using "=", "!="
+"""
+
 import attr
 import canonicaljson
 
@@ -5,9 +33,9 @@ from ..ssl_crypto import keys as ssl_crypto__keys
 
 @attr.s(repr=False)
 class Metablock(object):
-  """ Objects with Base class Metablock have a __repr__ method
-  that returns a canonical pretty printed JSON string and also dumped to a
-  file. """
+  """Objects with base class Metablock have a __repr__ method
+  that returns a canonical pretty printed JSON string and can be dumped to a
+  file """
   def __repr__(self):
     return canonicaljson.encode_pretty_printed_json(attr.asdict(self))
 
@@ -15,12 +43,12 @@ class Metablock(object):
     with open(filename, 'wt') as fp:
       fp.write("{}".format(self))
 
+
 @attr.s(repr=False)
 class Signable(Metablock):
-  """ Objects of signable can sign themselves, i.e. their __repr__
-  without the signatures property, and store the signature to a signature
-  property.
-  """
+  """Objects with base class Signable can sign their payload (a canonical
+  pretty printed JSON string not containing the signatures attribute) and store
+  the signature (signature format: ssl_crypto__formats.SIGNATURE_SCHEMA) """
   signatures = attr.ib([])
 
   @property
@@ -30,8 +58,12 @@ class Signable(Metablock):
     return canonicaljson.encode_pretty_printed_json(payload)
 
   def sign(self, key):
-    """ Signs the canonical JSON repr of itself (without the signatures property)
-    and adds the signatures to its signature properties. """
+    """Signs the canonical JSON representation of itself (without the
+    signatures property) and adds the signatures to its signature properties.
+
+    Todo:
+      - Verify key format
+    """
 
     # XXX LP: Todo: Verify key format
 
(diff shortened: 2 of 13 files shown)


toto/models/common.py
```diff
@@ -30,21 +58,27 @@ def payload(self):
...
     passed key, and if the signature is valid.
 
-    Exceptions
-      Invalid key format Exception
-      Signature not found Exception
+    Todo:
+      - Verify key format
+      - raise InvalidKeyFormat
+      - raise SignatureNotFound
+      - raise WrongKeyForSignature (or similar)
```
SantiagoTorres: Have we ticketized these yet?

toto/models/layout.py
```diff
@@ -9,23 +43,63 @@
...
+  Todo:
+    - Warn if someone tries to add the private portion of a key to the keys
+
+  """
 
   _type = attr.ib("layout", init=False)
   steps = attr.ib([])
   inspect = attr.ib([])
   keys = attr.ib([])
   expires = attr.ib("")
```
SantiagoTorres: Maybe it's worth moving this to a datetime object with a default timeout (e.g., a month).

toto/models/layout.py
```diff
@@ -39,16 +113,42 @@ def read(data):
         keys=data.get("keys"), expires=data.get("expires"),
         signatures=data.get("signatures"))
 
+
 @attr.s(repr=False)
 class Step(models__common.Metablock):
+  """
+  Represents a step of the supply chain performed by functionary.
```
SantiagoTorres: by *a* functionary.

toto/models/layout.py
```diff
@@ -39,16 +113,42 @@ def read(data):
...
+    - Currently we store the list representation of matchrules to make
+    de-/serialization easier. But we actually want to store the objects.
 
+  """
   _type = attr.ib("step", init=False)
   name = attr.ib()
   material_matchrules = attr.ib([])
   product_matchrules = attr.ib([])
   pubkeys = attr.ib([])
   expected_command = attr.ib("")
```
SantiagoTorres: We are missing a `threshold` field as of the new specification.

toto/models/link.py
```diff
@@ -1,9 +1,66 @@
...
+          },... }
+
+    byproducts:
+        a dictionary in the format of
+          {
+            "stdout": <standard output of the executed command>
+            "stderr": <standard error of the executed command>
+          }
+
+    ran_command:
```
SantiagoTorres: I think this field should be just "command"

toto/models/link.py
```diff
@@ -14,16 +71,22 @@ class Link(models__common.Signable):
   return_value = attr.ib(None)
 
   def dump(self, filename=False):
+    """Write pretty printed JSON represented of self to a file with filename.
+    If no filename is specified, a filename will be created using the link name
+    + '.link'-suffix """
```
SantiagoTorres: we need to add the keyid prefix now.

toto/models/matchrule.py
```diff
@@ -1,18 +1,94 @@
...
 
   source_type = attr.ib("", init=False)
 
+
   @staticmethod
   def read(data):
-
     # XXX LP: needs some better checking
     # e.g. move checking to validator altogether
     if len(data) == 5 and data[0] == "MATCH" and data[1] == "MATERIAL":
```
SantiagoTorres: This *does* need better checking. Can we ticketize?

toto/models/matchrule.py
```diff
@@ -73,29 +165,33 @@ def verify_rule(self, item_link, step_links):
...
+  """Check if path is only in products and not in materials """
   path = attr.ib([])
 
+
   def __iter__(self):
     return iter(["CREATE", "{}".format(self.path)])
 
+
   def verify_rule(self, item_link, step_links):
-    """ materials and products are in the form of:
```
SantiagoTorres: Hmm, this doesn't seem to have a docstring anymore?

toto/models/validators.py
```diff
@@ -1,3 +1,9 @@
+"""
+Currently unused
+Needs revison!!
+Should be used to validate the format of objects
+"""
+
```
SantiagoTorres: If this is unused, can we take it out of the PR?

toto/runlib.py
```diff
@@ -17,21 +17,18 @@
...
-
-  TODO
-    * Decide on metadata location
-    * Properly duplicate stdout/stderr
-
+Todo:
+  - Decide on metadata location
+  - Properly duplicate stdout/stderr
+  - record_artifacts_as_dict needs revision
+  - Change key load
```
SantiagoTorres: Some of these could become tickets.

toto/runlib.py
```diff
@@ -169,9 +165,8 @@ def create_link_metadata(name, materials, products, byproducts,
 
 def run_link(name, materials, products, toto_cmd_args, record_byproducts=False):
   """Performs all actions associated with toto run-link.
-  XXX: This should probably be atomic, i.e. all or nothing"""
+  XXX LP: This should probably be atomic, i.e. all or nothing"""
```
SantiagoTorres: I personaly prefer the FIXME/TODO macros, as they are highlighted automatically by most IDE's. Although it's just a personal preference.

toto/verifylib.py
```diff
@@ -13,28 +13,39 @@
...
+    - Exception catching:
+      Currently failing verification steps raise exceptions, which are caught
+      here. Maybe we should pass on the exceptions and catch them in a program
+      closer to the user (e.g. in the commandline interface toto-verify)
+
+    - Same goes for the return value:
+      IMHO the functions of this library should just raise exceptions
+      and don't return a value (Python style)
+      If toto-verify wants to return something let it do so, but we shouldn't
+      track return value with a global variable here, we really shouldn't!
```
SantiagoTorres: These looks like tickets also...

toto/log.py
```diff
@@ -1,19 +1,45 @@
...
+  """Logs verification routines that pass """
   logging.critical("PASSING: %s" % str(msg))
 
 def failing(msg):
+  """Logs verification routines that fail """
   logging.critical("FAILING: %s" % str(msg))
 
 def error(msg):
+  """Logs program failures """
   logging.error("EXCEPTION: %s" % str(msg))
```
SantiagoTorres: I think it'd be cool if we chose a standard method for string formatting. I've seen many people prefer `"{}".format` over `"%s" % ()`, but I don't know.

We should read up on [this](https://www.python.org/dev/peps/pep-3101/) and make a decision. Maybe it's also worth propagating this over the SSL coding guideline for python.

toto/models/layout.py
lukpueh: Will be fixed with https://github.com/toto-framework/Toto/issues/9

toto/models/layout.py
lukpueh: Will be fixed with https://github.com/toto-framework/Toto/issues/7

toto/models/matchrule.py
lukpueh: Done. See issues 5 - 18

toto/runlib.py
lukpueh: Good thought. I must have adapted this from the `Seattle` codebase. Let's discuss this with regard to our [style guide lines](https://github.com/secure-systems-lab/code-style-guidelines).

toto/log.py
lukpueh: Choosing one way is a good idea. I personally prefer `"%s" % ()` because you find similar format specifier in other languages too. But then again, I'm not overly passionate about keeping it. :)

lukpueh:
@SantiagoTorres, thanks for your comments. I addressed them in 640699f which mostly consisted of writing GitHub issues and removing the todos from the docstrings.

toto/log.py
SantiagoTorres: I think we should prefer format over %, as I've seen many popular projects prefering it. Also, from the docs:

> *Note*: The formatting operations described here exhibit a variety of quirks that lead to a number of common errors (such as failing to display tuples and dictionaries correctly). Using the newer str.format() interface helps avoid these errors, and also provides a generally more powerful, flexible and extensible approach to formatting text.

SantiagoTorres:
Besides the string formatting syntax, I think this is ready to merge 👍

JustinCappos:
By the way, quite a bit of this goes against the lab's code style
guidelines.  If there are things you think we should add / revise, we
should discuss it.

lukpueh:
@JustinCappos  Okay, I thought I had most guidelines covered, but I'll re-check
I guess you probably had in mind:
- *"Comments should be complete sentences"* (although most of the comment samples in the guidelines don't strictly align with this)
- *"...should end in a period"* (will do)
- *"two spaces after a sentence-ending period"* (can do)
- *"every function in a module that will be called from other modules... purpose, arguments, exceptions, side effects, and return value"* (will do)

lukpueh:
As I gathered from a conversation with @JustinCappos his code-style comment mostly addresses the *object oriented programming* aspects of my code.
Since this is not really related to docstrings I would go with Santiago and merge now.

Note:
- I added a [comment to an existing issue](https://github.com/in-toto/in-toto/issues/14#issuecomment-256464348), where we can continue the *OO* discussion.
- I'd like to mention that docstrings are not finished everywhere with this PR, but since `in-toto` is heavy WIP - I already made a bunch of changes + PR's based on this branch that also include new docstrings -  I'd suggest to merge now and keep above comments in mind for further docstrings

toto/log.py
SantiagoTorres: @lukpueh ping here. What do we do about this?

SantiagoTorres:
Thanks! 👍

PR #4, opened 2016-10-10
Project owner tool

SantiagoTorres:
this adds the first iteration of the project owner interactive tool. This is a side-project that I mostly intended for project owners that want to edit/create their layouts. It's still in rough shape, but we can start reviewing the code. It does need some cleanup (I'm mostly pep8 compliant), as there are some unnecessary modules.

Here's a screencast about the basic idea:
[asciicast](https://asciinema.org/a/dveulv7rigy7hswjtmlbb6gdk)

### DIFF
--- a/toto/designlib/__init__.py
+++ b/toto/designlib/__init__.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python
 """
-    <name> 
+    <name>
         designlib
 
     <description>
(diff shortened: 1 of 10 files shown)


lukpueh:
This looks really cool and should be very helpful!

SantiagoTorres:
I think it's time to close this. I'll subtree it out and close it. It's been a while since...

Issue #13, opened 2016-10-17
Use validators for models and format/schema for crypto


lukpueh:
We share crypto specific libraries with [the update framework](https://github.com/theupdateframework/tuf) and started off toto using TUF's format/schema validation. As we are more on the OO side than TUF, we will want to replace format/schema validation with model validators, e.g. using [`attr`'s validator](https://attrs.readthedocs.io/en/stable/api.html#validators) module.
- [ ] implement validators (the downside of `attr.validate` is that it seems to be hooked to the `__init__`, which is not ideal if we want to set attributes post-instantiation and then validate)
- [x] remove `toto/formats.py`
- [x] keep using and `toto/ssl_crypto/formats.py` for crypto specific data structures, and add where appropriate

lukpueh:
Layout model validator should prevent from adding private key portions to the `keys` field.

Issue #12, opened 2016-10-17
PKI helper instead of throw-away util


lukpueh:
Currently we provide a throw-away `create_and_persist_or_load_key` function which served its purpose during development. 
In future we definitely want something a little more sophisticated, although PKI is not part of the toto specification.
- Add pki helper interface to create and store public/private keypairs (offer standardized format, eg. PEM) 
- Change key loading calls throughout the codebase to load keys in various formats or die
- Disable the possibility of adding the private portion of a key to the toto layout

lukpueh:
Fixed in https://github.com/in-toto/in-toto/commit/f25721b16b155db1cedd9c02eaadd674b6d53b3f

Issue #6, opened 2016-10-17
User input, output, exceptions and return values


lukpueh:
Meaningful `exceptions` are preferred over zero vs. non-zero return values in `verifylib` and `runlib` (this is the pythonic way). 
The command line tools `toto-run` and `toto-verify` on the other hand could catch those exceptions, print meaningful output to stderr and return zero or non-zero values.

lukpueh:
Replace comma based `catch Exception, e:` with explicit name binding syntax `catch Exception as e` as [recommended in PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations).

lukpueh:
@SantiagoTorres suggested to raise a custom class based exception if the caller doesn't pass an `iterable` to `import_rsa_public_keys_from_files_as_dict`. Let's keep this in mind when tackling this issue.

lukpueh:
@vladimir-v-diaz suggested to catch `ImportError` and take out the `e` in [`import subprocess`](https://github.com/in-toto/in-toto/blob/develop/toto/runlib.py#L36).

lukpueh:
Let's rethink user feedback in general when addressing this issue. Some Exception messages should be give more context.

lukpueh:
Let me summarize my thoughts on user feedback and give some architectural context:
The current in-toto architecture has the following __components__:
 - *models*
Python classes that represent in-toto specific data structures (Layout, Link, ...), plus mostly methods for schema validation and de-/serialization 
 - *libraries*
Provide in-toto logic, work on objects defined in the model (run in-toto step, create link metadata with recorded artifacts, verify layout, ...)
- *command line tools*
User-facing tools to run complete in-toto workflows (toto-run, in-toto-record, toto-verify) using the libraries

This is what I __expect__ from those components in terms of user feedback.
- *exceptions and return values*
Library functions and model methods raise defined custom Python exceptions if something goes wrong ([EAFP](https://docs.python.org/2/glossary.html#term-eafp)).
Command line tools handle exceptions and provide meaningful user feedback and return values (no stack traces) ([LBYL](https://docs.python.org/2/glossary.html#term-lbyl)).
- *verbosity*
Callers might want command line tools as well as library functions to provide informative user feedback (e.g. "Doing X...", "Creating Y...", or the like). Command line tools should provide a `-v, --verbose` flag to turn on/off such feedback. Calls to the library should default to no output but could provide a programmatic way of enabling/disabling it.
[...]

PR #19, opened 2016-10-18
Adds PKI functions to toto.utils

lukpueh:
Fixes most of https://github.com/in-toto/in-toto/issues/12
- Adds proper functions to create, encrypt, store to and read from file, and decrypt keys. Keys use RSA and PEM format 
- Adds util wrappers for above functions that additionally prompt the user for a password to decrypt the keys
- Adapts toto-run/toto-verify and demo to use these functions

*Additional notes:*
- above functions are similar to the key functions in [TUF's repository tool](https://github.com/theupdateframework/tuf/tree/develop/tuf#keys). Since TUF wants to keep them in the `repository_tool` module and we don't want to import that (it is TUF specific) we have to replicate the functionality here
- *don't allow private keys in layout (https://github.com/in-toto/in-toto/issues/12)*  will be fixed with https://github.com/in-toto/in-toto/issues/13
- better user feedback will be given with https://github.com/in-toto/in-toto/issues/6

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -25,10 +25,10 @@ def main():
   """
   # Create keys
   print "Generate keypair for Alice..."
-  toto.util.prompt_generate_and_write_rsa_keypair("alice")
+  toto.util.generate_and_write_rsa_keypair("alice")
 
   print "Generate keypair for Bob..."
-  toto.util.prompt_generate_and_write_rsa_keypair("bob")
+  toto.util.generate_and_write_rsa_keypair("bob")
 
   alice_public = toto.util.import_rsa_key_from_file("alice.pub")
   bob_public = toto.util.import_rsa_key_from_file("bob.pub")
@@ -74,7 +74,7 @@ def main():
 
   # Sign and dump layout
   print "Load alice private key to sign layout..."
-  alice_private = toto.util.prompt_import_rsa_key_from_file("alice")
+  alice_private = toto.util.import_rsa_key_from_file("alice")
   layout.sign(alice_private)
   layout.dump()
 
--- a/demo.py
+++ b/demo.py
@@ -23,24 +23,29 @@ def main():
   # Define the supply chain
   #############################################################################
   """
+  # Create keys
+  print "Generate keypair for Alice..."
+  toto.util.prompt_generate_and_write_rsa_keypair("alice")
 
-  # Get keys
-  alice_key = toto.util.create_and_persist_or_load_key("alice")
-  bob_key = toto.util.create_and_persist_or_load_key("bob")
+  print "Generate keypair for Bob..."
+  toto.util.prompt_generate_and_write_rsa_keypair("bob")
+
+  alice_public = toto.util.import_rsa_key_from_file("alice.pub")
+  bob_public = toto.util.import_rsa_key_from_file("bob.pub")
 
   # Create Layout
-  layout = m.Layout.read({ 
+  layout = m.Layout.read({
     "_type": "layout",
     "expires": "EXPIRES",
     "keys": {
-        alice_key["keyid"]: alice_key,
-        bob_key["keyid"]: bob_key
+        alice_public["keyid"]: alice_public,
+        bob_public["keyid"]: bob_public
     },
     "steps": [{
         "name": "write-code",
         "material_matchrules": [],
         "product_matchrules": [["CREATE", "foo.py"]],
-        "pubkeys": [alice_key["keyid"]],
+        "pubkeys": [alice_public["keyid"]],
         "expected_command": "vi",
       },
       {
@@ -51,7 +56,7 @@ def main():
         "product_matchrules": [
             ["CREATE", "foo.tar.gz"],
         ],
-        "pubkeys": [bob_key["keyid"]],
+        "pubkeys": [bob_public["keyid"]],
         "expected_command": "tar zcvf foo.tar.gz foo.py",
       }],
     "inspect": [{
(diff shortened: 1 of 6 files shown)


demo.py
```diff
@@ -23,24 +23,29 @@ def main():
   # Define the supply chain
   #############################################################################
   """
+  # Create keys
+  print "Generate keypair for Alice..."
```
SantiagoTorres: I understand this is demo code, so I'm not completely worried about this, but we should use the `print()` function rather than the `print` statement.

demo.py
```diff
@@ -51,7 +56,7 @@ def main():
         "product_matchrules": [
             ["CREATE", "foo.tar.gz"],
         ],
-        "pubkeys": [bob_key["keyid"]],
+        "pubkeys": [bob_public["keyid"]],
```
SantiagoTorres: Wouldn't it be worth using our models to do this instead? so we don't have to perform these little changes every time.

toto/util.py
```diff
@@ -1,31 +1,108 @@
-"""
-    TODO: markup
-"""
```
SantiagoTorres: Is this missing docstrings? or are they part of #3?

toto/util.py
```diff
@@ -1,31 +1,108 @@
...
+
+  public = rsa_key['keyval']['public']
+  private = rsa_key['keyval']['private']
+  private_pem = toto.ssl_crypto.keys.create_rsa_encrypted_pem(private, password)
+
+  with open(filepath + '.pub', 'w') as fo_public, \
+      open(filepath, 'w') as fo_private:
+
+    fo_public.write(public.encode('utf-8'))
+    fo_private.write(private_pem.encode('utf-8'))
```
SantiagoTorres: Is there a reason we need to do these operations in the same context?

toto/util.py
```diff
@@ -1,31 +1,108 @@
...
+  """
+
+  toto.ssl_crypto.formats.PATH_SCHEMA.check_match(filepath)
+  toto.ssl_crypto.formats.PASSWORD_SCHEMA.check_match(password)
+
+  rsa_key = toto.ssl_crypto.keys.generate_rsa_key()
+
+  public = rsa_key['keyval']['public']
+  private = rsa_key['keyval']['private']
+  private_pem = toto.ssl_crypto.keys.create_rsa_encrypted_pem(private, password)
```
SantiagoTorres: we may not want to have an encrypted pem every time. Why don't we default password=None and then act accordingly?

demo.py
```diff
@@ -68,7 +73,9 @@ def main():
   })
 
   # Sign and dump layout
-  layout.sign(alice_key)
+  print "Load alice private key to sign layout..."
+  alice_private = toto.util.prompt_import_rsa_key_from_file("alice")
+  layout.sign(alice_private)
```
SantiagoTorres: I see that you made this change, but nowhere in layout.py you changed to use the PEM (or ed25519) instead of the key dictionary... is this expected?

SantiagoTorres: The same happens with link. I wonder if you forgot to add...

demo.py
vladimir-v-diaz: You might need to include `from __future__ import print_function` at the top of the module to
recognize the print() function.

>
Note
This function is not normally available as a built-in since the name print is recognized as the print statement. To disable the statement and use the print() function, use this future statement at the top of your module:
from __future__ import print_function

https://docs.python.org/2/library/functions.html#print

demo.py
lukpueh: Internally we keep using the key dictionary. I just changed the way how keys are stored. Therefor no changes in the models are required.

toto/util.py
lukpueh: Indeed, this is missing docstrings. Maybe I should have omitted the *everywhere* in the PR's title.

toto/util.py
lukpueh: Valid point. I will fix this.

toto/util.py
lukpueh: Hm. I think I misused the `with` statement. Actually I intended something like a transaction, i.e. write both files or none. But this is not what `with` does.
I can definitely use two `with` blocks.

lukpueh:
Addressed requested changes:
- separate `with` file contexts - 	b2ae929
- make RSA key passwords optional for generating keypair and loading keys 	25828f9

demo.py
lukpueh: I created an issue to address python2/3 compatibility https://github.com/in-toto/in-toto/issues/23.

SantiagoTorres: @vladimir-v-diaz Actually, the `from __future__...` statement works the other way around:

```
santiago at ~/.../programas/git ✔ python2
Python 2.7.12 (default, Jun 28 2016, 08:31:05) 
[GCC 6.1.1 20160602] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> print "kek"
kek
>>> print("kek")
kek
>>> from __future__ import print_function
>>> print("kek")
kek
>>> print "kek"
  File "<stdin>", line 1
    print "kek"
              ^
SyntaxError: invalid syntax
>>> 
```
Regardless of that technicallity, I think it's a good idea to add the include statement so we don't have forward-incompatible code.

demo.py
SantiagoTorres: Instead of writing the layout itself, instantiate Link and Inspection objects to add to a Layout object and then dump it. Maybe it's a little bit more tedious in the short term, but should be less "volatile" (and easier to update if something breaks).

I don't think this is anything major though.

lukpueh: Okay. Don't worry about it. `demo.py` won't exist for much longer.

SantiagoTorres:
Just rebased, verified and approved 👍

SantiagoTorres:
merged via 3697e8b19a08ffb1537835f606a5cfedbeaa7244

Issue #15, opened 2016-10-17
Miscellaneous material/product (artifact) enhancements 


lukpueh:
- Support file path wildcards for artifact flags in `toto-run` and consequently in `toto-verify`
- Re-think "./" left-stripping of paths when recording artifacts (maybe this could be done in verification instead?)
- revise `record_artifacts_as_dict` function 

Later:
- think of a more sophisticated way of recording artifacts than by passing file paths as command line arguments (this might deserve a separate ticket)

lukpueh:
Replace comma based `catch Exception, e:` with explicit name binding syntax `catch Exception as e` as [recommended in PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations).

lukpueh:
Fixes for above items can be found in:
- Support pattern wildcards - https://github.com/in-toto/in-toto/pull/37, https://github.com/in-toto/in-toto/pull/70
- Re-think `'./'`- left-stripping - https://github.com/in-toto/in-toto/pull/56/commits/c0624605147b85cb1709355a5178a53dbee66f00
- revise record_artifact_as_dict - https://github.com/in-toto/in-toto/pull/56 (excludes), https://github.com/in-toto/in-toto/pull/58 (base path)
- more sophisticated way of recording artifacts (will open ticket when required)

Issue #8, opened 2016-10-17
"Relaxed" command alignment


lukpueh:
Command alignment verification passes if the step command as defined in the layout and the related command as recorded by the link metadata are completely equal. If only the first `x` parameters are equal (where `x` is the minimum count of both commands) `toto-verify` should output a warning.

lukpueh:
Fixed in https://github.com/in-toto/in-toto/pull/22/commits/c6db4fee4aa86d146769e1b8f3d625d105a9294f will merge with https://github.com/in-toto/in-toto/pull/22.

PR #22, opened 2016-10-24
Libify runlib and verifylib - fixes #5

### LINKED ISSUE #5: libraryfy runlib and verifylib
- [x] `verifylib.py` :Currently the verifylib is implemented as one long function. We should separate the single verification steps into function units that are called from the commandline tool `toto-verify.py`.
  Also, functions in verifylib should not contain log/debug/info statements but rather raise meaningful exceptions that might be caught in `toto-verify.py`
- [x] `runlib.py`
  Runlib is already split into separate functions but there is one  "call-them-all" function in runlib which should be moved to `toto-run.py`. Same as above applies for log statements here.

lukpueh:
- Changes runlib and verifylib to provide logically separated units of functionality that are called from the command line interfaces toto-run and toto-verify respectively.
- The libraries, in case something goes wrong, solely raise exceptions, whereas the user interfaces catch those exceptions and give proper user feedback (stdout/stderr and return values).
- Enhances the docstrings above all in the interface functions

Other changes
- Adds utility function to load publickeys from files into a keydict
- Adds Layout method to load link metadata files into a linkdict 
- Changes verify_signature function to only pass if all signatures verify
- Changes toto-run.py --name option to --step-name

### DIFF
--- a/toto/models/common.py
+++ b/toto/models/common.py
@@ -66,19 +66,20 @@ class Signable(Metablock):
     signature = ssl_crypto__keys.create_signature(key, self.payload)
     self.signatures.append(signature)
 
-  def verify_signature(self, key):
-    """Verifies if the object contains a signature matching the keyid of the
-    passed key, and if the signature is valid."""
+  def verify_signatures(self, keys_dict):
+    """Verifies all signatures of the object using the passed key_dict."""
 
-    # XXX LP: Todo: Verify key format
+    if not self.signatures or len(self.signatures) <= 0:
+      raise Exception("No signatures found")
 
     for signature in self.signatures:
-      if key["keyid"] == signature["keyid"]:
-        return ssl_crypto__keys.verify_signature(key, signature,
-            self.payload)
-    else:
-      # XXX LP: Replace exception with KeyNotFound exception (or return false?)
-      raise Exception("Signature with keyid not found")
+      keyid = signature["keyid"]
+      try:
+        key = keys_dict[keyid]
+      except:
+        raise Exception("Signature key not found, key id is %s" % keyid)
+      if not ssl_crypto__keys.verify_signature(key, signature, self.payload):
+        raise Exception("Invalid signature")
 
 
 @attr.s(repr=False, cmp=False)
--- a/toto/models/layout.py
+++ b/toto/models/layout.py
@@ -35,7 +35,7 @@ import attr
 # import validators
 from . import common as models__common
 from . import matchrule as models__matchrule
-
+from . import link as models__link
 @attr.s(repr=False)
 class Layout(models__common.Signable):
   """
@@ -104,6 +104,31 @@ class Layout(models__common.Signable):
         keys=data.get("keys"), expires=data.get("expires"),
         signatures=data.get("signatures"))
 
+  def import_step_metadata_from_files_as_dict(self):
+    """
+    <Purpose>
+      Iteratively loads link metadata files for each Step of the Layout
+      from disk and returns a dict with Link names as keys and Link objects
+      as values.
+
+    <Arguments>
+      None
+
+    <Exceptions>
+      TBA (see https://github.com/in-toto/in-toto/issues/6)
+
+    <Side Effects>
+      Calls functions to read files from disk
+
+    <Returns>
+      A dictionary with Link names as keys and Link objects as values.
+
+    """
+    step_link_dict = {}
+    for step in self.steps:
+      link = models__link.Link.read_from_file(step.name + '.link')
+      step_link_dict[step.name] = link
+    return step_link_dict
 
 @attr.s(repr=False)
 class Step(models__common.Metablock):
--- a/toto/runlib.py
+++ b/toto/runlib.py
@@ -27,8 +27,6 @@
 import sys
 import os
 import tempfile
-import logging
-
 
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
 # install and use the much more recent subprocess32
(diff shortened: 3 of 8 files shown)


lukpueh:
Also fixes "relaxed" command alignment  https://github.com/in-toto/in-toto/issues/8

lukpueh:
Rebased and force pushed onto tip of branch key-utils 25828f933cec0146150ad368202ba4793cb7bf54

toto/models/common.py
```diff
@@ -66,19 +66,20 @@ def sign(self, key):
...
-        return ssl_crypto__keys.verify_signature(key, signature,
-            self.payload)
-    else:
-      # XXX LP: Replace exception with KeyNotFound exception (or return false?)
-      raise Exception("Signature with keyid not found")
+      keyid = signature["keyid"]
+      try:
+        key = keys_dict[keyid]
+      except:
+        raise Exception("Signature key not found, key id is %s" % keyid)
```
SantiagoTorres, line 80: We are still not raising "KeyNotFound", can we leave the FIXME or open a ticket about it?

toto/models/layout.py
```diff
@@ -104,6 +104,15 @@ def read(data):
         keys=data.get("keys"), expires=data.get("expires"),
         signatures=data.get("signatures"))
 
+  def import_step_metadata_from_files_as_dict(self):
```
SantiagoTorres, line 107: Is this function going to be used by other modules? If so, maybe we need a more elaborate docstring. Otherwise, we may want to prepend this function name with a _ to hint that it's a private function.

toto/runlib.py
```diff
@@ -35,53 +35,63 @@
...
+  with ssl_crypto.formats.HASHDICT_SCHEMA. """
+  toto.ssl_crypto.formats.HASHALGORITHMS_SCHEMA.check_match(hash_algorithms)
+  hash_dict = {}
 
-def record_artifacts_as_dict(artifacts):
-  """Takes a list of directories and/or filenames as input and creates a dict
-  with filepaths as keys and their files' hashes as values.
+  for algorithm in hash_algorithms:
+    digest_object = toto.ssl_crypto.hash.digest_filename(filepath, algorithm)
+    hash_dict.update({algorithm: digest_object.hexdigest()})
```
SantiagoTorres, line 58: Is there a reason we use  `update` and not vanilla `hash_dict[algorithm]=digest_object.hexdigest()`?

toto/runlib.py
```diff
@@ -35,53 +35,63 @@
...
 
-  if not artifacts:
-    return artifacts_dict
 
-  def _hash_artifact(filepath, hash_algorithms=['sha256']):
-    """Takes a filename and hashes the respective file's contents
-    using the passed hash_algorithms. Returns a HASHDICT"""
+def _normalize_path(path):
+  """Internal helper that strips "./" on the left side of the path. """
+  if path.startswith("./"):
```
SantiagoTorres, line 67: This is rather platform specific. What would happen if os.path.sep is "\" (e.g., windows)?

toto/toto-run.py
```diff
@@ -19,87 +19,125 @@
...
+
+def in_toto_run(step_name, key_path, material_list, product_list,
+    link_cmd_args, record_byproducts=False):
+  """Load link signing private keys from disk and runs passed command, storing
+  its materials, by-products and return value, and products into link metadata
+  file. The link metadata file is signed and stored to disk. """
+  try:
+    log.doing("load link signing key...")
+    key = toto.util.prompt_import_rsa_key_from_file(key_path)
+  except Exception, e:
```
SantiagoTorres, line 51: I think the `Exception as e` construct is more forward-compatible than this tuple. We may want to verify this.

toto/toto-run.py
```diff
@@ -19,87 +19,125 @@
...
-      usage="python -m %s --name <unique name>\n" \
-            "            [--materials <filepath>[,<filepath> ...]]\n" \
-            "             --products <filepath>[,<filepath> ...]\n" \
-            "             --key <filepath>\n" \
-            "             --record-byproducts\n" \
-            "             -- <cmd> [args]" % (os.path.basename(__file__), ))
-
-  # Option group for toto specific options, e.g. material and product
-  toto_args = parser.add_argument_group("Toto options")
+      usage="./toto-verify.py --step-name <unique step name>\n" \
```
SantiagoTorres: I think we should drop the "./" part of it. Even better, we could say"

```
usage="{} --step-name <unique step name\n".format(sys.argv[0]) 
```

But this is mere pedantry.

toto/toto-run.py
```diff
@@ -19,87 +19,125 @@
...
+            "                [--materials <filepath>[,<filepath> ...]]\n" \
+            "                [--products <filepath>[,<filepath> ...]]\n" \
+            "                 --key <functionary private key path>\n" \
+            "                [--record-byproducts]\n" \
```
SantiagoTorres: Maybe it's better to enclose multi-line strings using parentheses, instead of `\`. I personally prefer it because of the reasons explained [here](https://stackoverflow.com/questions/3346230/wrap-long-lines-in-python/18160132#18160132). I don't think this is a blocker though.

toto/toto-verify.py
```diff
@@ -20,7 +20,7 @@
 
   Example Usage:
   ```
-  python -m toto.toto-verify --layout <root.layout> --layout-key <layout-key>
+  ./toto-verify.py --layout <root.layout> --layout-keys <layout-key>
```
SantiagoTorres: Same comment as above.

toto/toto-verify.py
```diff
@@ -29,29 +29,88 @@
...
+    help="Key(s) to verify root layout signature (separated ',')")
 
   args = parser.parse_args()
 
-  layout_key = toto.util.prompt_import_rsa_key_from_file(args.layout_key)
-
-  retval = toto.verifylib.toto_verify(args.layout, layout_key)
-  sys.exit(retval)
+  layout_key_paths = args.layout_keys.split(',')
+  in_toto_verify(args.layout, layout_key_paths)
```
SantiagoTorres, line 119: Hmm, I'm troubled here. Is there a reason `in_toto_verify` is not part of the library? I know we may have talked about this before, but I'm starting to wonder if we should expose this function to everyone willing to use verifylib.

toto/util.py
```diff
@@ -97,6 +97,15 @@ def import_rsa_key_from_file(filepath, password=None):
...
 
+def import_rsa_public_keys_from_files_as_dict(filepaths):
+  """Takes a list of filepaths to RSA public keys and returns them as a
+  dictionary conformant with ssl_crypto.formats.KEYDICT_SCHEMA."""
+  key_dict = {}
+  for filepath in filepaths:
+    key = import_rsa_key_from_file(filepath)
+    keyid = key['keyid']
+    key_dict[keyid] = key
+  return key_dict
```
SantiagoTorres, line 108: It may be a good idea to add statements to check whether the arguments of this function are sane (e.g., am I even getting a list?!)

toto/models/common.py
lukpueh: There is already a ticket for this - https://github.com/in-toto/in-toto/issues/6

toto/models/layout.py
lukpueh: Fixed in e5bb78e

toto/runlib.py
lukpueh: I admit it, I'm guilty of copy and paste. Maybe we should ask @vladimir-v-diaz [about this](https://github.com/theupdateframework/tuf/blame/develop/tuf/util.py#L410-L412)?

toto/runlib.py
lukpueh: Good point. I actually don't know how `os.walk(".")` works on windows, but on *NIX it prepends `./`.
There is already a ticket for this issue (see https://github.com/in-toto/in-toto/issues/15).

toto/runlib.py
SantiagoTorres: `git`. `blames`. fired. :gun:

toto/toto-run.py
lukpueh: You are right. Pep8 has a [recommendation](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) about this.
I added a comment to https://github.com/in-toto/in-toto/issues/6#issuecomment-256982234 where we should address this.

toto/toto-run.py
lukpueh: I used `sys.argv[0]` before. But I thought it would be nicer to align the subsequent lines which was easier*\* with having the file name hardcoded.

*\* I know, I could have used `len` and `format` with some padding option

lukpueh: Do you want me to change this here?

SantiagoTorres: I ran a couple of examples locally (firefox, nvim, gcc, etc). Most of them just use the name of the binary sans the `./`, even if they are aliased (i.e., they don't take argv[0] into account).

Maybe the easiest would be to just use `"toto-verify..py --step-name..."`

toto/toto-verify.py
lukpueh: Yes there is. `in_toto_verify` gives a lot of user feedback (catches exceptions and prints their messages, takes care of the return value) which IMHO should not happen in the library.

lukpueh: Same applies to `in_toto_run`, which even prompts for user input (when loading password encrypted keys) which also should not happen in the library.

SantiagoTorres: Ah, so that's why. Then disregard this thread. 👍

toto/util.py
lukpueh: Hm. I'm not so sure about it. If the the functions gets the wrong arguments it should raise an exception anyway. Maybe we want  something else than a `'TypeError' ... not iterable`?

toto/toto-run.py
lukpueh: Wanna create a ticket? :)

toto/toto-run.py
lukpueh: I guess this should be `toto-verify.py ...`?

toto/util.py
SantiagoTorres: >  Maybe we want something else than a 'TypeError' ... not iterable?

This is exactly what I mean. Is it a list, or an iterable? As of now, this function works with anything that implements `__iter__`, including a string. I don't know if this is intended, or even if it's a big deal, but it's something we should consider. We are offloading the sanity of our arguments to the for loop in there and `import_rsa_key_from_file`.

lukpueh: Yep. I think it is fine to just pipe through the exceptions from `import_rsa_key_from_file`. It will complain if the caller passes an `iterable` that does not contain file paths to keys.
We can use  a custom class based exception that is raised if the caller doesn't pass an `iterable`. But I'd prefer doing this here: https://github.com/in-toto/in-toto/issues/6#issuecomment-256983523

toto/runlib.py
vladimir-v-diaz: Sorry, I can't remember/figure out why I used update().  I might have been using update() elsewhere at the time and used update() here out of habit.

I think `hash_dict[algorithms]=digest_object.hexdigest()` works just fine here.

vladimir-v-diaz: @SantiagoTorres Did you update that image of a gun?  I swear it was a realistic one before and not a toy gun.

SantiagoTorres: Lol, I think it depends on your computer/font.

PR #24, opened 2016-10-27
Adds default expiration and verification (fix #7)

### LINKED ISSUE #7: Verify expirationdate
@SantiagoTorres [proposed to use a `datetime`](https://github.com/toto-framework/Toto/pull/3#discussion_r83093316) object for the layout's `expires` field with a default timeout (e.g. a month). This is a good idea. Also we should check this upon `toto-verify`.

lukpueh:
Fixes https://github.com/in-toto/in-toto/issues/7
- If no expiration date is specified a default expiration
  -- one month from now in UTC+00:00 (Zulu time) --
  is set in the Layout's read method
- Adds layout expiration verification method in verifylib
- Calls layout expiration verification in in-toto-verify
- Removes "EXPIRES" string from demo layout (creates a default now)
- Adds new third party requirements: python-dateutil, iso8601

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -36,7 +36,7 @@ def main():
   # Create Layout
   layout = m.Layout.read({
     "_type": "layout",
-    "expires": "EXPIRES",
+    "expires": "",
     "keys": {
         alice_public["keyid"]: alice_public,
         bob_public["keyid"]: bob_public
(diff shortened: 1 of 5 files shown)


SantiagoTorres:
Thanks!

PR #27, opened 2016-10-28
Adds subprocess fallback for POSIX users

lukpueh:
Since we only encourage POSIX systems to use subprocess32 we add a fallback to subprocess if the import fails and output a warning.

### DIFF
--- a/toto/runlib.py
+++ b/toto/runlib.py
@@ -27,11 +27,18 @@
 import sys
 import os
 import tempfile
+import logging
+
 
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
 # install and use the much more recent subprocess32
 if os.name == 'posix' and sys.version_info[0] < 3:
-  import subprocess32 as subprocess
+  try:
+    import subprocess32 as subprocess
+  except Exception, e:
+    logging.warning("POSIX users (Linux, BSD, etc.) are strongly encouraged to"
+        + " install and use the much more recent subprocess32")
+    import subprocess
 else:
   import subprocess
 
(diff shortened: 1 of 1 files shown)


toto/runlib.py
```diff
@@ -27,11 +27,18 @@
...
 
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
 # install and use the much more recent subprocess32
 if os.name == 'posix' and sys.version_info[0] < 3:
-  import subprocess32 as subprocess
+  try:
+    import subprocess32 as subprocess
+  except Exception, e:
+    logging.warning("POSIX users (Linux, BSD, etc.) are strongly encouraged to"
+        + " install and use the much more recent subprocess32")
```
SantiagoTorres: Take the + out please

SantiagoTorres (changes requested):

SantiagoTorres (approved):

toto/runlib.py
```diff
@@ -27,11 +27,18 @@
...
+import logging
+
 
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
 # install and use the much more recent subprocess32
 if os.name == 'posix' and sys.version_info[0] < 3:
-  import subprocess32 as subprocess
+  try:
+    import subprocess32 as subprocess
+  except Exception, e:
```
vladimir-v-diaz, line 38: Should you check for the more specific `ImportError` exception here?

``` Python
>>> import subprocess32
ImportError: No module named subprocess32
```

You are still using the older `Exception, e` convention.  However, in this case, I'd recommend to drop the `e` since it's not used.

``` Python
try:
  import subproess32 as subprocess
except ImportError:
  import subprocess
```

lukpueh: Good idea. Will do in https://github.com/in-toto/in-toto/issues/6#issuecomment-257575261.

PR #29, opened 2016-10-31
Drop fixmes

SantiagoTorres:
@lukpueh I rebased develop and dropped the FIXME commit, for some reason this is the diff that I see. Any ideas of what may have caused this? 

This seems like what we want though...


lukpueh:
Fixed.

PR #31, opened 2016-11-02
Removes OO form matchrules, updates AS PATH

### LINKED ISSUE #14: Store matchrules as Matchrule objects and not as lists
Currently we store the list representation of `Matchrule` objects in the according `Step` and `Inspection` fields. This is a workaround that allows us to use  `attr`'s convenient `asdict` function when dumping the layout to a file. 

I suggest to [implement our own `asdict` function](https://github.com/hynek/attrs/issues/103#event-825440411) allowing us to store `Matchrules` as objects but returning their list representation instead of  their dict representation when dumping the container layout.

lukpueh:
- Matchrules are no longer OO
  they are simply stored as list and all the verification happens
  in verifylib
- This also fixes #14, an ugly workaround to use attr's asdict
- The commit also updates the MATCH matchrules to take an
  optional target path which allows for matching artifacts that
  have different paths


toto/verifylib.py
```diff
@@ -251,47 +253,321 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
-  verify them."""
+def verify_match_rule(rule, source_type, source_link, target_links):
+  """
+  <Purpose>
+    Verifies that the source artifact (depending on the list the rule was
+    extracted from a material or product) hash matches the target artifact
+    (depending on the 2nd element of the rule a material or product).
+
+    Also verifies:
+    That the target_link as idetfied by ("FROM" <step>) exists in
```
SantiagoTorres: identified*

toto/verifylib.py
```diff
@@ -251,47 +253,321 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+            A dictionary of Link objects with Link names as keys.
+            The Link objects relate to Steps.
+            The contained materials and products are used as verification target.
+
+  <Exceptions>
+    raises an Exception if the rule is not conformant with the rule format.
+    raises an if a matchrule does not verify.
+    TBA (see https://github.com/in-toto/in-toto/issues/6)
+
+    RuleVerficationFailed
```
SantiagoTorres: When this is raised would be good saying here.

toto/verifylib.py
```diff
@@ -251,47 +253,321 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+  # Verify that the target Link metadata object contains the material or product
+  # to match with.
+  if (target_path not in target_artifacts.keys()):
+    raise RuleVerficationFailed("'%s' not in target %ss" \
+        % (target_path, target_type))
+
+  # Verify that the recorded source artifact hash and the recorded target
+  # artifact hash are equal.
+  if (ComparableHashDict(source_artifacts[source_path]) != \
+      ComparableHashDict(target_artifacts[target_path])):
```
SantiagoTorres, line 357: Using these `ComparableHashDicts` is the way to go? couldn't we just compare fields within the source/target paths?

I wonder if this is an issue if, for example, one link provides two hashes and the other one provides only one.

This is a minor nit though.

toto/verifylib.py
```diff
@@ -251,47 +253,321 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+    link:
+            The Link object for the Item (Step or Inspection) that contains
+            the rule.
+
+  <Exceptions>
+    raises an Exception if the rule is not conformant with the rule format.
+    raises an if a matchrule does not verify.
+    TBA (see https://github.com/in-toto/in-toto/issues/6)
+
+    RuleVerficationFailed
```
SantiagoTorres: same here

toto/verifylib.py
```diff
@@ -251,47 +253,321 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+
+  <Side Effects>
+    None.
+
+  """
+  # FIXME: Validate rule format
+  path = rule[1]
+  if (path in link.materials.keys()):
+    raise RuleVerficationFailed("'%s' " \
```
SantiagoTorres: I don't think you need the `\` here

toto/verifylib.py
```diff
@@ -251,47 +253,321 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+            The Link objects relate to Steps.
+            The contained materials and products are used as verification target.
+
+  <Exceptions>
+    raises an Exception if a rule is not conformant with the rule format.
+    raises an Exception if a matchrule does not verify.
+    TBA (see https://github.com/in-toto/in-toto/issues/6)
+
+  <Side Effects>
+    Calls the respective verify_<rule-type>_rule function.
```
SantiagoTorres: Is this really a side-effect?

toto/verifylib.py
```diff
@@ -251,47 +253,321 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+            The Link objects relate to Steps.
+            The contained materials and products are used as verification target.
 
   <Exceptions>
     raises an Exception if a matchrule does not verify.
     TBA (see https://github.com/in-toto/in-toto/issues/6)
 
   <Side Effects>
-    Calls matchrule verification methods.
+    Calls helper functions to verify all matchrules of an item.
```
SantiagoTorres: Is this a side-effect?

SantiagoTorres (changes requested):
Some minor nits only. Thanks!

toto/verifylib.py
lukpueh: Okay, will give more info.

toto/verifylib.py
lukpueh: `ComparableHashDicts` is just a helper I wrote to compare two hash dicts of format `ssl_crypto.formats.HASHDICT_SCHEMA`, e.g.

``` python
{'sha256': '23432df87ab..',
'sha512': '34324abc34df..', ...}
```

The behavior is documented in its docstring:

``` python
    """Equal if the dicts have the same keys and the according values
    (strings) are equal"""
```

But we can talk about changing the behavior. Using it seems handy though.

toto/verifylib.py
lukpueh: Will change.

toto/verifylib.py
lukpueh: Will change.

toto/verifylib.py
lukpueh: You are right. I don't think it is a side-effect either. I guess I saw some similar "side effects" in the tuf projects and adapted the broad understanding. I'll take this out here.

toto/verifylib.py
lukpueh: See https://github.com/in-toto/in-toto/pull/31#discussion_r86243521

toto/verifylib.py
SantiagoTorres: Yeah, this was my bad. I think the current behavior is ok for the time being.

SantiagoTorres (approved):

SantiagoTorres:
👍 Thanks!

PR #32, opened 2016-11-04
ENH:toto-run:handle multiple products and materials

SantiagoTorres:
This fixes the argparse calls so that the --materials and --products
flags can recieve multiple paths as arguments without the need of quotes
or commas.

### DIFF
--- a/toto/toto-run.py
+++ b/toto/toto-run.py
@@ -93,8 +93,8 @@ def main():
   parser = argparse.ArgumentParser(
       description="Executes link command and records metadata",
       usage="toto-verify.py --step-name <unique step name>\n" +
-            "                     [--materials <filepath>[,<filepath> ...]]\n" +
-            "                     [--products <filepath>[,<filepath> ...]]\n" +
+            "                     [--materials <filepath>[ <filepath> ...]]\n" +
+            "                     [--products <filepath>[ <filepath> ...]]\n" +
             "                      --key <functionary private key path>\n" +
             "                     [--record-byproducts]\n" +
             "                      -- <cmd> [args]")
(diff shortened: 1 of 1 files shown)


toto/toto-run.py
```diff
@@ -93,8 +93,8 @@ def main():
   parser = argparse.ArgumentParser(
       description="Executes link command and records metadata",
       usage="toto-verify.py --step-name <unique step name>\n" +
```
lukpueh: Although it's not part of this PR. This should be `toto-run.py` and not `toto-verify.py`. Do you want to fix this here, or do we need an extra PR for this, or should we just hot push to develop?

toto/toto-run.py
```diff
@@ -132,9 +132,9 @@ def main():
   product_list = []
 
   if args.materials:
-    material_list = args.materials.split(",")
+    material_list = args.materials
   if args.products:
-    product_list = args.products.split(",")
+    product_list = args.products
```
lukpueh: Any reason not to add `args.materials` and `args.products` directly to the function call in [`in_toto_run`](https://github.com/in-toto/in-toto/pull/32/files#diff-8e16d631a0ef00bfa92de5d2ad6c1f65R139)?

lukpueh (changes requested):

toto/toto-run.py
```diff
@@ -108,9 +108,9 @@ def main():
 
   # XXX LP: We should allow path wildcards here and sanitze them
```
lukpueh: I guess you can delete this comment now.

toto/toto-run.py
SantiagoTorres: Not at all, I wanted to not pollute the diff (so that if we need to add any validation to it we donn't have problems). I could do so if you think this is better.

toto/toto-run.py
SantiagoTorres: Will do.

toto/toto-run.py
lukpueh: Hm. I'm fine with you polluting the diff.

SantiagoTorres:
@lukpueh Ready for another round 👍

lukpueh (approved):

PR #33, opened 2016-11-04
ENH:runlib:warn if the artifact to record does not exist

SantiagoTorres:
This is useful to notify users that the intended artifacts do not exist
(or they were a typo).

### DIFF
--- a/toto/runlib.py
+++ b/toto/runlib.py
@@ -100,6 +100,11 @@ def record_artifacts_as_dict(artifacts):
     return artifacts_dict
 
   for artifact in artifacts:
+
+    if not os.path.exists(artifact):
+      log.warning("path: {} does not exist, skipping..".format(artifact))
+      continue 
+
     if os.path.isfile(artifact):
       artifacts_dict[_normalize_path(artifact)] = _hash_artifact(artifact)
     elif os.path.isdir(artifact):


toto/runlib.py
```diff
@@ -100,6 +100,11 @@ def record_artifacts_as_dict(artifacts):
     return artifacts_dict
 
   for artifact in artifacts:
+
+    if not os.path.exists(artifact):
+      log.warning("path: {} does not exist, skipping..".format(artifact))
+      continue 
+
```
lukpueh, line 107: This is a good idea. Shouldn't it also work to have this as an `else` clause [below](https://github.com/in-toto/in-toto/pull/33/files#diff-c8942631fdeb6716abe5f5b4611c90cfR115)? Like so:

```
if is file:
  record
elif is directory:
 record using walk (flattens the tree)
else:
  warn
```

SantiagoTorres: Hmm as an addition? So this would be like the following:

```
if is file:
     record
elif isdict:
     record using walk
elif:
     record 
     log.warn("Attempted to record unknown filetype {}".format(artifact))
```

Or you're thinking of dropping the os.path.exists() altogether?

lukpueh: I was thinking of dropping `os.path.exists()`. It's either a file or a directory or does not exist, right?

SantiagoTorres: I thiknk os.path.exists() can work with nonregular files

SantiagoTorres: e.g., a link, or a pipe

lukpueh: Do you mean [`os.path.lexists()`](https://docs.python.org/2/library/os.path.html#os.path.lexists)?

lukpueh: Do we want to record pipes and links as artifacts?

lukpueh: Btw. [`isfile`](https://docs.python.org/2/library/os.path.html#os.path.isfile) does follow links.

SantiagoTorres: I don't mean os.path.lexists, which is pretty much the same in many cases.

I don't know if we want to record them as artifacts, but I guess we would want to have a behavior defined for it.

`isfile` does follow links, but that's just one case of special file.

SantiagoTorres: I do believe that having separate checks will let us, at least, print appropriate errors for each case.

lukpueh: So what about something like:

```
if islink:
elif isfile:
elif isdir:
....
```

?

SantiagoTorres: You seem to be set on not having the `os.path.exists()`. Is there any particular reason as to why?

lukpueh: IMHO it's a redundant condition. But I'm not over passionate. If you prefer it this way, let's merge.

SantiagoTorres: My reasoning is that the message in case of error is different:

```
if islink:
  ....
elif isfile:
  ...
elif isdir:
  ...
else: #error
   log.error("This is either an unsupported file type or the file does not exist!")
```

Whereas the redundant conditions give you more granularity:

``` python
if not os.path.exists(...):
   log.warning("Could not find file...!")
...
if islink:
...
else:
   log.warning("This file type is not supported by in-toto!")
```

lukpueh: (y)

PR #37, opened 2016-11-09
Matchrules: path expansion + verification changes + unittests

lukpueh:
- Enables the usage of shell-style path expansion in matchrules. See Python's [`fnmatch`](https://docs.python.org/2/library/fnmatch.html) module for allowed patterns.
- Matchrules are now verified in two directions:
  - Every recorded artifact (product or material) related to an item (Step or Inspection)  must be explicitly matched by a rule - **authorize artifacts**
  - Every artifact required by a matchrule must have been recored - **enforce artifacts**
- Adds unit tests for all matchrule functions.
- Makes third argument `verify_all_item_rules` optional (see commit message of https://github.com/in-toto/in-toto/commit/213a4c5187179b31d5fb6960673f2fb425001d11).
- Slightly modifies demo script to align with matchrule verification changes. 
- Raises  `Exception("...not implemented")` when verifying a "MODIFY" rule (specs need revision here).

### DIFF
--- a/demo/run_demo.py
+++ b/demo/run_demo.py
@@ -111,7 +111,7 @@ def main():
   parser = argparse.ArgumentParser()
   parser.add_argument("-n", "--no-prompt", help="No prompt.",
       action="store_true")
-  parser.add_argument("-c", "--clear", help="Clear created files.",
+  parser.add_argument("-c", "--clean", help="Remove files created during demo.",
       action="store_true")
   args = parser.parse_args()
 
--- a/demo/run_demo.py
+++ b/demo/run_demo.py
@@ -38,7 +38,7 @@ def supply_chain():
   prompt_key("Package (Carl)")
   os.chdir("../functionary_carl")
   package_cmd = ("python -m toto.toto-run" +
-                 " --step-name package --material foo.py" +
+                 " --step-name package --materials foo.py" +
                  " --products foo.tar.gz" +
                  " --key carl --record-byproducts" +
                  " -- tar zcvf foo.tar.gz foo.py")
@@ -79,7 +79,7 @@ def supply_chain():
   prompt_key("Package (Carl)")
   os.chdir("../functionary_carl")
   package_cmd = ("python -m toto.toto-run" +
-                 " --step-name package --material foo.py" +
+                 " --step-name package --materials foo.py" +
                  " --products foo.tar.gz" +
                  " --key carl --record-byproducts" +
                  " -- tar zcvf foo.tar.gz foo.py")
@@ -101,6 +101,7 @@ def supply_chain():
   verify_cmd = ("python -m toto.toto-verify" +
                 " --layout root.layout" +
                 " --layout-key alice.pub")
+
   print verify_cmd
   retval = subprocess.call(verify_cmd.split())
   print "Return value: " + str(retval)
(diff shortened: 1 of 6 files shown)


demo/run_demo.py
```diff
@@ -101,20 +101,46 @@ def supply_chain():
...
   retval = subprocess.call(verify_cmd.split())
   print "Return value: " + str(retval)
 
 
 def main():
   parser = argparse.ArgumentParser()
-  parser.add_argument("-n", "--no-prompt", help="No prompt",
+  parser.add_argument("-n", "--no-prompt", help="No prompt.",
+      action="store_true")
+  parser.add_argument("-c", "--clear", help="Clear created files.",
```
SantiagoTorres: should we rename this to _clean_ instead?

test/test_verifylib.py
```diff
@@ -0,0 +1,502 @@
...
+  def test_pass_ignore_case_keyword(self):
+    """["delete", "bar"], ["DELETE", "bar"], ignores keyword case, passes. """
+
+    rule1 = ["delete", "bar"]
+    rule2 = ["DELETE", "bar"]
+    self.assertIsNone(
+        verify_delete_rule(rule1, self.artifact_queue))
+    self.assertIsNone(
+        verify_delete_rule(rule2, self.artifact_queue))
+
```
SantiagoTorres, line 78: I think we can remove some of the whitespace in here.

toto/verifylib.py
```diff
@@ -303,261 +315,304 @@ def verify_match_rule(rule, source_type, source_link, target_links):
...
 
+#   if (path not in item_link.products.keys()):
+#     raise RuleVerficationFailed("'%s' "
+#         "not found in products of link '%s'" % (path, link.name))
 
-def _verify_rules(rules, source_type, source_link, target_links):
+#   if (ComparableHashDict(link.materials[path]) == \
+#       ComparableHashDict(link.products[path])):
+#     raise RuleVerficationFailed("hashes of product and material '%s' of link "
+#         "'%s' match - should have been modified" % (path, link.name))
```
SantiagoTorres: Could we take this code out of this commit? so we can have a cleaner diff when this function is added.

toto/verifylib.py
```diff
@@ -303,261 +315,304 @@ def verify_match_rule(rule, source_type, source_link, target_links):
...
+    Each rule will be applied on the artifacts currently in the queue, that is
+    if artifacts were already matched by a previous rule in the list they
+    cannot be matched again.
+
+    This can lead to ambiguity in case of conflicting rules, e.g. given a step
+    with a reported artifact "foo" and a rule list
+    [["CREATE", "foo"], ["DELETE", "foo"]].
+    In this case the verification would pass, because
+    verify_create_rule would remove the artifact from the artifact queue, which
+    would make "foo" appear as deleted for verify_delete_rule.
```
SantiagoTorres, line 490: 👍

SantiagoTorres (changes requested):
Really good! I just need 1 change and maybe two style issues. 

Thanks!

demo/run_demo.py
lukpueh: Sounds sensible

test/test_verifylib.py
lukpueh: This is actually the first time that I adhere to the [blank line style guide line](https://github.com/secure-systems-lab/code-style-guidelines#blank-lines).  :)
So far I didn't use this particular guide line because I thought it unnecessarily spreads up the code.
But in this case - many very short functions with one liner doc strings - I think it makes the code look less cluttered.

Anyway, you are probably right that we should do it either everywhere or nowhere.

SantiagoTorres (approved):

SantiagoTorres:
Thanks! 👍

PR #39, opened 2016-11-11
Update matchrules

lukpueh:
Adds modify rule verification back to verifylib. The rule is now called NOTMATCH and has one of the following formats.
```python
["NOTMATCH", "MATERIAL", "<source and target path pattern>", "FROM", "<step name>"]
["NOTMATCH", "PRODUCT", "<source and target path pattern>", "FROM", "<step name>"]
["NOTMATCH", "MATERIAL", "<source path pattern>", "AS", "<target path pattern>", "FROM", "<step name>"]
["NOTMATCH", "PRODUCT", "<source path pattern>", "AS", "<target path pattern>", "FROM", "<step name>"]
```
The verification itself has changed as well. Now it verifies that no file in the artifact queue matched by source path pattern has a related file in the target artifacts matched by the target artifact pattern with the same hash.

### DIFF
--- a/test/test_verifylib.py
+++ b/test/test_verifylib.py
@@ -22,7 +22,8 @@ import unittest
 from toto.models.link import Link
 from toto.models.layout import Step, Inspection
 from toto.verifylib import verify_delete_rule, verify_create_rule, \
-    verify_match_rule, verify_item_rules, verify_all_item_rules
+    verify_match_rule, verify_item_rules, verify_all_item_rules, \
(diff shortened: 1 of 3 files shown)


lukpueh:
Let's not merge this until we have sorted out _all_ open questions regarding our current matchrule specification.

lukpueh:
I think we don't have to drag along this obsolete PR any further. Let's keep the branch though, in case we decide to provide a NOTMATCH rule at some point.

PR #40, opened 2016-11-11
MAINT:verifylib, toto-verify: libify in_toto_verify

SantiagoTorres:
Integrators of Toto may want to directly access a "seamless"
verification routine to avoid any misunderstandings on errors upon
implementation. Move "holistic" in_toto_verify function to verifylib so
integrators can use it.

### DIFF
--- a/toto/toto-verify.py
+++ b/toto/toto-verify.py
@@ -29,79 +29,6 @@ import argparse
 import toto.util
 import toto.verifylib
 import toto.log as log
-from toto.models.layout import Layout
-
-def _die(msg, exitcode=1):
-  log.failing(msg)
-  sys.exit(exitcode)
-
-def in_toto_verify(layout_path, layout_key_paths):
-  """Loads layout file and layout keys from disk and performs all in-toto
-  verifications."""
-
-  try:
-    log.doing("load layout...")
-    layout = Layout.read_from_file(layout_path)
-  except Exception, e:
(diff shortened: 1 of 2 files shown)


lukpueh:
Actually I'd rather have you not touch `toto/toto-verify.py` in this case. I think it makes sense to keep this user (cli) facing function in the cli tool and have an equivalent function in toto_verify without the Exception catchers and the log statements. If a verification part fails when calling the library function directly, it will get an exception with the appropriate message and return value.

SantiagoTorres:
The problem with doing this is that we have the same code in two places, which will cause problems down the line.

I think these "interface issues" could be solved better if we namespace the logger instead.

lukpueh:
@SantiagoTorres, I spent some thought on your comments to this discussion and wrote them up in https://github.com/in-toto/in-toto/issues/6#issuecomment-263988714. I think this approach will resolve *all* issues. :) Let me know what you think.

SantiagoTorres:
This feels as if it's addressed by #60 . Should we close it, @lukpueh ?

lukpueh:
Actually, it will be addressed by #61 (libify-in-toto-verify), which I plan to submit later today. #60 is libify-in-toto-run. But yes, we can close, or close once I've submitted.

lukpueh:
Moved to https://github.com/in-toto/in-toto/pull/62. Please review and merge over there. Thanks!

PR #42, opened 2016-11-15
FIX:.gitmodules: use https instead of ssh

SantiagoTorres:
This will make the demo code work even if users do not have a github
account. It also gets rid of the 301 that may be confusing some people

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "toto/ssl_crypto"]
 	path = toto/ssl_crypto
-	url = git@github.com:/santiagotorres/ssl_crypto
+	url = https://github.com/santiagotorres/ssl_crypto


lukpueh:
We are currently using https://github.com/vladimir-v-diaz/ssl_crypto/commit/542caedfc2c1d60f6d30ae1188aa68ad7cb7a389 for in-toto, which does not exist in [secure-systems-lab/ssl_crypto](https://github.com/secure-systems-lab/ssl_crypto).

SantiagoTorres:
hmm, why is that not part of the secure-systems-lab version of the repo?

vladimir-v-diaz:
This commit was part of a different branch (optional-password?) that wasn't
moved over to the secure-systems-lab organization.  I told Lukas that I
would work on adding this to
https://github.com/secure-systems-lab/ssl_crypto

## 

vladimir.v.diaz@gmail.com

## PGP fingerprint = ACCF 9DCA 73B9 862F 93C5  6608 63F8 90AA 1D25 3935

SantiagoTorres:
Hmm, so what's the plan?

vladimir-v-diaz:
Whatever is in the [optional-password](https://github.com/vladimir-v-diaz/ssl_crypto/tree/optional-password) branch will be merged with SSL's [master](https://github.com/secure-systems-lab/ssl_crypto/tree/master).

SantiagoTorres:
@vladimir-v-diaz could we do this sometime soon? I think merging this will make the student's lifes easier.

vladimir-v-diaz:
I have a meeting with one of the AppSec groups at 1.  I will work on this
after the meeting.  There are some merge conflicts and I need to propagate
changes to TUF.

## 

vladimir.v.diaz@gmail.com

## PGP fingerprint = ACCF 9DCA 73B9 862F 93C5  6608 63F8 90AA 1D25 3935

lukpueh:
We used to use [ssl_crypto@just-crypto](https://github.com/vladimir-v-diaz/ssl_crypto/tree/just-crypto) and then [ssl_crypto@optional-password](https://github.com/vladimir-v-diaz/ssl_crypto/tree/optional-password).
Both branches are not yet merged into master.

vladimir-v-diaz:
We renamed the `just-crypto` branch to `master` when we moved `ssl_commons` and `ssl_crypto` to the `secure-systems-lab` organization.  After the branch rename I had to rename tufformats.py (that was added to TUF) to formats.py and fix conflicts.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "toto/ssl_crypto"]
 	path = toto/ssl_crypto
-	url = git@github.com:/santiagotorres/ssl_crypto
+	url = https://github.com/secure-systems-lab/ssl_crypto
```
lukpueh: For now we should just update the protocol. Once we have done all the necessary merges and tested the new additions of `secure-systems-lab/ssl_crypto` we can also change the address.

SantiagoTorres: I guess I can do this and we can figure this out later

SantiagoTorres:
@lukpueh, @vladimir-v-diaz I force-pushed. Let me know if this is what we want.

vladimir-v-diaz:
TUF's `.gitmodules` uses https: https://github.com/theupdateframework/tuf/blob/develop/.gitmodules

I didn't make this decision, but it seems to work.  I use SSH when I git clone TUF.

SantiagoTorres:
@vladimir-v-diaz yeah, that's why we are moving to https. People that aren't too familiar with ssh can't seem to make it work.

vladimir-v-diaz:
The `optional-password` branch has been integrated with `master`:
https://github.com/secure-systems-lab/ssl_crypto/pull/1
https://github.com/secure-systems-lab/ssl_crypto/commit/45f5bddc6fcc56b6de83f92b93a9374ea56c1ba9
https://github.com/secure-systems-lab/ssl_crypto/commit/ff19a508023cec859a3e2bc3a897030075743792

I also successfully propagated the changes to TUF, but encountered one problem and left a [comment](https://github.com/secure-systems-lab/ssl_crypto/blob/master/tests/test_pycrypto_keys.py#L31-L44) in the affected modules.

vladimir-v-diaz:
Note: You cannot run tests from within `ssl_crypto/tests`, rather they must be executed from the root directory with `python -m unittest discover`.

`python -m tuf.ssl_crypto.tests.test_pycrypto_keys` appears to work, too.

vladimir-v-diaz:
What doesn't seem to work on my set up:

``` Bash
$ python tuf/ssl_crypto/tests/test_pycrypto_keys.py
  File "tuf/ssl_crypto/tests/test_pycrypto_keys.py", line 36, in <module>
    from ...ssl_commons import exceptions as ssl_commons_exceptions
ValueError: Attempted relative import in non-package
```

``` Bash
$ cd tuf/ssl_crypto/tests/
$ python test_pycrypto_keys.py
  File "test_pycrypto_keys.py", line 36, in <module>
    from ...ssl_commons import exceptions as ssl_commons_exceptions
ValueError: Attempted relative import in non-package
```

lukpueh:
I added an inline comment regarding relative imports here:
 https://github.com/secure-systems-lab/ssl_crypto/commit/ff19a508023cec859a3e2bc3a897030075743792#commitcomment-19850954

SantiagoTorres:
ping @lukpueh

lukpueh:
(y)

PR #47, opened 2016-11-18
Change command alignment verification to pass or warn

lukpueh:
- ~~Adds 3 custom class based exception and modifies functions in `verifiylib` and in `common.Signable` to use them.~~
- ~~Replaces use of `log` with a `print` statement for a user warning~~

This PR used to be `Add verifylib exceptions` but the focus changed during review so it was renamed. This PR now:
- changes command alignment verification (see https://github.com/in-toto/in-toto/pull/47/commits/d91479397964958bcfd357c68b6d08e713e3d283 for reasoning) + unittests
- Adds and uses custom `SignatureVerificationError`
- Adds and uses custom `LayoutExpiredError`


toto/exceptions.py
```diff
@@ -1,5 +1,15 @@
 from toto.ssl_commons.exceptions import Error
 
+class SignatureVerificationError(Error):
+  """Indicates a signature verification Error. """
+
+class LayoutExpiredError(Error):
+  """Indicates that the layout expired. """
+  pass
```
SantiagoTorres, line 9: Some have pass, some others don't. Is that intentional?

toto/models/common.py
```diff
@@ -29,6 +29,7 @@
 import attr
 import canonicaljson
 
+from ..exceptions import SignatureVerificationError
```
SantiagoTorres: Does this need to be a relative import? or can we import toto.exceptions?

toto/verifylib.py
```diff
@@ -213,11 +213,15 @@ def verify_command_alignment(command, expected_command):
 
   for i in range(min(command_len, expected_command_len)):
     if command[i] != expected_command[i]:
-      raise Exception("Command '%s' and expected command '%s' do not align" \
+      raise CommandAlignmentFailed(
+          "Command '%s' and expected command '%s' do not align" \
```
SantiagoTorres: shoudln't we be moving to `format`?

SantiagoTorres: Also, I'm not sure if you need the line splice since this is within parentheses.

toto/verifylib.py
```diff
@@ -213,11 +213,15 @@ def verify_command_alignment(command, expected_command):
...
           % (command, expected_command))
   if command_len != expected_command_len:
-    log.warning("Command '%s' and expected command '%s' do not fully align" \
-        % (command, expected_command))
+    # FIXME we don't want print statements in a library
+    # The warning will be obsolete with a fix for
+    # https://github.com/in-toto/in-toto/issues/46
+    print "Command '%s' and expected command '%s' do not fully align" \
```
SantiagoTorres: can we add parentheses to it so we don't break when we  hit python 3?

lukpueh:
Thanks @SantiagoTorres. Addressed your suggestions in 8c33184 and 9247f33.

toto/verifylib.py
```diff
@@ -213,11 +213,15 @@ def verify_command_alignment(command, expected_command):
...
+      raise CommandAlignmentFailed(
+          "Command '{0}' and expected command '{1}' do not align"
+          .format(command, expected_command))
   if command_len != expected_command_len:
-    log.warning("Command '%s' and expected command '%s' do not fully align" \
-        % (command, expected_command))
+    # FIXME we don't want print statements in a library
+    # The warning will be obsolete with a fix for
+    # https://github.com/in-toto/in-toto/issues/46
+    print("Command '{0}' and expected command '{1}' do not fully align"
```
SantiagoTorres: I'm curious, why don't we use the logger instance as it was done before? 
`log.warning` sounded like a good idea.

lukpueh: Hm. I talked to @vladimir-v-diaz about it. We agreed that the `logging` module should not be used to give direct feedback to the user, but rather to *log* errors/warnings/infos to some place (could be a file) intended for the developer/debugger.

vladimir-v-diaz: We figured that if in-toto is going to be used as a library, the calling application probably doesn't want to see messages from the library printed/logged to the console (at least not by default).  For applications that would like to have these log statements show up on the console, perhaps an option can be added for them?

SantiagoTorres: Ok, then it would make sense to add exceptions to pr #40 and let the user-facing tool handle them?

lukpueh: So I re-thought all of this and summarized my conclusions in https://github.com/in-toto/in-toto/issues/6#issuecomment-263988714. Nevertheless, this still does not solve this particular issue. If we raise an exception instead of a warning here, the exception could be caught and translated to a warning in the user tool, but this would also abort the the layout verification, which we don't want.

lukpueh: Solutions:
 - return a list of "pending command verification exceptions"  that can be raised, caught and logged once the entire verification has finished
 - just log the warning in the library

lukpueh:
This PR is starting to diverge from its original intent. The most recent commit drastically changes command alignment verification and adds unittests. The changes were triggered by a discussion over the introduced `CommandAlignmentFailed` Exception, which eventually was removed again in the most recent commit.

SantiagoTorres:
Yeah, so maybe we should close it and start over once we decide what's the best path forward?

lukpueh:
I don't think that's necessary. We could:
 - Rename the PR
 - Update the PR's description
 - (rename the branch?)

Issue #46, opened 2016-11-18
Expected command pattern instead of strict and relaxed alignment


lukpueh:
Currently we pass command verification upon strict (must be equal) and relaxed (must have an equal prefix) alignment of recorded command and expected command. The following examples show in what cases the alignment passes or fails.

```python
# Left command is the command recorded by toto-run.py
# Right command is the expected command defined in the layout

# Passes:
["vi", "file1", "file2"] ~ ["vi", "file1", "file2"]

# Passes with warning:
["vi", "file1", "file2"] ~  ["vi"]
["make", "install"] ~  ["make"]

# Fails:
["vi", "file1", "file2"] ~ ["make", "install"]
["/usr/bin/vi", "file1"] ~ []
["vi", "file1", "file2"] ~ ["vi", "file2", "file1"]
["/usr/bin/vi", "file1"] ~ ["vi", "file1"]
```
As one can see in above examples that semantically equal commands sometimes fail whereas syntactically  (relaxed)  equal commands can pass although they have very different semantics.

@vladimir-v-diaz brought up the idea of using match patterns. We could use [regex](https://docs.python.org/2/library/re.html) or a subset of regex like implemented by [fnmatch](https://docs.python.org/2/library/fnmatch.html). This would allow either fail or pass and get rid of the relaxed pass.

In any case command alignment does not provide strong security guarantees, since recorded commands can be easily faked using alias, symlinks or the like.

JustinCappos:
I think that just recording what was run and having the client software
eventually warn (error?) is perhaps the best policy.  Getting too advanced
with this probably isn't worth the time.  So, simply doing a strict check
may be best.  Yes, we know someone could bypass this, but it prevents many
types of accidental 'foot-shooting'...  :)

lukpueh:
Sounds good to me.

PR #48, opened 2016-11-20
13 use validators for the models

### LINKED ISSUE #13: Use validators for models and format/schema for crypto

SantiagoTorres:
This patch series refactors code in the models and adds the validate function to ensure the properties of a model are populated properly. I also added unit tests for these methods.

closes #13


lukpueh:
cdc03038575eb7742dcb21ec5fea62fc955ccf2b should not be part of this PR

lukpueh:
I think we agreed to [avoid trailing whitespaces](https://github.com/secure-systems-lab/code-style-guidelines/issues/5). Mind stripping those?

toto/models/common.py
```diff
@@ -43,6 +44,14 @@ def dump(self, filename):
     with open(filename, 'wt') as fp:
       fp.write("{}".format(self))
 
+  def validate(self):
+    """ Aggregates all the validate methods of itself and its subclasses """
```
lukpueh: This could use a little more information, à la *purpose, arguments, exceptions, side effects, and return value*. Especially since it those *esoteric* stuff like class inspection. :)

SantiagoTorres:
Fixed

toto/models/layout.py
```diff
@@ -136,6 +143,62 @@ def import_step_metadata_from_files_as_dict(self):
       step_link_dict[step.name] = link
     return step_link_dict
 
+  def _validate_type(self):
+    """ ensure that the type string is set to layout"""
```
lukpueh: I think your oneliner docstrings should look like this, according to our code style guidelines:
```python
"""Ensure that the type string is set to layout. """
```
Please change everywhere in this PR.

toto/models/layout.py
```diff
@@ -136,6 +143,62 @@ def import_step_metadata_from_files_as_dict(self):
...
+    if self._type != "layout":
+      raise FormatError("Invalid _type value for layout (Should be 'layout')")
+
+    return True
+
+  def _validate_expires(self):
+    """ ensure that the expiration field is a proper timestamp"""
+    try:
+      date = parse(self.expires)
```
lukpueh, line 155: This should be more restrictive. The required schema is [ISO8601_DATETIME_SCHEMA](https://github.com/secure-systems-lab/ssl_crypto/blob/master/formats.py#L94).

toto/models/layout.py
```diff
@@ -136,6 +143,62 @@ def import_step_metadata_from_files_as_dict(self):
...
+    return True
+
+  def _validate_keys(self):
+    """ ensure that the key dictionary contains the right keys"""
+    if type(self.keys) != dict:
+      raise FormatError("keys dictionary is malformed!")
+
+    for keyid in self.keys:
+      toto.ssl_crypto.formats.KEYID_SCHEMA.check_match(keyid)
+      toto.ssl_crypto.formats.ANYKEY_SCHEMA.check_match(self.keys[keyid])
```
lukpueh: Do we need a comment here or in the docstrings that this raises `FormatError` in the case of not matching?

lukpueh: Btw. there is a [KEYDICT_SCHEMA](https://github.com/secure-systems-lab/ssl_crypto/blob/master/formats.py#L352) already combining these two.

toto/models/layout.py
```diff
@@ -136,6 +143,62 @@ def import_step_metadata_from_files_as_dict(self):
...
+      raise FormatError("Malformed date string in layout!")
+
+    return True
+
+  def _validate_keys(self):
+    """ ensure that the key dictionary contains the right keys"""
+    if type(self.keys) != dict:
+      raise FormatError("keys dictionary is malformed!")
+
+    for keyid in self.keys:
```
lukpueh: Let's use `iteritems` to iterate over dictionaries. And let's use it with `six`, to prepare for python2/3 compatibility.

toto/models/common.py
```diff
@@ -43,6 +44,14 @@ def dump(self, filename):
...
+    for method in inspect.getmembers(self, predicate=inspect.ismethod):
+        if method[0].startswith("_validate_"):
+          if not method[1]():
+            return False
+
+    return True
```
lukpueh: I don't think we should mix `True`/`False` and `Exception`/`None` styles here. As far as I understand the methods return either `True` if validation passes or raise an `Exception` else. I think we can omit the bool retruns here and in the `_validate_*` methods.

toto/models/layout.py
```diff
@@ -136,6 +143,62 @@ def import_step_metadata_from_files_as_dict(self):
...
+      raise FormatError("The inspect field should a be a list!")
+
+    for inspection in self.inspect:
+      if not isinstance(inspection, Inspection):
+        raise FormatError("The inspect list should only contain inspections!")
+
+      inspection.validate()
+
+    return True
+
```
lukpueh, line 203: Maybe it's worth adding a *validate step and inspection name uniqueness* part here?

toto/models/layout.py
```diff
@@ -176,6 +239,49 @@ def read(data):
...
+    """ check that the pubkeys is a list of keyids"""
+    if type(self.pubkeys) != list:
+      raise FormatError("The pubkeys field should be a list!")
+
+    for keyid in self.pubkeys:
+      toto.ssl_crypto.formats.KEYID_SCHEMA.check_match(keyid)
+
+  def _validate_expected_command(self):
+    """ check that the expected command is a command-like string"""
+    # TODO: Maybe be a little bit more granularity
```
lukpueh: What do you mean by granularity? I think this is fine, even if we decide to change this to [`expected_command_pattern`](https://github.com/in-toto/in-toto/issues/46), checking for a string should be fine.

toto/models/layout.py
```diff
@@ -207,3 +313,30 @@ def read(data):
...
+    """ check that the product matchrules contains a list of matchrules"""
+    if type(self.product_matchrules) != list:
+      raise FormatError("The product matchrules should be a list!")
+
+    for matchrule in self.product_matchrules:
+      toto.models.matchrule.check_matchrule_syntax(matchrule)
+
+  def _validate_run(self):
+    """ check that the expected command is a command-like string"""
+    # TODO: Maybe be a little bit more granularity
```
lukpueh: See comment above. I think this is okay!

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
+"""
+<Module Name>
+  toto/models/validators.py
```
lukpueh: Is this `toto/models/validators.py` or `toto/models/matchrule.py`? I think `matchrule_validators.py` would make sense.

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+from toto.ssl_commons.exceptions import FormatError
+import toto.ssl_crypto.formats
+
+def _validate_match_rule(keywords):
+  """ private helper to verify the syntax of the MATCH matchrule """
+  MATERIAL_OR_PRODUCT = {'PRODUCT', 'MATERIAL'}
+
+  if not isinstance(keywords, list):
+    raise FormatError("this matching rule is not a list")
```
lukpueh: Btw. I think we should allow the rule to be written as strings. A `.split()` when parsing the rule is not very expensive, whereas a split when writing the rule somehow feels stupid. I for instance script my rules like so: 
```python
material_matchrules = ["MATCH PRODUCT yadda yadda".split(), ...]
```

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
...
+    Given a list of keywords (e.g., ["CREATE", "foo"]), verify that the
+    syntax corresponds to the valid statements of the matchrules described
+    in the specification
+
+  <Returns>
+    None
+
+  <Raises>
+    FormatError: if the keywords provided do not match the matchrule syntax
+  """
```
lukpueh: Please stick to the code style guidelines: *purpose, arguments, exceptions, side effects, and return value*.

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
...
+
+def _validate_match_rule(keywords):
+  """ private helper to verify the syntax of the MATCH matchrule """
+  MATERIAL_OR_PRODUCT = {'PRODUCT', 'MATERIAL'}
+
+  if not isinstance(keywords, list):
+    raise FormatError("this matching rule is not a list")
+
+  if len(keywords) == 5:
+    rule, artifact, target, from_keyword, step = keywords
```
lukpueh: I've only seen this with list returning functions. But it looks nifty with given lists too. :)

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
...
+  This module provides functions to ensure the syntax of the matchrules is
+  correct.
+
+"""
+from toto.ssl_commons.exceptions import FormatError
+import toto.ssl_crypto.formats
+
+def _validate_match_rule(keywords):
+  """ private helper to verify the syntax of the MATCH matchrule """
+  MATERIAL_OR_PRODUCT = {'PRODUCT', 'MATERIAL'}
```
lukpueh: Is there a reason why you initialize this as a `set`? If so, I'd prefer to not use the literal set notation. I think it's easily confused with a dictionary.

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
...
+  if as_keyword != "AS":
+    raise FormatError("AS should come after the step name")
+
+  if artifact not in MATERIAL_OR_PRODUCT:
+    raise FormatError("Target should be either MATERIAL or PRODUCT!")
+
+def _validate_generic_rule(keywords):
+  """ private helper that verifies the syntax of the other rules """
+
+  VALID_OPERATIONS = {'CREATE', 'MODIFY', 'DELETE',}
```
lukpueh: See comment [above](https://github.com/in-toto/in-toto/pull/48/files#diff-7f9b331782acd3d05b10f3ab0a7609f0R24).

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
...
+
+  if not isinstance(keywords, list):
+    raise FormatError("this matching rule is not a list")
+
+  if len(keywords) != 2:
+    raise FormatError("Wrong rule format")
+
+  rule, artifact = keywords
+
+  toto.ssl_crypto.formats.PATH_SCHEMA.check_match(artifact)
```
lukpueh: Actually this can be a *path pattern*. Since [`PATH_SCHEMA`](https://github.com/secure-systems-lab/ssl_crypto/blob/master/formats.py#L123) only requires it to be any string, we are fine here.

toto/models/matchrule.py
```diff
@@ -0,0 +1,102 @@
...
+  if not isinstance(keywords, list):
+    raise FormatError("this matching rule is not a list")
+
+  if len(keywords) == 5:
+    rule, artifact, target, from_keyword, step = keywords
+    as_keyword = "AS"
+    new_target = target
+
+  elif len(keywords) == 7:
+    rule, artifact, target, from_keyword, step, as_keyword, new_target = keywords
```
lukpueh: Please take a look at the current [matchrule implementation](https://github.com/in-toto/in-toto/blob/develop/toto/verifylib.py#L280-L285)
It's actually:
```
MATCH (MATERIAL|PRODUCT) <path pattern>  [AS <target path pattern>] FROM <step name>
```

lukpueh: This also affects the rest of this function and the unittests.

toto/models/layout.py
```diff
@@ -207,3 +313,30 @@ def read(data):
     return Inspection(name=data.get("name"), run=data.get("run"),
         material_matchrules=data.get("material_matchrules"),
         product_matchrules=data.get("product_matchrules"))
+
+  def _validate_type(self):
+    """ ensure that the type field is set to inspection"""
+    if self._type != "inspection":
+      raise FormatError("The _type field should be aset to inspection!")
```
lukpueh, line 322: Tiny typo.

test/models/test_inspection.py
```diff
@@ -0,0 +1,93 @@
...
+    """ Populate a base layout that we can use """
+    self.inspection = Inspection("some-inspection")
+
+  def test_wrong_type(self):
+    """ test the type field within Validate() """
+
+    with self.assertRaises(FormatError):
+      self.inspection._type = "wrong"
+      self.inspection._validate_type()
+      self.assertFalse(self.inspection.validate())
```
lukpueh: I don't understand this. You never reach the `assertFalse` because the `FormatError` gets raised before. Please see my comment [below](https://github.com/in-toto/in-toto/pull/48/files#diff-cc71a9e93777c3c30e19a468fd9fede4R54) regarding Exception vs Bool strategy. The comment applies for most of the unittests below.

test/models/test_inspection.py
```diff
@@ -0,0 +1,93 @@
...
+      self.inspection._validate_product_matchrules()
+      self.assertFalse(self.inspection.validate())
+
+    # for more thorough tests, check the test_matchrule.py module
+    self.inspection.product_matchrules = [["CREATE", "foo"]]
+    self.inspection._validate_product_matchrules()
+    self.assertFalse(self.inspection.validate())
+
+  def test_wrong_run(self):
+
```
lukpueh, line 91: A onliner docstring would help here.

test/models/test_layout.py
```diff
@@ -0,0 +1,134 @@
...
+from toto.ssl_commons.exceptions import FormatError
+import toto.ssl_crypto
+
+class TestLayoutValidator(unittest.TestCase):
+  """Test verifylib.verify_delete_rule(rule, artifact_queue) """
+
+  def setUp(self):
+    """ Populate a base layout that we can use """
+    self.layout = Layout()
+    self.layout.expires = datetime.datetime.now().isoformat()
```
lukpueh: Please check out this [comment]( https://github.com/in-toto/in-toto/pull/48/files#diff-03000fa12a2e142e6c48c5bff130c268R156).

test/models/test_inspection.py
```diff
@@ -0,0 +1,93 @@
...
+    """ test that the material matchrule validators catch malformed ones """
+
+    with self.assertRaises(FormatError):
+      self.inspection.material_matchrules = [["NONFOO"]]
+      self.inspection._validate_material_matchrules()
+      self.assertFalse(self.inspection.validate())
+
+      self.inspection.material_matchrules = "PFF"
+      self.inspection._validate_material_matchrules()
+      self.assertFalse(self.inspection.validate())
```
lukpueh: You only get until [line 48](https://github.com/in-toto/in-toto/pull/48/files#diff-ba23fca1006b681698d19e43ee88da7eR48). You need a separate `with self.assertRaises(FormatError)` for each exception that gets raised. This is a re-occurring thing in the unit tests below.

test/models/test_layout.py
```diff
@@ -0,0 +1,134 @@
...
+
+  def test_wrong_key_dictionary(self):
+    """ test that the keys dictionary is properly populated """
+    rsa_key_one = toto.ssl_crypto.keys.generate_rsa_key()
+    rsa_key_two = toto.ssl_crypto.keys.generate_rsa_key()
+
+    with self.assertRaises(FormatError):
+
+      # FIXME: attr.ib reutilizes the default dictionary, so future constructor
+      # are not empty...
```
lukpueh: We should think about getting rid of `attr.ib`.

test/models/test_matchrule.py
```diff
@@ -0,0 +1,77 @@
...
+      check_matchrule_syntax(rule)
+
+      rule = ["MATCH", "PRODUCT", "foo", "FROM", "step", "NOT-AS"]
+      check_matchrule_syntax(rule)
+
+    rule = ["MATCH", "PRODUCT", "foo", "FROM", "step"]
+    check_matchrule_syntax(rule)
+
+    rule = ["MATCH", "PRODUCT", "foo", "FROM", "step", "AS", "bar"]
+    check_matchrule_syntax(rule)
```
lukpueh, line 81: Please see [this comment](https://github.com/in-toto/in-toto/pull/48/files#diff-7f9b331782acd3d05b10f3ab0a7609f0R35).

lukpueh (changes requested):
Thanks for all the changes. Overall, very smart and clean code. Please consider my comments and we are good to go.

lukpueh:
Btw. I wished that this PR would also add validation calls throughout the codebase where appropriate. Eg. lines like [`# FIXME: Validate rule format`](https://github.com/in-toto/in-toto/blob/develop/toto/verifylib.py#L324)

lukpueh:
Overall, clean and smart code and  much appreciated additions, @SantiagoTorres. Thanks for the PR!

SantiagoTorres:
@lukpueh let me add a commit with your comments and another addressing the FIXME's

toto/models/matchrule.py
SantiagoTorres: I initialize this as a set because it is faster to check for membership and easier to extend. Which notation would you prefer?

toto/models/matchrule.py
SantiagoTorres: The way you do it is fine. But for the sake of the validator (that checks what should be in the metadata) , I think we should explicitly require lists.

SantiagoTorres:
@lukpueh ready for another go

toto/models/layout.py
```diff
@@ -197,7 +196,10 @@ def _validate_inspections(self):
 
       inspection.validate()
 
-    return True
+      if inspection.name in names_seen:
+        raise FormatError("There is a repeated name in the steps! "
```
lukpueh, line 200:
```python
FormatError("There is a repeated name in the inspections")
```

toto/models/layout.py
```diff
@@ -315,28 +317,27 @@ def read(data):
         product_matchrules=data.get("product_matchrules"))
 
   def _validate_type(self):
-    """ ensure that the type field is set to inspection"""
+    """Private method to ensure that the type field is set to inspection."""
     if self._type != "inspection":
       raise FormatError("The _type field should be aset to inspection!")
```
lukpueh, line 322: Please change "aset" to "set"

toto/models/matchrule.py
lukpueh: Got it. Actually, I'm fine with this notation.

toto/models/matchrule.py
lukpueh: 👍  Let's delegate this issue: https://github.com/in-toto/in-toto/issues/51

toto/matchrule_validators.py
```diff
@@ -32,22 +32,24 @@ def _validate_match_rule(keywords):
     new_target = target
 
   elif len(keywords) == 7:
-    rule, artifact, target, from_keyword, step, as_keyword, new_target = keywords
+    (rule, artifact, target, as_keyword, new_target, from_keyword,
```
lukpueh: You call the left path argument "target" and the right path argument "new_target" which is different from the existing naming scheme in verifylib.
```python
["MATCH", "MATERIAL", "<path pattern>", "FROM", "<step name>"]
["MATCH", "PRODUCT", "<path pattern>", "FROM", "<step name>"]
["MATCH", "MATERIAL", "<path pattern>", "AS", "<target path pattern>", "FROM", "<step name>"]
["MATCH", "PRODUCT", "<path pattern>", "AS", "<target path pattern>", "FROM", "<step name>"]
```
I'm not happy with either of them but I think we should stay consistent, or at least add a `# FIXME` comment.

toto/models/matchrule.py
lukpueh: Ping

toto/matchrule_validators.py
```diff
@@ -65,7 +67,7 @@ def _validate_generic_rule(keywords):
 
   toto.ssl_crypto.formats.PATH_SCHEMA.check_match(artifact)
 
-  if rule not in VALID_OPERATIONS:
+  if rule not in VALID_OPERATIONS and rule.upper() not in VALID_OPERATIONS:
     raise FormatError("{} is not a valid rule!".format(rule))
 
 def check_matchrule_syntax(keywords):
```
lukpueh, line 73: Ping https://github.com/in-toto/in-toto/pull/48#discussion_r88950672

test/models/test_inspection.py
```diff
@@ -58,15 +65,20 @@ def test_wrong_material_matchrules(self):
...
+    with self.assertRaises(FormatError):
       self.assertFalse(self.inspection.validate())
 
-      self.inspection.product_matchrules = "PFF"
+    self.inspection.product_matchrules = "PFF"
+    with self.assertRaises(FormatError):
       self.inspection._validate_product_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.inspection.validate())
```
lukpueh: Don't need assertFalse here.

test/models/test_inspection.py
```diff
@@ -58,15 +65,20 @@ def test_wrong_material_matchrules(self):
...
-    """ test that the product matchrule validatores catch malformed ones """
+    """Test that the product matchrule validators catch malformed values."""
 
+    self.inspection.product_matchrules = [["NONFOO"]]
     with self.assertRaises(FormatError):
-      self.inspection.product_matchrules = [["NONFOO"]]
       self.inspection._validate_product_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.inspection.validate())
```
lukpueh: Don't need assertFalse here

test/models/test_inspection.py
```diff
@@ -26,30 +26,37 @@ class TestInspectionValidator(unittest.TestCase):
...
 
+    with self.assertRaises(FormatError):
       self.inspection.material_matchrules = "PFF"
       self.inspection._validate_material_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.inspection.validate())
```
lukpueh: Don't need assertFalse here

test/models/test_inspection.py
```diff
@@ -26,30 +26,37 @@ class TestInspectionValidator(unittest.TestCase):
...
   def test_wrong_material_matchrules(self):
-    """ test that the material matchrule validators catch malformed ones """
+    """Test that the material matchrule validators catch malformed ones."""
 
     with self.assertRaises(FormatError):
       self.inspection.material_matchrules = [["NONFOO"]]
       self.inspection._validate_material_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.inspection.validate())
```
lukpueh: Don't need assertFalse here

test/models/test_inspection.py
```diff
@@ -26,30 +26,37 @@ class TestInspectionValidator(unittest.TestCase):
...
-    """ test the type field within Validate() """
+    """Test the type field within Validate()."""
 
+    self.inspection._type = "wrong"
     with self.assertRaises(FormatError):
-      self.inspection._type = "wrong"
       self.inspection._validate_type()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.inspection.validate())
```
lukpueh: Don't need assertFalse here

test/models/test_inspection.py
```diff
@@ -75,13 +87,13 @@ def test_wrong_product_matchrules(self):
     self.assertFalse(self.inspection.validate())
```
lukpueh: I wonder why the test does not fail here. Shouldn't validate always return `None`?

test/models/test_step.py
```diff
@@ -27,44 +27,53 @@ class TestStepValidator(unittest.TestCase):
...
-    """ test the type field within Validate() """
+    """Test the type field within Validate()."""
 
+    self.step._type = "wrong"
     with self.assertRaises(FormatError):
-      self.step._type = "wrong"
       self.step._validate_type()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.step.validate())
```
lukpueh: No need for assertFalse

test/models/test_step.py
```diff
@@ -27,44 +27,53 @@ class TestStepValidator(unittest.TestCase):
...
-    """ test that the material matchrule validators catch malformed ones """
+    """Test that the material matchrule validators catch malformed ones."""
+
+    self.step.material_matchrules = [["NONFOO"]]
     with self.assertRaises(FormatError):
-      self.step.material_matchrules = [["NONFOO"]]
       self.step._validate_material_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.step.validate())
```
lukpueh: No need for assertFalse

test/models/test_step.py
```diff
@@ -27,44 +27,53 @@ class TestStepValidator(unittest.TestCase):
...
 
-      self.step.material_matchrules = "PFF"
+    self.step.material_matchrules = "PFF"
+    with self.assertRaises(FormatError):
       self.step._validate_material_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.step.validate())
```
lukpueh: No need for assertFalse

test/models/test_step.py
```diff
@@ -73,49 +82,54 @@ def test_wrong_material_matchrules(self):
...
-    """ test that the product matchrule validatores catch malformed ones """
+    """Test that the product matchrule validators catch malformed ones."""
 
+    self.step.product_matchrules = [["NONFOO"]]
     with self.assertRaises(FormatError):
-      self.step.product_matchrules = [["NONFOO"]]
       self.step._validate_product_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.step.validate())
```
lukpueh: No need for assertFalse

test/models/test_step.py
```diff
@@ -73,49 +82,54 @@ def test_wrong_material_matchrules(self):
...
 
-      self.step.product_matchrules = "PFF"
+    self.step.product_matchrules = "PFF"
+    with self.assertRaises(FormatError):
       self.step._validate_product_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.step.validate())
```
lukpueh: No need for assertFalse

test/models/test_step.py
```diff
@@ -73,49 +82,54 @@ def test_wrong_material_matchrules(self):
     self.assertFalse(self.step.validate())
```
lukpueh: I wonder why the test does not fail here. Shouldn't validate always return `None`?

test/models/test_step.py
```diff
@@ -73,49 +82,54 @@ def test_wrong_material_matchrules(self):
...
+    with self.assertRaises(FormatError):
       self.step._validate_product_matchrules()
+
+    with self.assertRaises(FormatError):
       self.assertFalse(self.step.validate())
 
     # for more thorough tests, check the test_matchrule.py module
     self.step.product_matchrules = [["CREATE", "foo"]]
     self.step._validate_product_matchrules()
     self.assertFalse(self.step.validate())
```
lukpueh: I wonder why the test does not fail here. Shouldn't validate always return `None`?

lukpueh (changes requested):
A few very small change requests. Please fix and we are good to go!

toto/matchrule_validators.py
SantiagoTorres: Hmm? I don't know how that comment relates to this.

SantiagoTorres: ping @lukpueh

lukpueh: The comment relates to `check_matchrule_syntax`'s docstring.
Name, Description, Returns, Raises != Purpose, Arguments, Exceptions, Side Effects, Return Value

toto/matchrule_validators.py
```diff
@@ -83,7 +83,7 @@ def check_matchrule_syntax(keywords):
   <Returns>
     None
 
-  <Raises>
+  <Exceptions>
```
lukpueh, line 85: Almost there. :) Please remove `<Name>`, change `<Description>` to `<Purpose>` and add `<Arguments>`.

lukpueh (changes requested):

toto/matchrule_validators.py
SantiagoTorres: lol damnit, I totally forgot these...

Issue #41, opened 2016-11-14
Layout keys dict instead of list


lukpueh:
The [docstring](https://github.com/in-toto/in-toto/blob/develop/toto/models/layout.py#L63-L65) of in-toto's `Layout` as well as the [default value](https://github.com/in-toto/in-toto/blob/develop/toto/models/layout.py#L74) suggest that the `keys` property should be a list. This does not conform with the in-toto specification and the demo implementation. 
The correct format is `ssl_crypto.formats.KEYDICT_SCHEMA`.

Please change docstring default value, and while you are at it, add model validators and unit tests!

PR #52, opened 2016-11-29
Adds in_toto_record {start, stop} command + minor changes/fixes

lukpueh:
In some cases it is not feasible to wrap a single command with toto-run, e.g. when allowedly editing a bunch of files shortly before building.
To not disrupt the in-toto supply chain we provide two additional cli tools that, where one only records materials to a temporary file and the second records products and finalizes the link metadata file.

- Adds new command line tool (with two sub commands) to perform a multi-part in-toto step
- Adds unittests for in_toto_record start/stop (requires `mock` package)
- Adds default values to optional parameters of runlib.create_link_metadata
- Changes `attr` class attribute defaults to use `attr.Factory` in order to circumvent strange memory re-use (see. https://github.com/in-toto/in-toto/commit/be7619173e13cbfb17093ac945a16083eb3d12db commit message for more details).
- Fixes wrong default type for `layout.keys` attributes (https://github.com/in-toto/in-toto/issues/41)
- Removes two obsolete utility classes:  
 - `GenericPathList` 
was used to *normalize* relative artifact paths at verification, which now happens at recording 
 - `ComparableHashDict`
was used to compare artifact hashes with multiple hash algorithms, deprecated by `util.flatten_and_invert_artifact_dict`

### DIFF
--- a/test/test_in_toto_record.py
+++ b/test/test_in_toto_record.py
@@ -1,0 +1,185 @@
+#!/usr/bin/env python
+
+"""
+<Program Name>
+  test_in_toto_record.py
+
+<Author>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+
+<Started>
+  Nov 28, 2016
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Test in_toto_record command line tool.
+
+"""
+
+import os
+import sys
+import unittest
+import logging
+import argparse
+from mock import patch
+
+from toto.util import generate_and_write_rsa_keypair
+from toto.models.link import Link
+from toto.in_toto_record import main as in_toto_record_main
+from toto.in_toto_record import in_toto_record_start, in_toto_record_stop
+
+# Suppress all the user feedback that we print using a base logger
+logging.getLogger().setLevel(logging.CRITICAL)
+
+def _try_remove_files(file_list):
+  if not isinstance(file_list, list):
+    return
+  else:
+    for path in file_list:
+      try:
+        os.remove(path)
+      except Exception, e:
+        pass
+
+
+class TestInTotoRecordMain(unittest.TestCase):
+  """Test in_toto_record's main() - requires sys.argv patching. """
+
+  @classmethod
+  def setUpClass(self):
+    """Generate key pair, dummy artifact and base arguments. """
+    generate_and_write_rsa_keypair("test-key")
+    self.args = [
+        "in_toto_record.py",
+        "--step-name",
+        "test-step",
+        "--key",
+        "test-key"
+    ]
+    self.test_artifact = "test-artifact"
+    open(self.test_artifact, "a").close()
+
+  @classmethod
+  def tearDownClass(self):
+    """Remove key pair, dummy artifact and link metadata. """
+    files_to_remove = [
+      "test-key", "test-key.pub", "test-step.link", self.test_artifact
+    ]
+    _try_remove_files(files_to_remove)
+
+
+  def test_command_start_stop_required_args(self):
+    """Test CLI command record start/stop with required arguments. """
+    with patch.object(sys, 'argv',  self.args + ["start"]):
+      in_toto_record_main()
+
+    with patch.object(sys, 'argv', self.args + ["stop"]):
+      in_toto_record_main()
+
+  def test_command_start_stop_optional_args(self):
+    """Test CLI command record start/stop with optional arguments. """
+    with patch.object(sys, 'argv', self.args + ["start", "--materials",
+        self.test_artifact]):
+      in_toto_record_main()
+
+    with patch.object(sys, 'argv', self.args + ["stop", "--products",
+        self.test_artifact]):
+      in_toto_record_main()
+
+
+
+
+
+class TestInTotoRecordStart(unittest.TestCase):
+  """"Test in_toto_record_start(step_name, key_path, material_list). """
+
+  @classmethod
+  def setUpClass(self):
+    """Generate key pair, dummy artifact and base arguments. """
+    self.step_name = "test-step"
+    self.key_path = "test-key"
+    generate_and_write_rsa_keypair(self.key_path)
+    self.test_material= "test-material"
+    open(self.test_material, "a").close()
+
+  @classmethod
+  def tearDownClass(self):
+    """Remove key pair, dummy artifact and link metadata. """
+    files_to_remove = [
+        self.key_path, self.key_path + ".pub", self.test_material,
+        "." + self.step_name + ".link-unfinished"
+      ]
+    _try_remove_files(files_to_remove)
+
+  def test_create_link_unfinished_file(self):
+    """Test record start creates a link metadata file. """
+    in_toto_record_start(self.step_name, self.key_path, [self.test_material])
+    open("." + self.step_name + ".link-unfinished", "r")
+
+  def test_record_test_material(self):
+    """Test record start records expected material. """
+    link = in_toto_record_start(
+        self.step_name, self.key_path, [self.test_material])
+    self.assertEquals(link.materials.keys(), [self.test_material])
+
+
+
+
+
+class TestInTotoRecordStop(unittest.TestCase):
+  """"Test in_toto_record_start(step_name, key_path, material_list). """
+
+  @classmethod
+  def setUpClass(self):
+    """Generate key pair, dummy artifact and base arguments. """
+    self.step_name = "test-step"
+    self.key_path = "test-key"
+    self.key_path2 = "test-key2"
+    generate_and_write_rsa_keypair(self.key_path)
+    generate_and_write_rsa_keypair(self.key_path)
+
+    self.test_product = "test-product"
+    open(self.test_product, "a").close()
+
+  @classmethod
+  def tearDownClass(self):
+    """Remove key pair, dummy artifact and link metadata. """
+    files_to_remove = [
+      self.key_path, self.key_path + ".pub", self.key_path2,
+      self.key_path2 + ".pub", self.test_product, self.step_name + ".link",
+      "." + self.step_name + ".link-unfinished"
+    ]
+    _try_remove_files(files_to_remove)
+
+  def test_record_test_product(self):
+    """Test record stop records expected product. """
+    in_toto_record_start(self.step_name, self.key_path, [])
+    link = in_toto_record_stop(
+        self.step_name, self.key_path, [self.test_product])
+    self.assertEquals(link.products.keys(), [self.test_product])
+
+  def test_replace_unfinished_file(self):
+    """Test record stop removes unfinished file. """
+    in_toto_record_start(self.step_name, self.key_path, [])
+    in_toto_record_stop(self.step_name, self.key_path, [])
+    with self.assertRaises(IOError):
+      open("." + self.step_name + ".link-unfinished", "r")
+    open(self.step_name + ".link", "r")
+
+  def test_missing_unfinished_file(self):
+    """Test record stop exits on missing unfinished file. """
+    with self.assertRaises(SystemExit):
+      in_toto_record_stop(self.step_name, self.key_path, [])
+
+  def test_wrong_signature_in_unfinished_file(self):
+    """Test record stop exits on wrong signature. """
+    in_toto_record_start(self.step_name, self.key_path, [])
+    with self.assertRaises(SystemExit):
+      in_toto_record_stop(self.step_name, self.key_path2, [])
+
+
+
+if __name__ == '__main__':
+  unittest.main()
--- a/test/test_in_toto_record.py
+++ b/test/test_in_toto_record.py
@@ -27,7 +27,8 @@ import shutil
 import tempfile
 from mock import patch
 
-from toto.util import generate_and_write_rsa_keypair
+from toto.util import generate_and_write_rsa_keypair, \
+    prompt_import_rsa_key_from_file
 from toto.models.link import Link
 from toto.in_toto_record import main as in_toto_record_main
 from toto.in_toto_record import in_toto_record_start, in_toto_record_stop
@@ -37,8 +38,10 @@ WORKING_DIR = os.getcwd()
 # Suppress all the user feedback that we print using a base logger
 logging.getLogger().setLevel(logging.CRITICAL)
 
-class TestInTotoRecordMain(unittest.TestCase):
-  """Test in_toto_record's main() - requires sys.argv patching. """
+class TestInTotoRecordTool(unittest.TestCase):
+  """Test in_toto_record's main() - requires sys.argv patching; and
+  in_toto_record_start/in_toto_record_stop - calls runlib and error logs/exits
+  on Exception. """
 
   @classmethod
   def setUpClass(self):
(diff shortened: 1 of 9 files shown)


test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+    generate_and_write_rsa_keypair("test-key")
+    self.args = [
+        "in_toto_record.py",
+        "--step-name",
+        "test-step",
+        "--key",
+        "test-key"
+    ]
+    self.test_artifact = "test-artifact"
+    open(self.test_artifact, "a").close()
```
SantiagoTorres: what about doing this in a tmpdir to avoid polluting $PWD?

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+
+  @classmethod
+  def tearDownClass(self):
+    """Remove key pair, dummy artifact and link metadata. """
+    files_to_remove = [
+      "test-key", "test-key.pub", "test-step.link", self.test_artifact
+    ]
+    _try_remove_files(files_to_remove)
```
SantiagoTorres: same here, we could replace this with `shutil.rmdir`

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+      in_toto_record_main()
+
+    with patch.object(sys, 'argv', self.args + ["stop", "--products",
+        self.test_artifact]):
+      in_toto_record_main()
+
+
+
+
+
```
SantiagoTorres, line 126: This is too much whitespace. Could you remove it please?

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+  def test_command_start_stop_optional_args(self):
+    """Test CLI command record start/stop with optional arguments. """
+    with patch.object(sys, 'argv', self.args + ["start", "--materials",
+        self.test_artifact]):
+      in_toto_record_main()
+
+    with patch.object(sys, 'argv', self.args + ["stop", "--products",
+        self.test_artifact]):
+      in_toto_record_main()
```
SantiagoTorres, line 84: I wonder if we need this use of mock, or if it's just a byproduct of a weird architecture.  Wouldn't it be easier if we had a function to call? e.g., `start_recording(**kwargs)` and `stop_recording(**kwargs)`

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+  def setUpClass(self):
+    """Generate key pair, dummy artifact and base arguments. """
+    self.step_name = "test-step"
+    self.key_path = "test-key"
+    self.key_path2 = "test-key2"
+    generate_and_write_rsa_keypair(self.key_path)
+    generate_and_write_rsa_keypair(self.key_path)
+
+    self.test_product = "test-product"
+    open(self.test_product, "a").close()
```
SantiagoTorres: why do we open it in "append" mode?

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+
+  @classmethod
+  def tearDownClass(self):
+    """Remove key pair, dummy artifact and link metadata. """
+    files_to_remove = [
+      self.key_path, self.key_path + ".pub", self.key_path2,
+      self.key_path2 + ".pub", self.test_product, self.step_name + ".link",
+      "." + self.step_name + ".link-unfinished"
```
SantiagoTorres: I think it'd be better if you used `os.path.join()` here

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+            List of file or directory paths that should be recorded as
+            materials.
+
+  <Exceptions>
+    None.
+
+  <Side Effects>
+    Loads private key from disk
+    Creates a file with  name ".<step_name>.link-unfinished"
+    Calls sys.exit(1) if an exception is raised
```
SantiagoTorres, line 74: would it be better to return with 127?

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+
+  <Returns>
+    Link object (unfinished)
+
+  """
+
+  unfinished_fn = "." + str(step_name) + ".link-unfinished"
```
SantiagoTorres: can we use format for this?

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+    log.doing("create preliminary link metadata...")
+    link = runlib.create_link_metadata(step_name, materials_dict)
+
+    log.doing("sign metadata...")
+    link.sign(key)
+
+    log.doing("store metadata to '{0}'...".format(unfinished_fn))
+    link.dump(unfinished_fn)
+
+    return link
```
SantiagoTorres: Why  is the lnk returned after dumping?

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+            layout.
+    key_path:
+            Path to private key to verify unfinished link metadata and
+            sign finished link metadata (PEM).
+    product_list:
+            List of file or directory paths that should be recorded as products.
+
+  <Exceptions>
+    Exception if signature fails
+    IOException if file ".<step_name>.link-unfinished" can't be read
```
SantiagoTorres: or written, I assume.

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+    Link object
+
+  """
+
+  unfinished_fn = "." + str(step_name) + ".link-unfinished"
+
+  try:
+    log.doing("load link signing key...")
+    key = toto.util.prompt_import_rsa_key_from_file(key_path)
+    keydict = {key["keyid"] : key}
```
SantiagoTorres: does this make the keyid be repeated inside the value? maybe we don't want to have this.

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+    # The unfinished link file must have been signed by the same key
+
+    log.doing("verify unfinished link signature...")
+    link.verify_signatures(keydict)
+
+    log.doing("record products...")
+    link.products = runlib.record_artifacts_as_dict(product_list)
+
+    log.doing("update signature...")
+    link.signatures = []
```
SantiagoTorres: I thought we didn't need this anymore, do we?

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+  parser.usage = ("\n"
+      "%(prog)s  --step-name <unique step name>\n{0}"
+               " --key <functionary private key path>\n"
+      "Commands:\n{0}"
+               "start [--materials <filepath>[ <filepath> ...]]\n{0}"
+               "stop  [--products <filepath>[ <filepath> ...]]\n"
+               .format(lpad))
+
+  in_toto_args = parser.add_argument_group("Toto options")
+  # FIXME: Name has to be unique!!! Where will we check this?
```
SantiagoTorres: I don't think we can check this here.

toto/in_toto_record.py
```diff
@@ -0,0 +1,221 @@
...
+  # FIXME: Do we limit the allowed characters for the name?
```
SantiagoTorres, line 142: Good point, this is not on the specification, so I need to consider it. I wouldn't worry about this in here though.

toto/models/common.py
```diff
@@ -73,7 +71,7 @@ class Signable(Metablock):
   """Objects with base class Signable can sign their payload (a canonical
   pretty printed JSON string not containing the signatures attribute) and store
   the signature (signature format: ssl_crypto__formats.SIGNATURE_SCHEMA) """
-  signatures = attr.ib([])
+  signatures = attr.ib(default=attr.Factory(list))
```
SantiagoTorres, line 75: 👍

toto/models/layout.py
```diff
@@ -69,17 +69,18 @@ class Layout(models__common.Signable):
         a list of Inspection objects
 
     keys:
-        a list of public keys used to verify the signature of link
-        metadata file related to a step
+        A dictionary of  public keys used to verify the signature of link
```
SantiagoTorres: Minor nit: remove the double space please.

toto/runlib.py
```diff
@@ -177,8 +177,8 @@ def execute_link(link_cmd_args, record_byproducts):
 
   return {"stdout": stdout_str, "stderr": stderr_str}, return_value
 
-def create_link_metadata(link_name, materials_dict, products_dict,
-    link_cmd_args, byproducts, return_value):
+def create_link_metadata(link_name, materials_dict={}, products_dict={},
+    link_cmd_args="", byproducts={}, return_value=None):
```
SantiagoTorres, line 182: do we want to have default values here? or default to non and use ifs instead?

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+  """"Test in_toto_record_start(step_name, key_path, material_list). """
+
+  @classmethod
+  def setUpClass(self):
+    """Generate key pair, dummy artifact and base arguments. """
+    self.step_name = "test-step"
+    self.key_path = "test-key"
+    self.key_path2 = "test-key2"
+    generate_and_write_rsa_keypair(self.key_path)
+    generate_and_write_rsa_keypair(self.key_path)
```
SantiagoTorres: did you mean key_path2 here?

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+import argparse
+from mock import patch
+
+from toto.util import generate_and_write_rsa_keypair
+from toto.models.link import Link
+from toto.in_toto_record import main as in_toto_record_main
+from toto.in_toto_record import in_toto_record_start, in_toto_record_stop
+
+# Suppress all the user feedback that we print using a base logger
+logging.getLogger().setLevel(logging.CRITICAL)
```
SantiagoTorres, line 39: Maybe it would be convenient to have a "disable" call on the logger that replaces everything with a null handler.

test/test_in_toto_record.py
SantiagoTorres: Now I see, maybe we could have "main()" take the sys.* arguments instead?

SantiagoTorres (changes requested):
It overall looks good. Most of my stuff were minor nits.

test/test_in_toto_record.py
lukpueh: I don't know if this changes much regarding convenience. In any case I'd like to rethink the logging/user feedback architecture with https://github.com/in-toto/in-toto/issues/6#issuecomment-259750991.

test/test_in_toto_record.py
lukpueh: Good idea! Will do.

test/test_in_toto_record.py
lukpueh: Ditto

test/test_in_toto_record.py
lukpueh: What do you mean by "main() taking the sys.* arguments"? Something like`main(sys.argv)`? I'm not sure if the architecture is weird. Let me know if you have architectural suggestions.

test/test_in_toto_record.py
lukpueh: Will do

test/test_in_toto_record.py
lukpueh: Well spotted, thanks!

test/test_in_toto_record.py
lukpueh: No reason. Will change to `w`

test/test_in_toto_record.py
lukpueh: Why? This is only a list of file names. Anyway, I will change this to use a tmp directory as you suggested earlier.

toto/in_toto_record.py
lukpueh: "command not found"? Does not seem suitable to me.

toto/in_toto_record.py
lukpueh: Yes we can. Will change.

toto/in_toto_record.py
lukpueh: TBH, I did this to inspect the link object in a unittest. But know that you mention it it does not seem so smart anymore.

toto/in_toto_record.py
lukpueh: `in_toto_record_stop` does not write to the file. It only reads it and then removes it.

lukpueh: I guess we would get an IOException if we don't have write access to the parent directory. Should we mention that?

lukpueh: Actually that would be an `OSError`.

toto/in_toto_record.py
lukpueh: Why not, this is `ssl_crypto.formats.KEYDICT_SCHEMA`

toto/models/layout.py
lukpueh: Will do

toto/in_toto_record.py
lukpueh: This is different. Here we load an unfinished link object and expect it to already have a signature (gets verified above). Then we update the object, clear the existing signatures and re-sign.

toto/runlib.py
lukpueh: Yes we do, because in some cases (e.g. `in_toto_record_start`) it makes sense to create an almost empty link metadata. I thought about your concern (especially given `attrs` fix in this PR) but if you look at the function the default data structures don't get updated. So we should't have a problem here.

test/test_in_toto_record.py
SantiagoTorres: well, you need specific info from argv right? (e.g., a list of products, a list of materials, etc.)

test/test_in_toto_record.py
SantiagoTorres: ah right, I also think format makes it more readable

toto/in_toto_record.py
SantiagoTorres: d'oh. You're right. I was reading the [glibc docs](https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html) and it irks me that 1 can mean "everything ok but unwanted output" (e.g., in the case of diff). So maybe 2 is unambiguous?

This is a minor nit though. I guess 1 is also a good choice.

toto/in_toto_record.py
SantiagoTorres: I mean, it's relatively convenient. I was curious as to why, maybe it's good interface and we can mention it as the return value in the docs?

toto/in_toto_record.py
SantiagoTorres: Yep, that's what I meant. OSError (I thought it was IOError) when you can't write the finished link.

toto/in_toto_record.py
SantiagoTorres: is it? I find it surprising that we need to key using the keyid and then keep the keyid inside the value also. This isn't something we should discuss in this PR though.

toto/in_toto_record.py
SantiagoTorres: Oh, right 👍

toto/runlib.py
SantiagoTorres: Yeah, I guess this is a non-issue. I can't think of an example in which create_link_metadata is called twice in the same process.

test/test_in_toto_record.py
lukpueh: The reason why I use `mock` here is that I want to test the argparser part of the cli tool.

toto/in_toto_record.py
lukpueh: I just realize that `in_toto_record_start` and `in_toto_record_stop` always raise a `SystemExit` exception. These are not library functions but should only be called by the command line tool.

toto/in_toto_record.py
lukpueh: I will change this because the function should only be called by the command line tool.

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+
+  def test_command_start_stop_optional_args(self):
+    """Test CLI command record start/stop with optional arguments. """
+    with patch.object(sys, 'argv', self.args + ["start", "--materials",
+        self.test_artifact]):
+      in_toto_record_main()
+
+    with patch.object(sys, 'argv', self.args + ["stop", "--products",
+        self.test_artifact]):
+      in_toto_record_main()
```
SantiagoTorres, line 84: Speaking of testing argparse, woudln't we want to test with extraneous input (e.g., wrong flags?)

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+  @classmethod
+  def tearDownClass(self):
+    """Change back to initial working dir and remove temp test directory. """
+    os.chdir(WORKING_DIR)
+    shutil.rmtree(self.test_dir)
+
+  def test_create_link_unfinished_file(self):
+    """Test record start creates a link metadata file. """
+    in_toto_record_start(self.step_name, self.key_path, [self.test_material])
+    open(self.link_name_unfinished)
```
SantiagoTorres: do we want to do something a little bit more thorough here?

test/test_in_toto_record.py
```diff
@@ -0,0 +1,185 @@
...
+    in_toto_record_start(self.step_name, self.key_path, [])
+    in_toto_record_stop(self.step_name, self.key_path, [])
+    with self.assertRaises(IOError):
+      open(self.link_name_unfinished, "r")
+    open(self.link_name, "r")
+
+  def test_missing_unfinished_file(self):
+    """Test record stop exits on missing unfinished file. """
+    with self.assertRaises(SystemExit):
+      in_toto_record_stop(self.step_name, self.key_path, [])
```
SantiagoTorres: Is this test assuming that the tests run in order? that may not always be the case iirc. Maybe we want to ensure that there is no unfinished file before the `with self.assertRaises()`

toto/in_toto_record.py
```diff
@@ -0,0 +1,215 @@
...
+    log.doing("create preliminary link metadata...")
+    link = runlib.create_link_metadata(step_name, materials_dict)
+
+    log.doing("sign metadata...")
+    link.sign(key)
+
+    log.doing("store metadata to '{0}'...".format(unfinished_fn))
+    link.dump(unfinished_fn)
+
+  except Exception, e:
```
SantiagoTorres: `Exception as e`, please

toto/in_toto_record.py
```diff
@@ -0,0 +1,215 @@
...
+
+  <Returns>
+    None.
+
+  """
+
+  unfinished_fn = ".{}.link-unfinished".format(step_name)
+
+  try:
+    log.doing("Start recording '{0}'...".format(step_name))
```
SantiagoTorres: Maybe we could use present continuous instead of simple present?
```
"recording {0} ..."
"recording materials..."
"creating preliminary link metadata"
"signing metadata"
"storing metadata in {0}"
```

toto/in_toto_record.py
```diff
@@ -0,0 +1,215 @@
...
+    Writes link file to disk "<step_name>.link"
+    Removes unfinished link file ".<step_name>.link-unfinished" from disk
+    Calls sys.exit(1) if an exception is raised
+
+  <Returns>
+    None.
+
+  """
+
+  unfinished_fn = ".{}.link-unfinished".format(step_name)
```
SantiagoTorres: Maybe we could keep this constant string as a module_level global
```python
UNFINISHED_FILENAME_FORMAT = ".{step-name}.link-unfinished"
...
def ...
  unfinished_fn = UNFINISHED_FILENAME_FORMAT.format(step_name=step_name)
```

toto/in_toto_record.py
```diff
@@ -0,0 +1,215 @@
...
+    link.signatures = []
+    link.sign(key)
+
+    log.doing("store metadata to file...")
+    link.dump()
+
+    log.doing("remove unfinished file...")
+    os.remove(unfinished_fn)
+
+  except Exception, e:
```
SantiagoTorres: exception as e please

SantiagoTorres (changes requested):
I think it's pretty much there, thanks!

toto/in_toto_record.py
lukpueh: Sounds good to me. Let's fix this with https://github.com/in-toto/in-toto/issues/6#issuecomment-264205129

toto/in_toto_record.py
lukpueh: Good idea.

toto/in_toto_record.py
lukpueh: Will do

test/test_in_toto_record.py
lukpueh: Okay, I can do this. The *problem* is that argparse does not raise meaningful exceptions per default, it just prints something like e.g. `in_toto_record.py: error: too few arguments` to the console and then exits. I see two ways:
- override arpgarse's `error()` method to raise meaningful exceptions which I can assert
- assert `SystemExit` (plus maybe buffer output in unittest and grep for argparse error message)

Any strong feelings about either one?

test/test_in_toto_record.py
lukpueh: Do you have something in mind? I want the individual function that are called in `test_create_link_unfinished_file` to get their own thorough unittests.

test/test_in_toto_record.py
SantiagoTorres: Hmm, I'm not sure, to me it feels like testing argparse is not completely necessary if we have a proper architecture. Imho, doing the following would suffice:

1.- change main() to receive explicit arguments
2.- test main to check for proper arguments
3.- Move the argparse function to a function that does that, parse args
4.- Test that function separately

test/test_in_toto_record.py
SantiagoTorres: e.g., something is saved, is that all we need to know? maybe we want to check that it is an actual piece of link metadata (maybe use the read method?). This test passes if you replace `in_toto_record_start()` with `touch $filename`

test/test_in_toto_record.py
lukpueh: `in_toto_record_start` and `in_toto_record_stop` already receive those explicit arguments and are tested separately. `main` just does the parsing and calls either start or stop.
I think we can leave main like it is. If we want to test the argparser for wrong arguments we still need to decide between the two approaches (check for custom ArgparseException or SystemExit).

SantiagoTorres: So this is what I don't understand, this is a script that has a `main` function and only runs that `main` function. Why do we declare `main` in the first place? is it going to be called from somewhere else (e.g., `import x; x.main()`)? if we do this, then would a caller need to mock sys also? Wouldn't it be more explicit to have main's functionality inside the if? My feeling is that this hinders functional separation, and that's why we are needing these mocks.

SantiagoTorres:
LGTM 💯

PR #54, opened 2016-12-02
Removes toto-bundle, bundlelib and layoutlib

lukpueh:
These files have been dragged along since one of the first commits but we never actually implemented them. If we need something like this we can always re-add. For now, let's get the unit test coverage up! :)


SantiagoTorres:
👍

Issue #43, opened 2016-11-17
Fix MATCH rule ambiguities


lukpueh:
The current rule type `MATCH` uses the following language:
```
MATCH (MATERIAL|PRODUCT) <path pattern> [AS <target path pattern>] FROM <step name>
```
Currently this reads (my interpretation of the specification):
>`MATERIALS` or `PRODUCTS`, depending on the rule's second element, of a step with `<step name>` and matched by `<target path pattern>`  (if specified) or `<path pattern>` (else), must have a hash that matches the hash of a `MATERIAL` or `PRODUCT`, depending on whether the rule is in the `material_matchrules` or `product_matchrules`, of a step that matches `<path pattern>` and has not been matched by a previous rule. The rule also fails if the target path pattern does not find any artifacts.

When I and @SantiagoTorres, who came up with the language and implemented it, tried to apply these rules in real world software supply chains, we both had difficulties to express what we wanted, which IMHO indicates that the language should be clarified and freed from ambiguities.
The following (or at least most of it) should be unmistakingly clear by just reading the rule:

- **What is the match source and what is the match target?**
   rule in one of step's rule lists VS.  `... FROM <step name>`

- **What is the source artifact type and what is the target artifact type?**
   rule in `material_matchrules` or `product_matchrules`  VS. `MATCH (MATERIAL|PRODUCT)...` 

- **Should there be a target artifact for every source artifact? (currently no)**
  material_matchrule (source): `MATCH PRODUCT *  FROM target-step`
  source materials : `hash(foo) = x`, `hash(bar) = y`
  target products:    `hash(foo) = x`
  Pass?

- **Should there be a source artifact for every target artifact? (currently yes)**
  material_matchrule (source): `MATCH PRODUCT *  FROM target-step`
  source materials : `hash(baz) = x`
  target products:    `hash(baz) = x `, `hash(zap) = y`
  Fail?

- **Should the rule fail if the path pattern does not find any artifacts?**
  `MATCH PRODUCT foo ...` should probably fail if there is no product foo?
  `MATCH PRODUCT * ...` should probably not fail if there is no product?

- **Do we solely match by hash in case of wildcard patterns?**
  material_matchrule (source): `MATCH PRODUCT *  FROM target-step`
  source materials : `hash(new/path/foo) = x`, `hash(bar) = y`
  target products:    `hash(foo) = x`, `hash(baz) = y`
  
[...]

PR #56, opened 2016-12-21
Adds setting to exclude artifacts in link metadata

lukpueh:
- Adds routine to exclude files, matched by defined patterns (uses fnmatch), when recording artifacts (materials or products) upon in-toto-run/in-toto-record
- For now the exclude patterns can be defined as a setting ARTIFACT_EXCLUDES. This is likely to become a command line arg or part of a config file, e.g. .in-toto-ignore
- Set ARTIFACT_EXLCUDES to ignore ".git" and link metadata files
- Adds unit tests

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -119,9 +119,10 @@ def record_artifacts_as_dict(artifacts):
   if not artifacts:
     return artifacts_dict
 
-  # print settings.ARTIFACT_EXCLUDES
+  # Exclude excluded files and dirs from the first level of tree traversel
   for artifact in _apply_exclude_patterns(artifacts,
         settings.ARTIFACT_EXCLUDES):
+
     if not os.path.exists(artifact):
       log.warning("path: {} does not exist, skipping..".format(artifact))
       continue
@@ -131,6 +132,12 @@ def record_artifacts_as_dict(artifacts):
 
     elif os.path.isdir(artifact):
       for root, dirs, files in os.walk(artifact):
+        # Also exclude excluded subdirs
+        # Modifying the dirs variable makes walk only recurse into
+        # remaining dirs
+        dirs[:] = _apply_exclude_patterns(dirs, settings.ARTIFACT_EXCLUDES)
+
+        # Finally exclude excluded tree nodes (files)
         for name in _apply_exclude_patterns(files, settings.ARTIFACT_EXCLUDES):
           filepath = os.path.join(root, name)
           artifacts_dict[_normalize_path(filepath)] = _hash_artifact(filepath)
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -28,6 +28,8 @@ import sys
 import os
 import tempfile
 import logging
+import fnmatch
+from simple_settings import settings
 
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
 # install and use the much more recent subprocess32
@@ -72,6 +74,14 @@ def _normalize_path(path):
   return path
 
 
+def _apply_exclude_patterns(names, exclude_patterns):
+  """Exclude matched patterns from passed names. """
+
+  for exclude_pattern in exclude_patterns:
+    excludes = fnmatch.filter(names, exclude_pattern)
+    names = list(set(names) - set(excludes))
+  return names
+
 def record_artifacts_as_dict(artifacts):
   """
   <Purpose>
@@ -81,6 +91,14 @@ def record_artifacts_as_dict(artifacts):
     The files a link command is executed on are called materials.
     The files that result form a link command execution are called
     products.
+
+    Excludes files that are matched by the file patterns specified in
+    ARTIFACT_EXCLUDES setting.
+    Note: Exclude patterns are applied on the basename, i.e. the pattern
+    "foo" excludes /foo as well as /subdir/foo
+
+    Exclude patterns are likely to become command line arguments or part of
+    a config file, e.g. .in-toto-ignore
 
   <Arguments>
     artifacts:
@@ -101,17 +119,19 @@ def record_artifacts_as_dict(artifacts):
   if not artifacts:
     return artifacts_dict
 
-  for artifact in artifacts:
-
+  # print settings.ARTIFACT_EXCLUDES
+  for artifact in _apply_exclude_patterns(artifacts,
+        settings.ARTIFACT_EXCLUDES):
     if not os.path.exists(artifact):
       log.warning("path: {} does not exist, skipping..".format(artifact))
       continue
 
     if os.path.isfile(artifact):
       artifacts_dict[_normalize_path(artifact)] = _hash_artifact(artifact)
+
     elif os.path.isdir(artifact):
       for root, dirs, files in os.walk(artifact):
-        for name in files:
+        for name in _apply_exclude_patterns(files, settings.ARTIFACT_EXCLUDES):
           filepath = os.path.join(root, name)
           artifacts_dict[_normalize_path(filepath)] = _hash_artifact(filepath)
 
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -67,13 +67,6 @@ def _hash_artifact(filepath, hash_algorithms=['sha256']):
   return hash_dict
 
 
-def _normalize_path(path):
-  """Internal helper that strips "./" on the left side of the path. """
-  if path.startswith("./"):
-      return path[2:]
-  return path
-
-
 def _apply_exclude_patterns(names, exclude_patterns):
   """Exclude matched patterns from passed names. """
 
@@ -81,6 +74,7 @@ def _apply_exclude_patterns(names, exclude_patterns):
     excludes = fnmatch.filter(names, exclude_pattern)
     names = list(set(names) - set(excludes))
   return names
+
 
 def record_artifacts_as_dict(artifacts):
   """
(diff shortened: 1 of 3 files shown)


lukpueh:
Exclude patterns are applied 3 times:
(1) On the first level of the directory tree when artifacts could be either files or directories,
(2) on intermediate directories, i.e. dirs that are not on the first level, and
(3) on tree leaves that are not on the first level.

*EXAMPLE:*
```
# Directory Tree
.
├── foo.pyc
└── sub
    ├── .git
    └── bar.pyc

# Setting
ARTIFACT_EXCLUDES = [".git", "*.pyc]
# Call record_artifacts_as_dict to record from root of above tree
record_artifacts_as_dict(".")
```
(1) excludes foo.pyc
(2) excludes sub/.git
(3) excludes bar.pyc

test/test_runlib.py
```diff
@@ -22,24 +22,126 @@
...
+    """Traverse dir and subdirs. Exclude subsubdir. """
+    settings.ARTIFACT_EXCLUDES = ["subsubdir"]
+    artifacts_dict = record_artifacts_as_dict(["."])
+
+    for key, val in artifacts_dict.iteritems():
+      ssl_crypto.formats.HASHDICT_SCHEMA.check_match(val)
+
+    self.assertListEqual(sorted(artifacts_dict.keys()),
+        sorted(["foo", "bar", "subdir/foosub"]))
+    settings.ARTIFACT_EXCLUDES = []
```
SantiagoTorres: I wonder if it would be a good idea to have this under a `try:finally:` or a `TearDown()` so that clearing the excludes is always performed and doesn't pollute ensuing tests.

test/test_runlib.py
```diff
@@ -22,24 +22,126 @@
...
+
+    self.test_dir = tempfile.mkdtemp()
+    os.chdir(self.test_dir)
+    open("foo", "w").write("foo")
+    open("bar", "w").write("bar")
+
+    os.mkdir("subdir")
+    os.mkdir("subdir/subsubdir")
+    open("subdir/foosub", "w").write("foosub")
+    open("subdir/subsubdir/foosubsub", "w").write("foosubsub")
```
SantiagoTorres, line 110: I'm curious, could we add a file that hashes to the same (e.g,. a second file with "foosub") to trigger the "files with the same hash" cornercase?

test/test_runlib.py
```diff
@@ -22,24 +22,126 @@
...
+
+  @classmethod
+  def tearDownClass(self):
+    """Change back to initial working dir and remove temp test directory. """
+    os.chdir(WORKING_DIR)
```
SantiagoTorres: I'm thinking that maybe this global WORKING_DIR could be part of the object and populated upon `setUpClass()`. Just for the sake of symmetry with `self.test_dir`

test/test_runlib.py
```diff
@@ -22,24 +22,126 @@
...
 import tempfile
+from in_toto import ssl_crypto
 from in_toto.runlib import in_toto_record_start, in_toto_record_stop, \
-    UNFINISHED_FILENAME_FORMAT
+    UNFINISHED_FILENAME_FORMAT, record_artifacts_as_dict, \
+    _apply_exclude_patterns
 from in_toto.util import generate_and_write_rsa_keypair, \
     prompt_import_rsa_key_from_file
 from in_toto.models.link import Link
 from in_toto.exceptions import SignatureVerificationError
```
SantiagoTorres, line 32: Is this used anywhere?

SantiagoTorres: Same for `Link`

in_toto/settings.py
```diff
@@ -65,6 +65,8 @@
 # 'pyca-cryptography']
 GENERAL_CRYPTO_LIBRARY = 'pycrypto'
 
+# FIXME: Add as command line argument or to config file, e.g .in-toto-ignore
+ARTIFACT_EXCLUDES=["*.link*", ".git"]
```
SantiagoTorres: I like these defaults. I wonder if we can extend this list with `*.pyc` and  `*~`

in_toto/runlib.py
```diff
@@ -82,6 +92,14 @@ def record_artifacts_as_dict(artifacts):
     The files that result form a link command execution are called
     products.
 
+    Excludes files that are matched by the file patterns specified in
+    ARTIFACT_EXCLUDES setting.
+    Note: Exclude patterns are applied on the basename, i.e. the pattern
+    "foo" excludes /foo as well as /subdir/foo
```
SantiagoTorres: I see that this is a consequence of using fnmatch. Maybe in the future we want to change this behavior? (e.g,. have `**/.git` to ignore all .git folders in any subdirectory and `.git` to only ignore the one in root).

SantiagoTorres (commented):
I think none of these issues is a blocker, but I'd like to see what's your insight on my comments before I merge. 👍

test/test_runlib.py
lukpueh: Yes. :)

test/test_runlib.py
lukpueh: Yes and no. This TestCase only checks `runlib.record_artifacts_as_dict` which recursively traverses the passed list of paths and stores each file together with its hash. It does not care for hash collisions.

The "files with the same hash" corner case only appears in `verifylib.verify_match_rule`, where we try to map files from one artifact dict (as recorded above) to the files from another artifact dict.

Moreover, the hash collision issue has been dealt with in: https://github.com/in-toto/in-toto/commit/75ca2ce6f4b04fb9ca3bda66b282371883de9eb1
which I used for the [in-toto in Seattle](https://github.com/in-toto/in-toto/tree/seattle) show case, but which has not been PRed yet, because of ongoing discussions in https://github.com/in-toto/in-toto/issues/43.

test/test_runlib.py
lukpueh: Yes, makes sense. My reasoning was to reduce some redundant code, because there are multiple test cases that work in a similar way in terms of a temp directory. But I guess the global is only a half-baked approach. I'll move the global to the setUpClass methods.

test/test_runlib.py
lukpueh: Excellent idea. Will do.

in_toto/runlib.py
lukpueh: Actually it is not a consequence of using fnmatch. I implemented this behavior on purpose, because I thought this is how .gitignore works. But maybe I was wrong. Do you want me to change this now?

test/test_runlib.py
SantiagoTorres: kk, I was just making sure. I grepped for it and it's indeed used.

test/test_runlib.py
SantiagoTorres: Hmm, I still think it wouldn't hurt to have this to avoid any sort of "regression"? (I know it's not a regression, but oh well)

in_toto/runlib.py
SantiagoTorres: Hmm, I'm assuming it would be a good idea to mimic the behavior/semantic that .gitignore does to simplify adoption. Both files, gitignore and in-totoignore, serve the same purpose, so why not?

We could ticketize it and move forward with the PR though.

test/test_runlib.py
lukpueh: Okay, I'll add it since it does not hurt as you say.

in_toto/runlib.py
lukpueh: You know what. I'll add it right away. Let me revise the .gitignore behavoir...

lukpueh: sdf

lukpueh:
Updated in c062460:

*EXAMPLE:*
```
# Directory Tree
.
├── foo.pyc
└── sub
    ├── .git
    └── bar.pyc

# Setting
ARTIFACT_EXCLUDES = [".git", "*.pyc]
# Call record_artifacts_as_dict to record from root of above tree
record_artifacts_as_dict(".")
```
(1) excludes foo.pyc
(2) ~~excludes sub/.git~~ (`"*.git"` would also exclude `sub/.git`)
(3) excludes bar.pyc

in_toto/runlib.py
```diff
@@ -119,28 +133,56 @@ def record_artifacts_as_dict(artifacts):
...
+        for dirname in dirs:
+          norm_dirpath = os.path.normpath(os.path.join(root, dirname))
+          dirpaths.append(norm_dirpath)
+
+        # Apply exlcude patterns on normalized dirpaths
+        dirpaths = _apply_exclude_patterns(dirpaths, settings.ARTIFACT_EXCLUDES)
+
+        # Reset and refill dirs with remaining names after exclusion
+        # Modify (not reassign) dirnames to only recurse into remaining dirs
+        dirs[:] = []
```
SantiagoTorres, line 168: hmm, I wonder if this is the behavior that we want. You are assigning the same list to all the elements in dir... Have you verified this behavior?

lukpueh: `dirs[:] = []` is equal to `del dirs[:]` is equal to `clear the list` which is the behavior I want

lukpueh: Modifying the `dirs` list prevents `os.walk` from further recursing into directories that are no longer in the list.

lukpueh: From the python [`os.walk docs`](https://docs.python.org/2/library/os.html#os.walk)
> ...the caller can modify the dirnames list in-place (perhaps using del or slice assignment), and walk() will only recurse into the subdirectories whose names remain in dirnames...

SantiagoTorres: Right, I just verified this behavior. Merging!

PR #58, opened 2017-01-05
Adds artifact base path setting (fix #44)

### LINKED ISSUE #44: Deal with (relative) artifacts path recordings
It can occur that the directory in which `toto-run` is executed and the base of the directory to record `materials` and `products` are not the same.
Consider below directory layout and a `toto-run` to be executed in `b` and record materials  in `a`:
```shell
├─/path/to/project
          ├── a
          │   ├── bar
          │   └── foo
          └── b
# in /path/to/project/b
toto-run.py --materials ../a
```
this results in a `materials` list:
```python
materials = [
   "../a/bar" : <hash>
   "../a/foo" : <hash>
]
```
Currently  matchrules would need to add dots or double-dots in the path pattern to match the recorded artifacts. This can be circumvented by using wildcard patterns, which does not seem ideal.

Two possible solutions:
- Ignore leading dots/double dots at verification time
- Add a `base-path` command line option to `toto-run`, record absolute artifact paths, leftstrip base path from absolute artifact path before writing link metadata.

lukpueh:
Adds new optional setting `ARTIFACT_BASE_PATH`. If set, it will be used as starting point when traversing the file system to record artifacts, e.g. when using `in-toto-run` or `in-toto-record`.

Example:
```
# Given the following directory structure
├─/artifact/base/path/
      ├── bar
      └── foo

ARTIFACT_BASE_PATH = "/artifact/base/path"

# Run from any directory
in-toto-run --materials . --products . ...
# will record "bar" and "foo" as materials and products
```

- The base path is ignored when running inspection upon `in-toto-verify`.
- This PR will conflict with https://github.com/in-toto/in-toto/pull/56. Let's first merge https://github.com/in-toto/in-toto/pull/56 first and I will fix the conflicts.
- Fixes https://github.com/in-toto/in-toto/issues/44

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -29,6 +29,7 @@ import os
 import tempfile
 import logging
 import fnmatch
+
 from simple_settings import settings
 
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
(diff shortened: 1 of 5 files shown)


SantiagoTorres:
I think this LGTM, I couldn't find anything that stood out. However, I'm worried that it could be sleep deprivation. Let me take a better look later tonight.

PR #59, opened 2017-01-05
Adds parentheses in multi-line imports (fix #57)

### LINKED ISSUE #57: Use parentheses in multi-line imports
I've noticed that we are doing imports as follows:

```python
from x import y, z, \
           i
```
While this is a correct way of doing it, I'd suggest we move to the parentheses construction instead:

```python
from x import (y, z,
         i)
```
The rationale behind is that our code style (which is based on pep8) [strongly hints of using parentheses for line continuation](https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length)<sup>*</sup>. [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) also makes the same suggestion:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation. 

If I recall correctly, tools such as flake or pep8 mark this as a style error....

--- 
<sub>\*Should we also consider updating the example in our style guideline? </sub>

lukpueh:
PEP8 suggests to use parentheses in line continuation in
preference to using backslashes.

This replaces backslash line continuation with parentheses in
all multi-line imports.

Fixes https://github.com/in-toto/in-toto/issues/57.

### DIFF
--- a/test/test_in_toto_record.py
+++ b/test/test_in_toto_record.py
@@ -27,8 +27,8 @@ import shutil
 import tempfile
 from mock import patch
 
-from in_toto.util import generate_and_write_rsa_keypair, \
-    prompt_import_rsa_key_from_file
+from in_toto.util import (generate_and_write_rsa_keypair,
+    prompt_import_rsa_key_from_file)
 from in_toto.models.link import Link
 from in_toto.in_toto_record import main as in_toto_record_main
 from in_toto.in_toto_record import in_toto_record_start, in_toto_record_stop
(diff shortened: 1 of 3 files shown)


SantiagoTorres:
pretty self, contained. If I merge this does this interfere with the other PRs?

lukpueh:
Yes it does, but it's fine, please go ahead. I'll rebase, resolve, re-push.

SantiagoTorres:
LGTM!! Thanks!

PR #60, opened 2017-01-06
Enhance separation between in-toto-run cli tool and library

lukpueh:
Moves complete in_toto_run workflow to `runlib.in_toto_run`. 

The user-facing in-toto-run command line tool just takes care of command line argument parsing and calls `runlib.in_toto_run`. In case an exception occurs it is caught in the calling CLI tool, which exits gracefully.

The CLI tool also takes care of loading the key from disk potentially prompting for a password which should not happen in the library.
- Adds an optional `--verbose` flag to `in-toto-run` CLI tool that turns on logging statements in the library.
- Makes `runlib.in_toto_run`'s key argument optional, so that it can also be used when running inspections, which do not need to sign generated link metadata.
- Changes `verifylib` to dump link metadata when running inspections.
- Enhances logging statements in runlib.
- in_toto_run/runlib is now similarly structured as in_toto_record/runlib. See discussion in https://github.com/in-toto/in-toto/issues/6#issuecomment-263988714.


in_toto/in_toto_run.py
```diff
@@ -32,66 +32,63 @@
 import os
 import sys
 import argparse
-import in_toto.util
-import in_toto.runlib
-import in_toto.log as log
+from in_toto import util
+from in_toto import runlib
+from in_toto import log
```
SantiagoTorres: 👍

in_toto/in_toto_run.py
```diff
@@ -102,37 +99,53 @@ def main():
...
+    key = util.prompt_import_rsa_key_from_file(args.key)
+  except Exception as e:
+    log.error("in load key - {}".format(args.key))
+    sys.exit(1)
+
+  in_toto_run(args.step_name, args.materials, args.products,
+      args.link_cmd, key, args.record_byproducts)
 
-if __name__ == '__main__':
+if __name__ == "__main__":
```
SantiagoTorres, line 148: This is the CLI tool right? when we log.error, does it pint to the terminal, or is it silently failing right now?

in_toto/in_toto_run.py
```diff
@@ -102,37 +99,53 @@ def main():
...
+  if args.verbose:
+    log.logging.getLogger().setLevel(log.logging.INFO)
+
+  # We load the key here because it might prompt the user for a password in
+  # case the key is encrypted. Something that should not happen in the library.
+  try:
+    key = util.prompt_import_rsa_key_from_file(args.key)
+  except Exception as e:
+    log.error("in load key - {}".format(args.key))
+    sys.exit(1)
```
SantiagoTorres: Funny quirk (I've never stopped to think about this):

```python
santiago at ~ ✔ bpython
bpython version 0.16 on top of Python 3.5.2 /usr/bin/python
>>> import sys
>>> sys.exit(-1)
(-1,)
santiago at ~ ✗ echo $?
1
santiago at ~ ✔ bpython
bpython version 0.16 on top of Python 3.5.2 /usr/bin/python
>>> import sys
>>> sys.exit(1)
(1,)
santiago at ~ ✗ echo $?
1
```

in_toto/runlib.py
```diff
@@ -310,49 +311,74 @@ def create_link_metadata(link_name, materials_dict={}, products_dict={},
...
+    log.doing("Recording products '{}'...".format(", ".join(product_list)))
+  products_dict = record_artifacts_as_dict(product_list)
+
+  log.doing("Creating link metadata...")
+  link = create_link_metadata(name, materials_dict, products_dict,
       link_cmd_args, byproducts, return_value)
 
+  if key:
+    log.doing("Signing link metadata with key '{:.8}...'...".format(key["keyid"]))
+    link.sign(key)
```
SantiagoTorres, line 383: Hmm, I wonder if we want to try to load the key before we actually run anything...

test/test_in_toto_run.py
```diff
@@ -0,0 +1,161 @@
...
+from in_toto.util import (generate_and_write_rsa_keypair,
+    prompt_import_rsa_key_from_file)
+
+from in_toto.models.link import Link
+from in_toto.in_toto_run import main as in_toto_run_main
+from in_toto.in_toto_run import in_toto_run
+from in_toto.runlib import FILENAME_FORMAT
+
+
+
```
SantiagoTorres: Minor nit: this is too much whitespace isn't it?

test/test_in_toto_run.py
```diff
@@ -0,0 +1,161 @@
...
+  def test_main_optional_args(self):
+    """Test CLI command with optional arguments. """
+
+    args = [ "in_toto_run.py", "--step-name", self.test_step, "--key",
+        self.key_path, "--materials", self.test_artifact, "--products",
+        self.test_artifact, "--record-byproducts", "--", "echo", "test"]
+
+    with patch.object(sys, 'argv', args):
+      in_toto_run_main()
```
SantiagoTorres, line 94: I know that checking that there was a file written or not is better than checking that a link metadata object is not returned, but maybe we could do both?

test/test_runlib.py
```diff
@@ -216,6 +217,75 @@ def test_exclude_subsubdir(self):
     self.assertListEqual(sorted(artifacts_dict.keys()),
         sorted(["foo", "bar", "subdir/foosub1", "subdir/foosub2"]))
 
+class TestInTotoRecordRun(unittest.TestCase):
+  """"Test in_toto_run(name, material_list, product_list,
+  link_cmd_args, key=False, record_byproducts=False). """
```
SantiagoTorres: I'm not sure if this type of comment is descriptive enough.

test/test_toto_run.py
```diff
@@ -1,17 +0,0 @@
...
-
-class Test(unittest.TestCase):
-
-  @unittest.skip("Missing test implementation")
-  def test_unimplemented(self):
-    raise self.assertFalse("Missing test implementation!")
-
-# Run unit test.
-if __name__ == '__main__':
-  unittest.main()
```
SantiagoTorres: 👍

SantiagoTorres (changes requested):
Minor details/questions on this one. Thanks!

demo/run_demo.py
```diff
@@ -128,6 +128,7 @@ def main():
       "final_product/foo.tar.gz",
       "final_product/package.link",
       "final_product/write-code.link",
+      "final_product/tar.link",
```
lukpueh: We didn't dump inspections before, ergo there was no `tar.link`.  `verifylib. run_all_inspections` now calls `runlib.in_toto_run` which also dumps the metadata file, which we wanted to add for inspections (auditing reasons) anyway. Therefor, I also remove this file when clearing the demo final_product.

in_toto/in_toto_run.py
lukpueh: Apparently `exit` only uses the least significant 8 bits, i.e. [0-255].

in_toto/in_toto_run.py
lukpueh: Is this question related to the quotes in L150? 

Whether `log.error` prints or not depends on the logging level specified in `settings.py` and if the `--verbose` cli option is specified, which overrides the level in `settings.py` to print everything greater than `logging.INFO`.

in_toto/runlib.py
lukpueh: That's what we currently do. The key is loaded in the command line tool and passed to `runlib.in_toto_run`. See [in_toto_run#L139-L145](https://github.com/in-toto/in-toto/blob/87e3ddee7442fdec058ef44fe523108fdc1a013b/in_toto/in_toto_run.py#L139-L145)

test/test_in_toto_run.py
lukpueh: Yeah, I guess it is. Will change.

test/test_runlib.py
lukpueh: Hm. It is the TestCase level description. All test functions of this TestCase test the in_toto_run function. How it is tested is explained in each test function's docstring. But I can elaborate if you want me to.

test/test_in_toto_run.py
lukpueh: Yep. I can change this.

in_toto/runlib.py
SantiagoTorres: Are you sure? if we pass a path here to a key that does not exist (and maybe is not called using in-toto-run, we will run the command and then fail after we fail to load the key.

test/test_runlib.py
SantiagoTorres: what it feels like is as if you copied the function's signature for reference or something. Do you think that having the parameters in here helps at all?

in_toto/runlib.py
lukpueh: Correct. We should check on top of the function that the key is properly formatted.

test/test_runlib.py
lukpueh: Correct. I copied the signature to know what valid/invalid arguments I should test the function with. :)

SantiagoTorres: lol, I do the same thing :P

The way I see it, we could either remove them, or list them and say what a "valid" range of each parameter would look like (I think the former is less of a hassle tho)

in_toto/in_toto_run.py
lukpueh: Maybe you want to add you thoughts on this to https://github.com/in-toto/in-toto/issues/6?

test/test_runlib.py
lukpueh: Actually I did this in all my TestCases's class level docstrings so far. Is your concern worth a ticket? :)

SantiagoTorres: Nah, it's definitely not a big deal :)

test/test_in_toto_run.py
lukpueh: Just realized that this test actually calls the command `in-toto-run`, which does not return a link metadata object. Only the library function `runlib.in_toto_run` does.

demo/run_demo.py
lukpueh: No fix needed

in_toto/in_toto_run.py
lukpueh: No fix needed

in_toto/in_toto_run.py
lukpueh: No fix needed

in_toto/runlib.py
lukpueh: Fixed in https://github.com/in-toto/in-toto/pull/60/commits/f94843c4cf01ac057dec1a5b19a45a2648fb55de

test/test_in_toto_run.py
lukpueh: No fix needed

in_toto/in_toto_run.py
SantiagoTorres: my question is: what happens if the logger's handler is a file handler?

lukpueh: Well, right now the logger does not have file handler. And it is on us to decide how the logger is used. :)

I'd like to improve logging/user feedback in general (cf. https://github.com/in-toto/in-toto/issues/6). E.g. properly separate logging (which could use any handler) and user feedback (which must print to the terminal). We have discussed this before and decided that python's `logging` module can be used for both (given proper namespacing).

SantiagoTorres (approved):

PR #62, opened 2017-01-12
Enhance separation between in-toto-verify cli tool and library

lukpueh:
Moves complete in_toto_verify workflow to verifylib.in_toto_verify

The user-facing in-toto-verify command line tool just takes care of command line argument parsing and calls verifylib.in_toto_verify. In case an exception occurs it is caught in the calling CLI tool, which exits gracefully.

- Adds an optional --verbose flag to in-toto-verify CLI tool that turns on logging statements in the library.
- Enhances logging statements in verifylib.
- Changes CLI argument parsing for multiple layout keys.
- Adds a copy of metadata files from in_toto demo to unittest directory for verification testing
- in_toto_verify/verifylib is now similarly structured as in_toto_record/runlib. See discussion in #6 (comment).

- Takes over https://github.com/in-toto/in-toto/pull/40


in_toto/verifylib.py
```diff
@@ -593,4 +593,84 @@ def verify_all_item_rules(items, links, target_links=None):
...
 
 
+def in_toto_verify(layout_path, layout_key_paths):
+  """
+  <Purpose>
+    Does entire in-toto supply chain verification of a final product
+    by performing the following actions:
+        1.  Load root layout as specified by 'layout_path'
+        2.  Load root layout signing key(s) (project owner public key(s)) as
+            specified by 'layout_key_paths'
```
SantiagoTorres, line 603: layout_key_paths is a list? why is it plural?

in_toto/verifylib.py
```diff
@@ -593,4 +593,84 @@ def verify_all_item_rules(items, links, target_links=None):
...
+        3.  Verify layout signature(s)
+        4.  Verify layout expiration
+        5.  Load link metadata for every Step defined in the layout
+            NOTE: in_toto_verify searches for link metadata files by their step
+            name plus the extension '.link' in the current directory.
+            E.g. a step of name 'foo' is expected to be found in the current
+            directory at 'foo.link'.
```
SantiagoTorres, line 610: recall that we now use part of the keyid as a postfix (we could glob this out).

in_toto/verifylib.py
```diff
@@ -593,4 +593,84 @@ def verify_all_item_rules(items, links, target_links=None):
...
+        6.  For link metadata signatures
```
SantiagoTorres: should the following steps be indented?

in_toto/verifylib.py
```diff
@@ -593,4 +593,84 @@ def verify_all_item_rules(items, links, target_links=None):
...
+            List of paths to project owner public keys, used to verify the
+            layout's signature.
+
+  <Exceptions>
+    TBA (see https://github.com/in-toto/in-toto/issues/6)
+
+  <Side Effects>
+    Reads layout from disk
+    Reads project owner public key(s) from disk
+    Reads link metadata files from disk
```
SantiagoTorres: Maybe I'm wrong, but is reading a side effect? (I don't know what's the standard for this).

test/test_in_toto_verify.py
```diff
@@ -0,0 +1,157 @@
...
+    - in_toto_verify's main() - requires sys.argv patching;
+    - in_toto_verify - calls verifylib.in_toto_verify and error logs/exits
+      in case of a raised Exception.
+
+  Uses in-toto demo supply chain link metadata files and basic layout for
+  verification:
+
+  Copies the basic layout for different test scenarios:
+    - single-signed layout
+    - double-signed layout
```
SantiagoTorres: I don't know if I'm missing something. What do we mean by double-signed layout?

test/test_verifylib.py
```diff
@@ -584,6 +588,157 @@ def test_pass_verify_all_inspection_rules(self):
...
+  def test_verify_failing_link_metadata_files(self):
+    """Test fail verification with link metadata files not found. """
+    os.rename("package.link", "package.link.bak")
+    with self.assertRaises(IOError):
+      in_toto_verify(self.layout_single_signed_path, [self.alice_path])
+    os.rename("package.link.bak", "package.link")
+
+  def test_verify_failing_inspection_exits_non_zero(self):
+    """FIXME implement raise exception on inspection fail. """
+    pass
```
SantiagoTorres, line 730: Is this missing now? is there a ticket for this?

SantiagoTorres (commented):
I've just got a couple of questions, but I don't think any of these is a blocker.

in_toto/verifylib.py
lukpueh: `layout_key_paths` is a list that contains paths to public keys whose private portions were used to sign the layout. A layout can be signed by multiple keys, right?

in_toto/verifylib.py
lukpueh: Not yet. The keyid postfix will be in PR #63 or #64. :)  Which will also have to update this function (plus docstrings).

in_toto/verifylib.py
lukpueh: I'm never sure about these either. Just talked about it with @vladimir-v-diaz. He partially agrees with you. Then again, one could argue that reading modifies a file's `atime`.    Anyway, I guess it should be noted but it does not matter if it is called side effect or not, so I can move these operations to the `<Purpose>` section.

in_toto/verifylib.py
lukpueh: No, but the sentence seems to be incomplete. Thanks for spotting that something is odd.

test/test_in_toto_verify.py
lukpueh: Maybe this is not in the specification, but the current implementation allows to sign (and verify) a layout with multiple keys (see also https://github.com/in-toto/in-toto/pull/62#discussion_r96055667)

test/test_verifylib.py
lukpueh: Correct and correct. We discussed this with the VCS AppSec group, since they would have appreciated mentioned behavior. They planned to submit a ticket but never did.

in_toto/verifylib.py
SantiagoTorres: ping

in_toto/verifylib.py
SantiagoTorres: I'd advice to just nix it, but I"m not married with the idea

test/test_in_toto_verify.py
SantiagoTorres: This is fine (it's not specified yet), and I don't see it as a blocker. I think it'd be clearer if we called this (multiple-signatures-on-layout) and then specify it's 2 to convey the fact that it supports multiple signature, not just one case or the other.

test/test_verifylib.py
SantiagoTorres: hmm, could we ticketize this then?

lukpueh: https://github.com/in-toto/in-toto/issues/63

PR #64, opened 2017-01-17
Adds unit tests for util functions

lukpueh:
`in_toto/util.py` mostly consists of interface function that use ssl_crypto to create and load encrypted and non-encrypted RSA keys. This PR adds unit tests for these functions.

Additionally, it fixes a whitespace, unifies quotes and adds a format check in `util.py`

### DIFF
--- a/in_toto/util.py
+++ b/in_toto/util.py
@@ -103,6 +103,10 @@ def import_rsa_public_keys_from_files_as_dict(filepaths):
   key_dict = {}
   for filepath in filepaths:
     key = import_rsa_key_from_file(filepath)
+    #FIXME: Add public key format check to ssl_crypto formats
+    if key["keyval"].get("private"):
+      raise in_toto.ssl_commons.exceptions.FormatError(
+          "Public keys should not have a private portion.")
     keyid = key["keyid"]
     key_dict[keyid] = key
   return key_dict
--- a/in_toto/util.py
+++ b/in_toto/util.py
@@ -41,8 +41,8 @@ def generate_and_write_rsa_keypair(filepath, password=None):
 
   rsa_key = in_toto.ssl_crypto.keys.generate_rsa_key()
 
-  public = rsa_key['keyval']['public']
-  private = rsa_key['keyval']['private']
+  public = rsa_key["keyval"]["public"]
+  private = rsa_key["keyval"]["private"]
 
   if password:
     private_pem = in_toto.ssl_crypto.keys.create_rsa_encrypted_pem(private,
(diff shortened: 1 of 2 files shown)


in_toto/util.py
```diff
@@ -103,7 +103,11 @@ def import_rsa_public_keys_from_files_as_dict(filepaths):
   key_dict = {}
   for filepath in filepaths:
     key = import_rsa_key_from_file(filepath)
-    keyid = key['keyid']
+    #FIXME: Add public key format check to ssl_crypto formats
+    if key["keyval"].get("private"):
```
SantiagoTorres, line 107: not sure if this is functionally equivalent but:

```python
if "private" in key["keyval"]:
    # blow up
```
May be better because it only checks for membership, it doesn't get the actual thing out of it

SantiagoTorres (commented):
I just have one *comment* which I don't think it's a blocker or even worth fixing. Let me know if you'd like to change it before I merge.

in_toto/util.py
lukpueh: You are absolutely right. The thing is that [`KEYVAL_SCHEMA`](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/formats.py#L215-L218) requires us to always have a dict key named "private", even if it is empty.
I plan on submitting a PR for a `PUBKEYVAL_SCHEMA` that does what you propose, which would also replace this inline check.

vladimir-v-diaz:
According to the KEYVAL_SCHEMA you posted, the "private" field is
optional.  You should be able to load an object of this type without a dict
key named "private".

$ python
Python 2.7.6 (default, Oct 26 2016, 20:30:19)
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import securesystemslib.formats
>>> keyval = {'public': "123"}
>>> securesystemslib.formats.KEYVAL_SCHEMA.matches(keyval)
True
>>> keyval = {'private': "456"}
>>> securesystemslib.formats.KEYVAL_SCHEMA.matches(keyval)
False
>>>


--
vladimir.v.diaz@gmail.com
PGP fingerprint = ACCF 9DCA 73B9 862F 93C5  6608 63F8 90AA 1D25 3935
--

SantiagoTorres:
I'm going to merge this. We can check the nuances of this later down the line. Thanks @vladimir-v-diaz and @lukpueh !

lukpueh:
@vladimir-v-diaz, you are right. I guess the "problem" is [`import_rsakey_from_public_pem`](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/keys.py#L1176) then, which adds an empty string to the private portion of the key.

PR #65, opened 2017-01-20
Updates to securesystemslib==0.10.04 (stand-alone)

### LINKED ISSUE #45: Replace ssl_crypto and ssl_commons submodules with securesystemslib
Make in-toto use the tip of [secure-systems-lab/ssl_crypto](https://github.com/secure-systems-lab/ssl_crypto) and [secure-systems-lab/ssl_commons](https://github.com/secure-systems-lab/ssl_commons) as submodules.

The tip of in-toto currently uses [vladimir-v-diaz/ssl_crypto@542caed](https://github.com/vladimir-v-diaz/ssl_crypto/tree/optional-password) as crypto submodule. The common package has not been sub-modularized yet at all.

This issue can be broken down into:
 - Make sure that [secure-systems-lab/ssl_crypto](https://github.com/secure-systems-lab/ssl_crypto) and [secure-systems-lab/ssl_commons](https://github.com/secure-systems-lab/ssl_commons) are up-to-date with what in-toto requires.
 - Replace package in-toto ssl_commons with submodule
 - Adapt in-toto to work with changes that were made in the submodules

lukpueh:
Fixes https://github.com/in-toto/in-toto/issues/45
- removes sub-packages `ssl_crypto` and `ssl_commons`
  they were merged into the stand-alone package `securesystemslib`
- updates requirements.txt, requirements-ci.txt and setup.py
  installation of `pycrypto` and `cryptography` is now handled by `securesystemslib`'s  
  requirements files
- updates in_toto demo README.md
- updates import statements
- updates` import_rsakey_from_pem` and `format_rsakey_from_pem`
  to` import_rsakey_from_private_pem` and `import_rsakey_from_public_pem`

**Noteworthy:**
`securesystemslib` used to be part of [`The Update Framework`](https://github.com/theupdateframework/tuf). When we started to implement `in-toto` we sought ways to use the common modules for both projects. We reviewed various approaches/tools and eventually decided to take out the common code from TUF, place it into a separate repository and re-integrate it as submodule in both projects.

This involved replacing absolute imports, e.g. `import (tuf|in_toto).securesystemslib.keys`, with relative imports in order to make `securesystemslib` agnostic to the parent project. Furthermore it forced the maintainers to learn new facets of git, e.g. cloning recursively.

The main **drawbacks** were employing new git commands and not being able to have `securesystemslib` as a stand-alone package and the benefits that come with it, e.g. requirements handling, established test setups, ...).

The **advantage** was that TUF and in_toto could henceforth be released without an additional external dependency, which some platforms (e.g. debian) would require us to vendor, especially in this case where `securesystemslib` is a mostly internal library. Not having to vendor plus the use of git submodules allowed us to easily sync with or report back developments to `securesystemslib`.

The pros and cons led to a conflict between the maintainers of in-toto and TUF. It was decided to create a stand-alone package. Unfortunately, there is no sensible way so that TUF can use `securesystemslib` as stand-alone package and in-toto use it as submodule.
- `git subtree` boils down to hard copying the files with the [feature](http://git.661346.n2.nabble.com/subtree-merges-lose-prefix-after-rebase-td7332850.html) of breaking `git rebase`
- Hard-copying the files would work, if the import statements  inside of `securesystemslib` hadn't been changed to require `securesystemslib` to be on the path (which is normal for stand-alone packages).
- If we decide to vendor, which we will eventually do, we will have to either patch the path (messy) or the import statements (messy)


in_toto/models/common.py
```diff
@@ -102,6 +102,7 @@ def verify_signatures(self, keys_dict):
       except:
```
vladimir-v-diaz: We should avoid blank except statements.

in_toto/models/common.py
```diff
@@ -102,6 +102,7 @@ def verify_signatures(self, keys_dict):
...
         raise SignatureVerificationError(
             "Signature key not found, key id is '{0}'".format(keyid))
-      if not keys.verify_signature(key, signature, self.payload):
+      if not securesystemslib.keys.verify_signature(
```
vladimir-v-diaz, line 108: This can potentially raise an exception. For example, `securesystems.exceptions.UnknownMethodError` is raised if `key` specifies an unsupported key type.  I guess this function assumes these arguments are valid, or verified elsewhere? (Perhaps assumptions of these sort should be stated in the docstring.)

lukpueh: That's actually fine. The method is expected to raise an exception if the signature does not verify (for whatever reason).

in_toto/matchrule_validators.py
```diff
@@ -35,40 +36,47 @@ def _validate_match_rule(keywords):
...
 
 def _validate_generic_rule(keywords):
   """ private helper that verifies the syntax of the other rules """
 
   VALID_OPERATIONS = {'CREATE', 'MODIFY', 'DELETE',}
 
   if not isinstance(keywords, list):
-    raise FormatError("this matching rule is not a list")
+    raise securesystemslib.exceptions.FormatError(
+        "this matching rule is not a list")
```
vladimir-v-diaz: I find it helpful if the error message includes the argument/value that is wrong.  When appropriate, examples of valid values also helps.

The following error message, used later in the code, is good:
```Python
raise securesystemslib.exceptions.FormatError("{} is not a valid rule!".format(rule))
```

in_toto/models/common.py
vladimir-v-diaz: Ahh okay.  You should probably add a doctring specifying which exceptions it raises, arguments, etc.

lukpueh: I am not sure, if the `<Exceptions>` docstring of a function should also list the possible exceptions of the functions it is calling.

in_toto/matchrule_validators.py
lukpueh: Agreed.

in_toto/exceptions.py
```diff
@@ -1,4 +1,4 @@
-from in_toto.ssl_commons.exceptions import Error
+from securesystemslib.exceptions import Error
```
vladimir-v-diaz, line 1:
```Python
class RuleVerficationFailed(Error):
```
The exception name here is inconsistent with the others.  Should it instead be `RuleVerificationFailedError`?

in_toto/models/common.py
```diff
@@ -99,9 +99,10 @@ def verify_signatures(self, keys_dict):
       keyid = signature["keyid"]
       try:
         key = keys_dict[keyid]
-      except:
+      except KeyError:
```
vladimir-v-diaz, line 105: This method doesn't validate it's arguments, so I'm not sure where/when/how its done, but `keys_dict[keyid]` can potentially raise `TypeError` if `keyid` happens to contain an unhashable type.

```Python
>>> dict1 = {}
>>> dict1[{'foo': 1}]
TypeError: unhashable type: 'dict'
```

in_toto/models/layout.py
```diff
@@ -245,45 +253,51 @@ def read(data):
...
 
   def _validate_threshold(self):
     """Private method to check that the threshold field is set to an int."""
     try:
       # int(self.threshold)
       # FIXME: we don't have support for threshold yet
       pass
-    except:
-      raise FormatError("Invalid threshold value for this step")
+    except Exception:
```
vladimir-v-diaz: Is this incomplete or a WIP?

```Python
def _validate_threshold(self):
    """Private method to check that the threshold field is set to an int."""
    try:
      # int(self.threshold)
      # FIXME: we don't have support for threshold yet
      pass
    except Exception:
"Invalid threshold value for this step")
```

in_toto/models/layout.py
```diff
@@ -245,45 +253,51 @@ def read(data):
...
+      raise securesystemslib.exceptions.FormatError(
```
vladimir-v-diaz, line 264: Debugging becomes difficult when you catch `Exception` and raise your own without logging or printing the original exception.

in_toto/models/layout.py
```diff
@@ -320,25 +334,29 @@ def read(data):
   def _validate_type(self):
     """Private method to ensure that the type field is set to inspection."""
     if self._type != "inspection":
-      raise FormatError("The _type field should be aset to inspection!")
+      raise securesystemslib.exceptions.FormatError(
+          "The _type field should be aset to inspection!")
```
vladimir-v-diaz: typo

requirements-ci.txt
```diff
@@ -1,5 +1,3 @@
-pycrypto
-cryptography
 pytest
 PyNaCl
```
vladimir-v-diaz: PyNaCl is also a dependency of the securesystemslib library.

vladimir-v-diaz:
I do not see any other issues with the securesystemslib bits.

in_toto/models/layout.py
```diff
@@ -148,58 +147,67 @@ def import_step_metadata_from_files_as_dict(self):
...
+          "keys dictionary is malformed!")
 
-    in_toto.ssl_crypto.formats.KEYDICT_SCHEMA.check_match(self.keys)
+    securesystemslib.formats.KEYDICT_SCHEMA.check_match(self.keys)
 
     for keyid, key in six.iteritems(self.keys):
       if 'private' in key and key['private'] != '':
-        raise FormatError("key: {} contains a private key part!".format(keyid))
+        raise securesystemslib.exceptions.FormatError(
+            "key: {} contains a private key part!".format(keyid))
```
vladimir-v-diaz: v0.10.4 now contains the schema for public keyvals:
https://github.com/secure-systems-lab/securesystemslib/releases/tag/v0.10.4
https://pypi.python.org/pypi?:action=display&name=securesystemslib&version=0.10.4

in_toto/models/layout.py
lukpueh: This is WIP. @RohanBhirangi is working on adding the threshold field.

in_toto/exceptions.py
lukpueh: Fixed in 2f39a18

in_toto/models/common.py
lukpueh: Fixed in ed60dd7

in_toto/models/layout.py
lukpueh: I updated to 0.10.4 in https://github.com/in-toto/in-toto/pull/65/commits/df25016ff32603935788b5a662d648a94256609e.
Unfortunately I can't rename the branch without loosing this GitHub discussion but I updated the PR title.

lukpueh:
@vladimir-v-diaz, thanks for the review! I added your suggestion to the big [user feedback/exception messages issue](https://github.com/in-toto/in-toto/issues/6#issuecomment-274845182), because user it should be improved all over the place.

in_toto/matchrule_validators.py
```diff
@@ -25,7 +25,8 @@ def _validate_match_rule(keywords):
 
   if not isinstance(keywords, list):
     raise securesystemslib.exceptions.FormatError(
-        "this matching rule is not a list")
+        "Wrong rule format, must be a list.\n\t"
```
vladimir-v-diaz, line 28: Not part of this PR, but the <Module Name> entry in the header block is wrong.

vladimir-v-diaz: Also not part of this PR, but our lab projects should choose a single way to wrap long lines.

I spoke with @awwad about this, but we think indenting wrapped lines looks best.  For example,

```Python
raise FormatError("Argument"
  " is improperly formatted.  Valid"
  " format of argument is X.")
```
Note, that's 2 spaces per indentation.

I've used the following in the past, but it's more difficult to read and edit:

```Python
raise FormatError("Argument is improperly"
                  " formatted.  Valid format"
                  " of argument is X.")
```

lukpueh: Re header block: Thanks, will fix.
Re indentation: I operate according to our code style guide lines here, which suggests [double-indentation for line continuation](https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length)

vladimir-v-diaz: That's correct, I forgot that we decided to use double-indentation.

in_toto/matchrule_validators.py
```diff
@@ -1,6 +1,6 @@
 """
 <Module Name>
-  in_toto/models/validators.py
+  in_toto/models/matchrule_validators.py
```
vladimir-v-diaz: I don't find a `matchrule_validators.py` in the `in-toto/models` directory.

lukpueh: Sorry! `git --amend` to the rescue. :)

in_toto/matchrule_validators.py
```diff
@@ -37,46 +38,51 @@ def _validate_match_rule(keywords):
...
+      " should be: MATCH (MATERIAL | PRODUCT)"
       " <path_pattern> [AS <target_path_pattern>] FROM <step>.\n\t"
-      "Got: {}".format(" ".join(keywords)))
+      "Got: {}".format(keywords))
 
   if rule != "MATCH" and rule.upper() != "MATCH":
-    raise securesystemslib.exceptions.FormatError(
-        "Wrong rule to verify! {}".format(rule))
+    raise securesystemslib.exceptions.FormatError("Wrong rule format.\n\t"
+        "Got: {}".format(keywords))
```
vladimir-v-diaz, line 48: The user could probably figure it out on their own, but a more helpful exception would emphasize that the rule keyword is wrong (i.e., must be `MATCH` or `match`).

in_toto/matchrule_validators.py
```diff
@@ -37,46 +38,51 @@ def _validate_match_rule(keywords):
...
-    raise securesystemslib.exceptions.FormatError(
-        "this matching rule is not a list")
+    raise securesystemslib.exceptions.FormatError("Wrong rule format,"
+        " rule must be a list.\n\t"
+        "Got: {}".format(keywords))
 
   if len(keywords) != 2:
-    raise securesystemslib.exceptions.FormatError("Wrong rule format")
+    raise securesystemslib.exceptions.FormatError("Wrong rule format.\n\t"
+        "Got: {}".format(keywords))
```
vladimir-v-diaz, line 79: It would help the user more if the exception message emphasized that len(keywords) must be 2, like you do with the others.

vladimir-v-diaz:
Okay, feel free to merge when you're done with the remaining comments.

PR #66, opened 2017-01-24
Removes external package simple_settings

lukpueh:
in-toto used to use simple_settings to provide a global settings
file that could also be accessed from submodules without specifying
the full path. Since submodules were removed `simple_settings` is
no longer needed.

The former submodule and now external dependency `securesystemslib`
has its own settings file, that should henceforth be accessed
programmatically.

E.g.:
```python
import securesystemslib.settings
securesystemslib.settings.RSA_CRYPTO_LIBRARY = "pyca-cryptography" # default
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,6 @@ cd in-toto
 
 # Install with pip in "develop mode"
 pip install -e .
-
-# Export the envvar required for "simple settings"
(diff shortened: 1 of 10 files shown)


SantiagoTorres:
LGTM 👍

PR #68, opened 2017-01-25
Raise exception on non-zero return of Inspection (fix #63)

### LINKED ISSUE #63: Raise exception on non-zero return of Inspection command
`Inspection` commands defined in the layout provide the possibility to execute arbitrary commands at verification time.
Inspections, like Steps, must have rules to authorize/enforce materials and products that are recorded before and after an Inspection command is run. Currently a failing rule is the only way to make an Inspection fail the verification.
Additionally, it might be desired to simply fail verification when the inspection command returns a non-zero value.

lukpueh:
Fixes https://github.com/in-toto/in-toto/issues/63

- Adds internal function that checks return values of shell commands, e.g. from inspections. Raises exception if the passed value is non-int and non-zero.
- Adds call to function in `verifylib.run_all_inspections`, i.e. verification aborts if a ran inspection returns non-zero.
- Fixes typos in comments
- Adds unittests

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -37,9 +37,48 @@ import in_toto.util
 import in_toto.runlib
 import in_toto.models.layout
 import in_toto.models.link
-from in_toto.exceptions import (RuleVerficationError, LayoutExpiredError)
+from in_toto.exceptions import (RuleVerficationError, LayoutExpiredError,
+    BadReturnValueError)
 from in_toto.matchrule_validators import check_matchrule_syntax
 import in_toto.log as log
+
+def _raise_on_bad_retval(return_value, command=None):
+  """
+  <Purpose>
+    Internal function that checks return values of shell commands, e.g. from
+    inspections. Raises exception if the passed value is non-int and non-zero.
+
+  <Arguments>
+    return_value:
+            The return value to be verified
+    command: (optional)
+            The command whose execution returned the value, used for exception
+            message.
+
+  <Exceptions>
+    BadReturnValueError if the return_value is non-int and non-zero
+
+  <Side Effects>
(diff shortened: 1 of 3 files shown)


in_toto/verifylib.py
```diff
@@ -37,10 +37,49 @@
 import in_toto.runlib
 import in_toto.models.layout
 import in_toto.models.link
-from in_toto.exceptions import (RuleVerficationError, LayoutExpiredError)
+from in_toto.exceptions import (RuleVerficationError, LayoutExpiredError,
+    BadReturnValueError)
 from in_toto.matchrule_validators import check_matchrule_syntax
 import in_toto.log as log
 
+def _raise_on_bad_retval(return_value, command=None):
```
SantiagoTorres, line 45: Just out of curiosity. This private method does not require documentation does it? (Not that I mind)

in_toto/verifylib.py
```diff
@@ -37,10 +37,49 @@
...
+  if command:
+    msg = "{0} from command '{1}'.".format(msg, command)
+  else:
+    msg = "{0}.".format(msg)
+
+  if not isinstance(return_value, int):
+    raise BadReturnValueError(msg.format(what="int"))
+
+  # TODO: in-toto specification suggests special behavior on
+  # return_value < 127, but does not fully define that behavior yet
```
SantiagoTorres: I think this comment means to say "return value == 127"

in_toto/verifylib.py
lukpueh: According to our code-style-guidelines its not required. But I guess it does not hurt either.

SantiagoTorres:
Thanks! 👍

PR #69, opened 2017-01-27
Add key threshold

RohanBhirangi:
# Adds threshold field
- added threshold field in the layout model
- added validate function to verify if threshold is an integer
- changed naming scheme of link metadata files from [name].link
to [name].[keyid].link
- changed the dump function so that it processes key as parameter
- modified the existing link data structure to accommodate
the threshold field
- modified several functions to process the new link data structure
- added threshold verification function and threshold verification
exception
- modified the demo application according to the new naming
scheme (of link objects) and threshold field


SantiagoTorres:
ping @RohanBhirangi , are you going to push changes to this branch? or should I start reviewing? (I ask bc the CI build failed).

RohanBhirangi:
@SantiagoTorres @lukpueh I have modified the unit tests and pushed my changes. You can start reviewing them.

demo/owner_alice/create_layout.py
```diff
@@ -61,6 +63,6 @@ def main():
   # Sign and dump layout to "layout.root"
   layout.sign(key_alice)
   layout.dump()
-
+
```
lukpueh: Please remove trailing whitespace.

demo/run_demo.py
```diff
@@ -34,6 +34,13 @@ def supply_chain():
   subprocess.call(write_code_cmd.split())
   copyfile("foo.py", "../functionary_carl/foo.py")
```
lukpueh: Carl writes his own `foo.py` now, he does not have to copy Bob's `foo.py` anymore. 
In general a key threshold does not really make sense for a `write-code` step. I guess it would make more sense for the `package` step. 
So you example serves for a POC, and we have to overhaul the demo anyway. But maybe you could change the demo to make `package` a threshold  step instead?

in_toto/models/layout.py
```diff
@@ -38,7 +38,7 @@
 from dateutil.parser import parse
 
 import in_toto.matchrule_validators
-import securesystemslib.exceptions
+from securesystemslib.exceptions import FormatError
```
lukpueh: I'm trying to get rid of `from` imports. Please revert this change and use `securesystemslib.exceptions.FormatError` below. See our [code style guidelines](https://github.com/secure-systems-lab/code-style-guidelines#imports) for more information on imports

demo/run_demo.py
lukpueh: Please consider also updating the [demo README](https://github.com/in-toto/in-toto/blob/develop/demo/README.md) to reflect your changes.

in_toto/models/layout.py
```diff
@@ -140,8 +140,11 @@ def import_step_metadata_from_files_as_dict(self):
     """
     step_link_dict = {}
     for step in self.steps:
-      link = models__link.Link.read_from_file(step.name + '.link')
-      step_link_dict[step.name] = link
+      key_link_dict = {}
+      for keyid in step.pubkeys:
+        link = models__link.Link.read_from_file(step.name + '.' + keyid + '.link')
```
lukpueh: The specification says to only use the first six bytes of the keyid in the filename.

lukpueh: We should wrap this call in a `try` `except`  block and pass on `IOError`s. Just because a functionary is authorized to perform a step (is listed in the step's `pubkeys`) does not mean that s/he has to perform the step.

in_toto/models/layout.py
```diff
@@ -140,8 +140,11 @@ def import_step_metadata_from_files_as_dict(self):
...
+        key_link_dict[keyid] = link
+      step_link_dict[step.name] = key_link_dict
```
lukpueh: Please update the function's docstring to reflect your changes.

in_toto/models/layout.py
```diff
@@ -237,6 +240,7 @@ class Step(models__common.Metablock):
   product_matchrules = attr.ib(default=attr.Factory(list))
   pubkeys = attr.ib(default=attr.Factory(list))
   expected_command = attr.ib("")
+  threshold = attr.ib(default=attr.Factory(int))
```
lukpueh: We do not need a factory for simple data types. You can read about why we use them for lists in this [commit message](https://github.com/in-toto/in-toto/commit/192bb35abd1b57521a5579ca57d123cce03d5448). Or take a look at [`attr`'s docs](https://attrs.readthedocs.io/en/stable/api.html).

in_toto/models/layout.py
```diff
@@ -256,13 +261,9 @@ def _validate_type(self):
...
     """Private method to check that the threshold field is set to an int."""
-    try:
-      # int(self.threshold)
-      # FIXME: we don't have support for threshold yet
-      pass
-    except Exception as e:
-      raise securesystemslib.exceptions.FormatError(
-          "Invalid threshold value for this step. Exception: {}".format(e))
+    if type(self.threshold) != int:
+      raise FormatError("Invalid threshold '{}' value for this step."
```
lukpueh: Minor nit: Pleas change to `"Invalid threshold value '{}', must be int."`!

in_toto/models/link.py
```diff
@@ -65,12 +66,19 @@ class Link(models__common.Signable):
   command = attr.ib("")
   return_value = attr.ib(None)
 
-  def dump(self, filename=False):
+  def dump(self, filename=False, key=False):
     """Write pretty printed JSON represented of self to a file with filename.
     If no filename is specified, a filename will be created using the link name
     + '.link'-suffix """
     # Magic: short circuiting and string formatting
```
lukpueh: Remove this comment. The short circuiting magic is gone.

in_toto/models/link.py
```diff
@@ -65,12 +66,19 @@ class Link(models__common.Signable):
...
-    super(Link, self).dump(filename or "%s.link" % self.name)
+    if filename:
+      fn = filename
+    elif key:
+      securesystemslib.formats.KEY_SCHEMA.check_match(key)
+      fn = "{}.{}.link".format(self.name, key['keyid'])
```
lukpueh: Same as above. Only use first 6 bytes of the keyid.

in_toto/models/link.py
```diff
@@ -65,12 +66,19 @@ class Link(models__common.Signable):
...
+    else:
+      fn = "{}.link".format(self.name)
+    super(Link, self).dump(fn)
```
lukpueh: Please update the docstrings and describe your algorithm. E.g. that/why key gets ignored if a filename is specified...

in_toto/runlib.py
```diff
@@ -50,8 +50,8 @@
 import securesystemslib.hash
 import securesystemslib.exceptions
 
-FILENAME_FORMAT = "{step_name}.link"
-UNFINISHED_FILENAME_FORMAT = ".{step_name}.link-unfinished"
+FILENAME_FORMAT = "{step_name}.{keyid}.link"
+UNFINISHED_FILENAME_FORMAT = ".{step_name}.{keyid}.link-unfinished"
```
lukpueh: We should rename this to something like short_keyid to distinguish from the full keyid.

in_toto/runlib.py
```diff
@@ -351,7 +351,10 @@ def in_toto_run(name, material_list, product_list,
     Newly created Link object
   """
 
-  fn = FILENAME_FORMAT.format(step_name=name)
+#  if key:
+#    fn = FILENAME_FORMAT.format(step_name=name, keyid=key["keyid"])
+#  else:
+#    fn = FILENAME_FORMAT.format(step_name=name, keyid=None)
```
lukpueh: Do we still need this?

in_toto/runlib.py
```diff
@@ -382,8 +385,8 @@ def in_toto_run(name, material_list, product_list,
     log.doing("Signing link metadata with key '{:.8}...'...".format(key["keyid"]))
     link.sign(key)
 
-  log.doing("Storing link metadata to '{}'...".format(fn))
-  link.dump()
+    log.doing("Storing link metadata to '{}'...".format(name+"."+key["keyid"]+".link"))
+    link.dump(key=key)
```
lukpueh: You could change this to:
```python
log.doing("Storing link metadata to '{}'...".format(FILENAME_FORMAT.format(step_name=name, keyid=keyid)))
```

in_toto/runlib.py
```diff
@@ -431,8 +434,8 @@ def in_toto_record_start(step_name, key, material_list):
   link.sign(key)
 
   log.doing("Storing preliminary link metadata to '{}'...".format(unfinished_fn))
-  link.dump(unfinished_fn)
-
+  link.dump(filename=unfinished_fn)
+
```
lukpueh: Please delete trailing whitespace.

in_toto/verifylib.py
```diff
@@ -232,15 +233,15 @@ def verify_all_steps_signatures(layout, links_dict):
   """
   for step in layout.steps:
     # Find the according link for this step
-    link = links_dict[step.name]
-
+    key_link_dict = links_dict[step.name]
+    for keyid in key_link_dict:
+      link = key_link_dict[keyid]
     # Create the dictionary of keys for this step
```
lukpueh: Fix comment indentation.

in_toto/verifylib.py
```diff
@@ -232,15 +233,15 @@ def verify_all_steps_signatures(layout, links_dict):
...
-    keys_dict = {}
-    for keyid in step.pubkeys:
-      keys_dict[keyid] = layout.keys[keyid]
-
+      keys_dict = {}
+      for keyid in step.pubkeys:
+        keys_dict[keyid] = layout.keys[keyid]
     # Verify link metadata file's signatures
```
lukpueh: Fix comment indentation.

in_toto/runlib.py
lukpueh: And we should move these constants to `link.py`

in_toto/verifylib.py
```diff
@@ -232,15 +233,15 @@ def verify_all_steps_signatures(layout, links_dict):
...
+    for keyid in key_link_dict:
+      link = key_link_dict[keyid]
     # Create the dictionary of keys for this step
-    keys_dict = {}
-    for keyid in step.pubkeys:
-      keys_dict[keyid] = layout.keys[keyid]
-
+      keys_dict = {}
+      for keyid in step.pubkeys:
+        keys_dict[keyid] = layout.keys[keyid]
```
lukpueh: You don't have to create a step's `keys_dict` for each link. Move this loop out of the direct enclosing loop.

in_toto/verifylib.py
```diff
@@ -232,15 +233,15 @@ def verify_all_steps_signatures(layout, links_dict):
...
     # Verify link metadata file's signatures
-    verify_link_signatures(link, keys_dict)
+      verify_link_signatures(link, keys_dict)
```
lukpueh: Please update the docstrings of the function.

in_toto/verifylib.py
```diff
@@ -232,15 +233,15 @@ def verify_all_steps_signatures(layout, links_dict):
   """
   for step in layout.steps:
     # Find the according link for this step
-    link = links_dict[step.name]
-
+    key_link_dict = links_dict[step.name]
```
lukpueh: Let's call this `link_dict` instead of `key_link_dict` and define its format somewhere (e.g. in docstrings of functions that use this data structure):
```python
{ <keyid> : <Link object>, ...}
```

in_toto/verifylib.py
```diff
@@ -232,15 +233,15 @@ def verify_all_steps_signatures(layout, links_dict):
...
+    for keyid in key_link_dict:
+      link = key_link_dict[keyid]
```
lukpueh: You can do this instead:
```python
  keys_dict = {}
  ...
```

in_toto/verifylib.py
```diff
@@ -300,10 +301,12 @@ def verify_all_steps_command_alignment(layout, links_dict):
   """
   for step in layout.steps:
     # Find the according link for this step
-    link = links_dict[step.name]
-    command = link.command
     expected_command = step.expected_command.split()
-    verify_command_alignment(command, expected_command)
+    key_link_dict = links_dict[step.name]
+    for keyid in key_link_dict:
+      link = key_link_dict[keyid]
```
lukpueh: See comments about [variable name](https://github.com/in-toto/in-toto/pull/69/files#r100394513) and [dictionary iterator](https://github.com/in-toto/in-toto/pull/69/files#r100394666) above.

in_toto/verifylib.py
```diff
@@ -625,7 +628,7 @@ def verify_all_item_rules(items, links, target_links=None):
 
   """
   if not target_links:
-    target_links = links
+      target_links = links
```
lukpueh: We use two spaces per indentation level. Revert change in this line!

in_toto/verifylib.py
```diff
@@ -687,7 +743,7 @@ def in_toto_verify(layout_path, layout_key_paths):
 
   log.doing("Reading layout...")
   layout = in_toto.models.layout.Layout.read_from_file(layout_path)
-
+
```
lukpueh: Remove trailing whitespace.

in_toto/verifylib.py
lukpueh: Hm. I can see now that we will need different types of link dictionaries:
```python
# A dictionary (1) with step names as keys and values that are
# dictionaries (2) with functionary keyids as keys and Link objects as values
  <step name> : {
    <keyid> : <Link object>,
    ...
    },
  ...
}
# A dictionary (3) with step names as keys and Link ojects as values
{
  <step name> : <Link object>,
  ...
}
```

- (1) regular format for all reported links for all items of a software supply chain
   name e.g.:  `chain_link_dict`
- (2) regular format for all reported links for an item (step or inspection)
  name e.g.: `item_link_dict`
- (3) condensed format for all items of a software supply chain, each containing an exemplary link used to verify artifact rules
  name e.g.: `chain_link_dict_condensed`, `chain_link_dict_reduced`, `chain_link_dict_short`, ...?

Any ideas?

in_toto/verifylib.py
```diff
@@ -634,6 +637,59 @@ def verify_all_item_rules(items, links, target_links=None):
...
+            threshold will be verified.
+
+    step_link_dict:
+            A dictionary of key-link pair with step names as keys. For each
+            step name, there are one or more keyids and corresponding
+            link objects.
+
+  <Exceptions>
+    raises an Exception if threshold is not verified.
+    ThresholdVerificationError()
```
lukpueh: ThresholdVerificationError if <what exactly?>

in_toto/verifylib.py
```diff
@@ -634,6 +637,59 @@ def verify_all_item_rules(items, links, target_links=None):
     verify_item_rules(item.name, item.product_matchrules,
         link.products, target_links)
 
+def verify_threshold_equality(layout, step_link_dict):
+  """
+  <Purpose>
+    Iteratively verifies threshold conditions and link artifacts of
+    passed dictionary (step_link_dict).
```
lukpueh: Please describe what "verifies threshold conditions and link artifacts" means

in_toto/verifylib.py
```diff
@@ -634,6 +637,59 @@ def verify_all_item_rules(items, links, target_links=None):
...
+  <Side Effects>
+    None.
+  
+  <Returns>
+    A dictionary containing one Link metadata object per step only if
+    the link artifacts of all link objects are identical for a step.
+  """
+  target_links = {}
+  for key_link in step_link_dict:
+    key_link_dict = step_link_dict[key_link]
```
lukpueh: See comments about [variable name](https://github.com/in-toto/in-toto/pull/69/files#r100394513) and [dictionary iterator](https://github.com/in-toto/in-toto/pull/69/files#r100394666) above.

in_toto/verifylib.py
```diff
@@ -634,6 +637,59 @@ def verify_all_item_rules(items, links, target_links=None):
...
+            step name, there are one or more keyids and corresponding
+            link objects.
+
+  <Exceptions>
+    raises an Exception if threshold is not verified.
+    ThresholdVerificationError()
+
+  <Side Effects>
+    None.
+
```
lukpueh: Remove trailing whitespaces.

in_toto/verifylib.py
```diff
@@ -634,6 +637,59 @@ def verify_all_item_rules(items, links, target_links=None):
...
+    None.
+  
+  <Returns>
+    A dictionary containing one Link metadata object per step only if
+    the link artifacts of all link objects are identical for a step.
+  """
+  target_links = {}
+  for key_link in step_link_dict:
+    key_link_dict = step_link_dict[key_link]
+
```
lukpueh: Remove trailing whitespaces.

in_toto/verifylib.py
```diff
@@ -634,6 +637,59 @@ def verify_all_item_rules(items, links, target_links=None):
...
+        for keyid in key_link_dict:
+          if i == 0:
+            link_dict = key_link_dict[keyid]
+            i = i + 1
+          else:
+            if link_dict.materials != key_link_dict[keyid].materials or link_dict.products != key_link_dict[keyid].products:
+              raise ThresholdVerificationError("Link artifacts do not match!".format())
+      target_links[key_link] = link_dict
+    else:
+      raise ThresholdVerificationError("Step not performed by enough functionaries!".format())
```
lukpueh: I think above algorithm can be simplified and I am not so sure if it does exactly what you intend to do. I am also not sure about what you intend to do, because there are not a lot of comments. :) Here is some general advise:
- Add comments that explain what you want to do
- Make sure that indentation is consistent
- Try to simplify your control flow statements, too much nesting makes it hard to read
- Read up on Python loops and how to use indices

And here is how I would write this function (in comments):
```python
# We are only interested in links that are related to steps defined in the
# Layout, so let's iterate over layout.steps
 
  # Skip steps that don't require multiple functionaries
  ...

  # Extract the step_link_dict for this step from the passed chain_link_dict
  link = chain_link_dict.get(step.name)
  ...

  # Check if we have at least <threshold> links for this step
  ...

  # AFICS `verify_all_steps_signatures` only verifies if each link of all steps
  # was signed by ANY of the functionaries authorized to perform this step.
  # But we also have to make sure we have at least <threshold> UNIQUE links,
  # i.e. signed by different functionaries. We should check this here and we
  # should not use the functionary keyid used as dictionary key, but
  # rather the keyid in the link's signature, because the former is taken from
  # the file name and could be forged.
  ...

  # Finally take one exemplary link (e.g. the first in the step_link_dict) and
  # iterate over the remaining links to compare their properties, above all
  # materials and products have to be equal for all links of a step.
  # I'm not sure if we care about byproducts, command and return value
  # maybe we should discuss this.
  
  # If no exception was raised so far, everything should be fine (the pythonic way)
```

in_toto/verifylib.py
```diff
@@ -634,6 +637,59 @@ def verify_all_item_rules(items, links, target_links=None):
...
+  return target_links
```
lukpueh: I know I told you that this function could also condense/reduce the dictionary that contains all links for each step of the supply chain to a dictionary with one link per step which will be used to verify the rules.

Now that I think of it, I'd say it is nicer  to write an extra function for this (separation of concerns trumps performance).

test/test_runlib.py
```diff
@@ -433,10 +431,14 @@ def test_missing_unfinished_file(self):
   def test_wrong_signature_in_unfinished_metadata(self):
     """Test record stop exits on wrong signature, no link recorded. """
     in_toto_record_start(self.step_name, self.key, [])
+    link_name = str("."+ self.step_name + "." + self.key["keyid"] + ".link-unfinished")
+    changed_link_name = str("."+ self.step_name + "." + self.key2["keyid"] + ".link-unfinished")
+    os.rename(link_name, changed_link_name)
```
lukpueh: I think you can use `UNFINISHED_FILENAME_FORMAT` here.

lukpueh (changes requested):
Thanks for the contributions, @RohanBhirangi! Overall it is a lovely PR. Don't let my comments discourage you. Most of them are easy to fix. And tell me if you disagree with something.

demo/run_demo.py
RohanBhirangi: We had talked about this earlier. If you remember, the problem was that the 2 foo.tar.gz files had different SHA codes. So we decided to make write-code a threshold step.

lukpueh: I think we should work around this by omitting the `z` flag from the `tar` command (just for the sake of the demo). The `z` makes `tar` use `gzip` to compress the archive, which in turn also stores filename and timestamp, in therefore leads to different hashes (see [SO post](http://stackoverflow.com/questions/36464358/why-do-the-md5-hashes-of-two-tarballs-of-the-same-file-differ)).

in_toto/models/layout.py
RohanBhirangi: When I use "import securesystemslib.exceptions.FormatError" instead of "from securesystemslib.exceptions import FormatError",
it gives me "ImportError: No module named FormatError"
Any ideas?

lukpueh: `FormatError` is not a module, it is a class of the module `securesystemslib.exceptions`.
So you have import the module:
```python
import securesystemslib.exceptions
```
and then use the class specifying the full path, e.g.
```python
raise securesystemslib.exceptions.FormatError("This is a format error!")
```

in_toto/models/layout.py
RohanBhirangi: Is there any way to change this to get only 6 bytes of just the keyid?
FILENAME_FORMAT.format(step_name=self.step_name, keyid=self.key["keyid"]))
This gets the entire keyid!

Or can we make changes to FILENAME_FORMAT = "{step_name}.{keyid}.link" to enable only the first 6 bytes?

RohanBhirangi: I resolved this! 👍

lukpueh: Great, I suppose you found [this line](https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L382) or located the relevant part in the [format specification mini language](https://docs.python.org/2/library/string.html#format-specification-mini-language)?

> The precision is a decimal number indicating how many digits should be displayed after the decimal point for a floating point value formatted with 'f' and 'F', or before and after the decimal point for a floating point value formatted with 'g' or 'G'. For non-number types the field indicates the maximum field size - in other words, how many characters will be used from the field content. The precision is not allowed for integer values.

in_toto/models/layout.py
RohanBhirangi: Can this be used instead " threshold = attr.ib(default=0) " ?

lukpueh: I do believe so yes (according to `attr`'s doc).

lukpueh:
Thanks for addressing some of my comments (https://github.com/in-toto/in-toto/pull/69/commits/5a85686a2eee9ab60d7f97ea15b3796019f88f33, https://github.com/in-toto/in-toto/pull/69/commits/5cd584d1a6aa40fe9dbe094dee72e6794dcf6318)! Is this PR ready for another round of review?

For future reference, please make sure that the commit messages list all (important) changes (plus a little bit of context) for instance here is what above two commits did (minus context) apart from what you indicated in the message:

**fixed layout.py and link.py**
- Added a bunch of demo files (I guess by accident, because they are gone in the subsequent commit)
- Modified demo `create_layout.py`
- Modified imports
- Removed factory in layout model threshold property 
- Changed link filename format 

**Fixed demo README**
- Removed demo files
- Moved `key_link_dict` and `keys_dict` initialization in `verify_all_steps_signatures` (btw. this is not what I asked for in my [review comment](https://github.com/in-toto/in-toto/pull/69#discussion_r100393391))

demo/run_demo.py
RohanBhirangi: As we discussed, I have kept the demo as it was earlier. Just changed a few minor things like the link names ([step-name].[keyid].link)

in_toto/models/layout.py
RohanBhirangi: I think it would make more sense to keep the default threshold value = 1.
As any step should at least be performed by one functionary.
**Already changed the default value. Tell me if you disagree.

in_toto/verifylib.py
RohanBhirangi: For now, I have used the following names for the dicts :-
1. "chain_link_dict" for the whole dict with step-names as keys and values that are
2. "key_link_dict" for the inner dict with functionary keyids as keys and Link objects as values
3. "reduced_chain_link_dict" for the condensed dict with step names as keys and Link ojects as values

We can decide on this! It won't be a big deal to change these names again :-)

in_toto/verifylib.py
RohanBhirangi: @lukpueh We decided not to do this!

in_toto/verifylib.py
RohanBhirangi: The algorithm is a bit simplified now. Less nesting.
We are skipping over the non-threshold steps.
The function also checks if the link file was indeed signed by different authorized functionaries (or rather was not signed by the same functionary to just satisfy the threshold).

in_toto/verifylib.py
RohanBhirangi: @lukpueh I haven't implemented this yet!
Do we need this? I don't think an extra function would be much of a problem.

lukpueh: Yes please!

RohanBhirangi:
@lukpueh @SantiagoTorres You may review the PR again! :-)

lukpueh:
@RohanBhirangi, please resolve any existing conflicts before we review.

demo/README.md
```diff
@@ -119,7 +119,7 @@ our software package `foo.tar.gz` and the related metadata files `root.layout`,
 `write-code.link` and `package.link`:
```
lukpueh: Please update link naming scheme.

demo/run_demo.py
```diff
@@ -118,17 +122,19 @@ def main():
...
       "final_product/alice.pub",
       "final_product/foo.py",
       "final_product/foo.tar.gz",
-      "final_product/package.link",
-      "final_product/write-code.link",
-      "final_product/tar.link",
+      "final_product/package.c1ae1e51.link",
+      "final_product/write-code.0c6c50a1.link",
+      "final_product/package.0c6c50a1.link",
+      "final_product/untar.link",
```
lukpueh: Thanks for adding this line. :)

demo/run_demo.py
```diff
@@ -118,17 +122,19 @@ def main():
...
       "owner_alice/root.layout",
-      "functionary_bob/write-code.link",
+      "functionary_bob/write-code.0c6c50a1.link",
       "functionary_bob/foo.py",
-      "functionary_carl/package.link",
+      "functionary_bob/package.0c6c50a1.link",
+      "functionary_bob/foo.tar.gz",
+      "functionary_carl/package.c1ae1e51.link",
       "functionary_carl/foo.tar.gz",
-      "functionary_carl/foo.py",
```
lukpueh: Please re-add this line.

test/test_runlib.py
```diff
@@ -375,18 +374,18 @@ def setUpClass(self):
...
     generate_and_write_rsa_keypair(self.key_path)
     generate_and_write_rsa_keypair(self.key_path2)
     self.key = prompt_import_rsa_key_from_file(self.key_path)
     self.key2 = prompt_import_rsa_key_from_file(self.key_path2)
 
+    self.step_name = "test-step"
+    self.link_name = "{}.{:.8}.link".format(self.step_name, self.key["keyid"])
+    self.link_name_unfinished = UNFINISHED_FILENAME_FORMAT.format(
+        step_name=self.step_name, short_keyid="{:.8}".format(self.key["keyid"]))
+
```
lukpueh: Please remove trailing whitespace.

in_toto/models/layout.py
```diff
@@ -135,13 +136,20 @@ def import_step_metadata_from_files_as_dict(self):
...
 
     """
     step_link_dict = {}
     for step in self.steps:
-      link = models__link.Link.read_from_file(step.name + '.link')
-      step_link_dict[step.name] = link
+      key_link_dict = {}
+      for keyid in step.pubkeys:
+        try:
+          link = models__link.Link.read_from_file(step.name + '.' + '{:.8}'.format(keyid) + '.link')
```
lukpueh: Use `FILENAME_FORMAT` here instead.

in_toto/models/layout.py
```diff
@@ -135,13 +136,20 @@ def import_step_metadata_from_files_as_dict(self):
...
+        except IOError as e:
+          raise IOError("Link file not found. Exception: {}".format(e))
```
lukpueh: This only runs through if every functionary authorized to carry out a step (`step.pubkeys`) actually did so. But note, authorized does not mean required. The `threshold` attribute should determine how many different functionaries are required to carry out the step and provide a link file.

in_toto/models/layout.py
```diff
@@ -237,6 +245,7 @@ class Step(models__common.Metablock):
   product_matchrules = attr.ib(default=attr.Factory(list))
   pubkeys = attr.ib(default=attr.Factory(list))
   expected_command = attr.ib("")
+  threshold = attr.ib(default=1)
```
lukpueh, line 258: Please add the attribute to the `Attributes` section in the class' docstring.

in_toto/models/layout.py
```diff
@@ -256,13 +266,10 @@ def _validate_type(self):
...
-    try:
-      # int(self.threshold)
-      # FIXME: we don't have support for threshold yet
-      pass
-    except Exception as e:
+    if type(self.threshold) != int:
       raise securesystemslib.exceptions.FormatError(
-          "Invalid threshold value for this step. Exception: {}".format(e))
+          "Invalid threshold '{}' value, must be an int."
+          .format(self.threshold))
```
lukpueh: https://github.com/in-toto/in-toto/pull/69#discussion_r100336540

in_toto/models/link.py
```diff
@@ -65,12 +69,25 @@ class Link(models__common.Signable):
...
+    """
+    Method to dump the link metdata files
+    If filename is provided, link is stored with that name
+    If key is provided, then the usual [step-name].[keyid].link
+    format is used to store the link
+    If nothing is provided (during inspection), link is stored
+    as [step-name].link
+    If both filename and key are provided, the key gets ignored
+    as first preference is given to the filename.
+    """
```
lukpueh: Minor nit: Please take a look at our [code style guidelines](https://github.com/secure-systems-lab/code-style-guidelines#documentation-strings) for docstrings.

in_toto/models/link.py
```diff
@@ -21,6 +21,10 @@
 import attr
 import json
 from . import common as models__common
+import securesystemslib.formats
+
+FILENAME_FORMAT = "{step_name}.{short_keyid}.link"
+UNFINISHED_FILENAME_FORMAT = ".{step_name}.{short_keyid}.link-unfinished"
```
lukpueh: I know I suggested `short_keyid` but now that I look at it, I think it makes more sense to pass a `keyid` instead  and truncate in the format string, what do you think? This is not a blocker.

in_toto/models/link.py
```diff
@@ -65,12 +69,25 @@ class Link(models__common.Signable):
   command = attr.ib("")
   return_value = attr.ib(None)
 
-  def dump(self, filename=False):
-    """Write pretty printed JSON represented of self to a file with filename.
```
lukpueh: IMHO this phrase was more descriptive/correct than "Method to dump the link metadata files"

in_toto/verifylib.py
```diff
@@ -230,17 +231,17 @@ def verify_all_steps_signatures(layout, links_dict):
     Verifies cryptographic Link signatures related to Steps of a Layout.
 
   """
+  key_link_dict = {}
```
lukpueh: Why do you assign an empty dictionary to `key_link_dict` here and then a dictionary of keyids as keys and links as values below?

in_toto/verifylib.py
```diff
@@ -230,17 +231,17 @@ def verify_all_steps_signatures(layout, links_dict):
...
+  keys_dict = {}
```
lukpueh: As discussed the other day, this undermines the per-step-functionary authorization.
If you incrementally add the per-step-keys to a global dictionary and use it to verify each step, then you unwillingly authorize all functionaries authorized to carry out any prior step, to also carry out the current step.

in_toto/verifylib.py
```diff
@@ -230,17 +231,17 @@ def verify_all_steps_signatures(layout, links_dict):
...
-    for keyid in step.pubkeys:
-      keys_dict[keyid] = layout.keys[keyid]
-
-    # Verify link metadata file's signatures
-    verify_link_signatures(link, keys_dict)
+    key_link_dict = chain_link_dict[step.name]
+    for keyid, link in six.iteritems(key_link_dict):
+      # Create the dictionary of keys for this step
+      for pubkeyid in step.pubkeys:
+        keys_dict[pubkeyid] = layout.keys[pubkeyid]
```
lukpueh: So what I commented in https://github.com/in-toto/in-toto/pull/69#discussion_r100393391 is still valid: You don't have to create a new `key_dict` for every link of a step, because the per-step-authorized keys apply to all links of a step.
Howerver, you do have to create a new keys_dict for every step. That's why I said to move this loop (L241-242) out of the direct enclosing loop, i.e. above (L239) but below (L236).

in_toto/verifylib.py
```diff
@@ -218,7 +219,7 @@ def verify_all_steps_signatures(layout, links_dict):
   <Arguments>
```
lukpueh, line 221: "...verifies the signatures of the Link object(s) related to ..."

in_toto/verifylib.py
```diff
@@ -635,6 +637,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+    # Extract the key_link_dict for this step from the passed chain_link_dict
+    key_link_dict = chain_link_dict[step.name]
+
+    # take one exemplary link (e.g. the first in the step_link_dict)
+    compare_link = key_link_dict.values()[0]
+
+    # Check if we have at least <threshold> links for this step
+    if len(key_link_dict) < step.threshold:
+      raise ThresholdVerificationError("Step not performed by enough functionaries!".format())
+
```
lukpueh: Please wrap line.

in_toto/verifylib.py
```diff
@@ -635,6 +637,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+      for key, key_link in six.iteritems(layout.keys):
+        if key == keyid:
+          keys_dict[key] = key_link
+          verify_link_signatures(link, keys_dict)
+          break
+
+      # compare their properties
+      if compare_link.materials != link.materials or compare_link.products != link.products:
+        raise ThresholdVerificationError("Link artifacts do not match!".format())
+
```
lukpueh: Please wrap lines.

in_toto/verifylib.py
```diff
@@ -635,6 +637,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+    # iterate over the remaining links to compare their properties
+    for keyid, link in six.iteritems(key_link_dict):
+      keys_dict = {}
+
+      # check if the authorized functionary has signed the link
+      for key, key_link in six.iteritems(layout.keys):
+        if key == keyid:
+          keys_dict[key] = key_link
+          verify_link_signatures(link, keys_dict)
+          break
```
lukpueh: I am not sure what you are trying to do here (L690-L695) but this still does not fail if we have multiple links (with different keyid'ed names) but reapeating signatures.

Imagine a step "foo" with a threshold of "2" and two authorized functionaries "bob" and "carl".
Somone could duplicate Bob's legitimate `foo.[Bob's keyid].link`, rename the copy to `foo.[Carl's keyid].link` and we would think everything is fine.

I guess an easy way to prevent this is to check if the link (L687) has a signature with a keyid that is equal to the purported keyid (L687).

in_toto/verifylib.py
```diff
@@ -635,6 +637,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+
+      # compare their properties
+      if compare_link.materials != link.materials or compare_link.products != link.products:
+        raise ThresholdVerificationError("Link artifacts do not match!".format())
+
+
+def reduce_chain_links(layout, chain_link_dict):
+  """
+  <Purpose>
+    Iterates throught the passed chain_link_dict and builds a dict with
```
lukpueh: Fix spelling.

in_toto/verifylib.py
```diff
@@ -635,6 +637,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+
+  <Returns>
+    A dictionary containing one Link metadata object per step only if
+    the link artifacts of all link objects are identical for a step.
+
+  """
+
+  reduced_chain_link_dict = {}
+
+  for step in layout.steps:
```
lukpueh: This is not what the docstrings say, I think you can actually iterate over the passed chain_link_dict, then you don't need to pass the layout at all.

in_toto/models/layout.py
RohanBhirangi: I have already changed this before. My fork reflects this change too. It's strange that you encountered this.

in_toto/models/link.py
RohanBhirangi: Do you mean something like FILENAME_FORMAT = "{}.{:.8}.link".format(step_name, keyid) ?

in_toto/verifylib.py
RohanBhirangi: I didn't get this!

P.S. I haven't touched the docstring of this function.

in_toto/verifylib.py
RohanBhirangi: So what I do (or was trying to do) is, I extract the link object from the chain_link_dict.
Then, I extract the entire key object corresponding to the key in the filename.
Then I pass them to verify_link_signatures(link, keys_dict). This ensures that the link object has been indeed signed by the same key as the filename.

For eg. if we have foo.[Bob's keyid].link and foo.[Carl's keyid].link, and someone duplicates legitimate foo.[Bob's keyid].link by renaming it to foo.[Carl's keyid].link. We basically check if foo.[Carl's keyid].link was indeed signed by Carl's key.

Note that I am passing only a single key object as keys_dict in verify_link_signatures(link, keys_dict) instead of all the keys. This restricts the possibility of the link object being signed by any other key.

in_toto/models/link.py
lukpueh: Exactly. This eliminates the need to call `FILENAME_FORMAT` (or `UNFINISHED_FILENAME_FORMAT`) with another formatter as the `short_keyid` argument.

in_toto/verifylib.py
lukpueh: I know, but you changed the implementation of the function to verify signatures of potentially multiple Links for one step. That's why I suggest to also change the docstring.

in_toto/verifylib.py
lukpueh: You are right. I think I was confused by the variable names: `keyid`,`link`, `key`, `key_link`. :)
I think instead of iterating over `layout.keys`, you can just fetch the key with the `keyid` from the outer loop and verify the link's signature with that key, e.g.:
```python
key_dict = {
  keyid: layout.keys.get(keyid)
}
verify_link_signatures(link, key_dict)
```
And maybe you want to wrap this in a try/except block and raise a `ThresholdVerificationError` with a descriptive message if the signature verification fails.

in_toto/models/layout.py
```diff
@@ -135,13 +137,26 @@ def import_step_metadata_from_files_as_dict(self):
...
     for step in self.steps:
-      link = models__link.Link.read_from_file(step.name + '.link')
-      step_link_dict[step.name] = link
+      key_link_dict = {}
+      link_count = 0
+      for keyid in step.pubkeys:
+        link_count += 1
+        try:
+          link = models__link.Link.read_from_file(FILENAME_FORMAT.format(
+            step_name=step.name, keyid=keyid))
```
lukpueh: We use [double-indentation for line continuation](https://github.com/secure-systems-lab/code-style-guidelines#indentation).

in_toto/models/layout.py
```diff
@@ -135,13 +137,26 @@ def import_step_metadata_from_files_as_dict(self):
...
 
     """
     step_link_dict = {}
     for step in self.steps:
-      link = models__link.Link.read_from_file(step.name + '.link')
-      step_link_dict[step.name] = link
+      key_link_dict = {}
+      link_count = 0
+      for keyid in step.pubkeys:
+        link_count += 1
```
lukpueh: Hint: Use enumerate eg.: 
```python
for idx, value in enumerate(list):
   ...
````

File: in_toto/models/layout.py [hunk 54]
```diff
@@ -135,13 +137,26 @@ def import_step_metadata_from_files_as_dict(self):
...
+        try:
+          link = models__link.Link.read_from_file(FILENAME_FORMAT.format(
+            step_name=step.name, keyid=keyid))
+        except IOError as e:
+          raise IOError("Link file not found. Exception: {}".format(e))
+        key_link_dict[keyid] = link
```
lukpueh: So this still fails if one of the authorized functionaries did not actually perform the step. Which is not what we want (c.f. [last review](https://github.com/in-toto/in-toto/pull/69/files/1b489d2d1918f0fca8f4a58881954800d5ab0565#r102794049)).

in_toto/models/layout.py
```diff
@@ -135,13 +137,26 @@ def import_step_metadata_from_files_as_dict(self):
...
+      if link_count < step.threshold:
+        raise ThresholdVerificationError("Step not performed" +
+          " by enough functionaries!".format())
```
lukpueh: You don't need the plus (`+`) here. 
Also, add double-indentation.

in_toto/models/layout.py
lukpueh: Minor-issimo:
I'd put the comma (`,`) before `value`, or `value` right after `threshold`, e.g.: *"Invalid threshold '{}', value must be an int."* or *"Invalid threshold  value '{}', must be an int.* :)

in_toto/verifylib.py
lukpueh: bump

in_toto/verifylib.py
```diff
@@ -227,20 +228,23 @@ def verify_all_steps_signatures(layout, links_dict):
...
     # Find the according link for this step
-    link = links_dict[step.name]
+    key_link_dict = chain_link_dict[step.name]
 
     # Create the dictionary of keys for this step
-    keys_dict = {}
-    for keyid in step.pubkeys:
-      keys_dict[keyid] = layout.keys[keyid]
+    for pubkeyid in step.pubkeys:
+        keys_dict[pubkeyid] = layout.keys[pubkeyid]
```
lukpueh: Indentation consistency (2 spaces)!

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+    # Extract the key_link_dict for this step from the passed chain_link_dict
+    key_link_dict = chain_link_dict[step.name]
+
+    # take one exemplary link (e.g. the first in the step_link_dict)
+    compare_link = key_link_dict.values()[0]
+
+    # Check if we have at least <threshold> links for this step
+    if len(key_link_dict) < step.threshold:
+      raise ThresholdVerificationError("Step not performed" +
+        " by enough functionaries!")
```
lukpueh: https://github.com/in-toto/in-toto/pull/69/files#r103533990

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+    # iterate over the remaining links to compare their properties
+    for keyid, link in six.iteritems(key_link_dict):
+      keys_dict = {keyid: layout.keys.get(keyid)}
+
+      # check if the authorized functionary has signed the link
+      try:
+        verify_link_signatures(link, keys_dict)
+      except KeyError as e:
+        raise ThresholdVerificationError("Link not signed by"+
+          " authoried functionary!")
```
lukpueh: https://github.com/in-toto/in-toto/pull/69/files#r103533990

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+
+
+      # compare their properties
+      if (compare_link.materials != link.materials or
+        compare_link.products != link.products):
+        raise ThresholdVerificationError("Link artifacts do" +
+          " not match!")
```
lukpueh: https://github.com/in-toto/in-toto/pull/69/files#r103533990

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+      try:
+        verify_link_signatures(link, keys_dict)
+      except KeyError as e:
+        raise ThresholdVerificationError("Link not signed by"+
+          " authoried functionary!")
+
+
+      # compare their properties
+      if (compare_link.materials != link.materials or
+        compare_link.products != link.products):
```
lukpueh: Double-indentation

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+        raise ThresholdVerificationError("Link artifacts do" +
+          " not match!")
+
+
+def reduce_chain_links(layout, chain_link_dict):
```
lukpueh: I think we don't need the layout argument anymore, right?

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
         link.products, target_links)
 
 
+def verify_threshold_equality(layout, chain_link_dict):
+  """
+  <Purpose>
+    Iteratively verifies threshold conditions (if a step has been
+    performed and signed by enough number of functionaries) and link
+    artifacts (materials and products of a step is same for different
+    fucntionaries) of passed dictionary (chain_link_dict).
```
lukpueh: Spelling (Hint: You can enable spell checking in SublimeText)

in_toto/verifylib.py
lukpueh: Spelling

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+  for step in layout.steps:
+    # Skip steps that don't require multiple functionaries
+    if step.threshold <= 1:
+      continue
+
+    # Extract the key_link_dict for this step from the passed chain_link_dict
+    key_link_dict = chain_link_dict[step.name]
+
+    # take one exemplary link (e.g. the first in the step_link_dict)
+    compare_link = key_link_dict.values()[0]
```
lukpueh: If you extract the compare_link after checking threshold below you won't get an index out of range in case of an empty key_link_dict.

in_toto/verifylib.py
```diff
@@ -635,6 +640,113 @@ def verify_all_item_rules(items, links, target_links=None):
...
+        " by enough functionaries!")
+
+    # iterate over the remaining links to compare their properties
+    for keyid, link in six.iteritems(key_link_dict):
+      keys_dict = {keyid: layout.keys.get(keyid)}
+
+      # check if the authorized functionary has signed the link
+      try:
+        verify_link_signatures(link, keys_dict)
+      except KeyError as e:
```
lukpueh: I think you don't want to catch a `KeyError` here but rather a `SignatureVerificationError`, right?

lukpueh (changes requested):
Almost there. One more round and we should be good to merge.

in_toto/verifylib.py
RohanBhirangi: Do we need to get an index out of range in case of an empty key_link_dict? So like does the extraction of compare_link have to be before the threshold check?

If we do not have a link object in the key_link_dict, it will fail the previous verification checks (verify_all_steps_signatures).

lukpueh: I don't fully understand your question.

What I was saying...
In your implementation you would get an index out of range error in case the `chain_link_dict` is empty. 

If you just move what you do in L683 a couple of lines below, you'd get a little bit of defensive programming for free, i.e. you'd probably raise your threshold verification before getting an index out of range error (independently of what we did earlier, e.g. in `verify_all_steps_signatures`).

in_toto/models/layout.py
```diff
@@ -135,13 +138,25 @@ def import_step_metadata_from_files_as_dict(self):
...
+      A dictionary with Link names as keys and a dict (key_id Link objects)
+      as values.
 
     """
     step_link_dict = {}
     for step in self.steps:
-      link = models__link.Link.read_from_file(step.name + '.link')
-      step_link_dict[step.name] = link
+      key_link_dict = {}
+      for idx, keyid in enumerate(step.pubkeys):
```
lukpueh: If you don't need an index, you don't need to use `enumerate`. :)

lukpueh (changes requested):
One last thing and we are good to go. :)

lukpueh (approved):

PR #70, opened 2017-02-08
Major artifact rule update to align with specs

lukpueh:
Aligns with latest specs and discussion in https://github.com/in-toto/in-toto/issues/43

**Changes in verifylib**
- Artifact rules updates
   - Changes MATCH rule format and behavior 
   ```
   MATCH <pattern> [IN <source-path-prefix>] WITH (MATERIALS|PRODUCTS) 
    [IN <destination-path-prefix>] FROM <step>
    ```
  - Changes CREATE/DELETE rule behavior
  - Re-adds MODIFY rule
  - Adds ALLOW/DISALLOW rules
- Adds hashdict equality check function
- Changes verify_item_rules arguments
  Receives additional source_type argument to know if the rules were in material_matchrules or 
  product_matchrules of a step or inspection.
  Removes artifacts arguments. Source artifacts are now taken from the passed links because they 
  contains all links now (steps and inspections).
- Changes verify_item_rules algorithm
  Does not return when artifact queue is empty. This prevents rules from passing just because they
  were skipped.
- Changes verify_all_item_rules arguments
  Receives only one link dictionary for step links and inspection links.

**Various related changes**
- Replaces matchrule_validators.py with artifact_rules.py
   Its function unpack_rule is now used instead of check_matchrule_syntax
  to parse all rules for rule data and check syntax.
 - Removes flatten and invert utility function. 
    Matching solely by hash leads to collisions.
 - Updates demo to use new rules
 - Updates unit tests


demo/owner_alice/create_layout.py
```diff
@@ -36,21 +36,20 @@ def main():
...
+            ["MATCH", "foo.tar.gz", "WITH", "PRODUCTS", "FROM", "package"],
             # FIXME: Without the "allow everything else" rule here
             # inspection would fail because of the metadata and other files
             # (.DS_STORE, layout key, ...) in the directory where the Inspection
             # is executed, which get recorded as materials.
             # The behavior is actually wanted in order to prevent sneaking
             # files. But we do have to think of a way to ignore
             # irrelevant files.
-            ["CREATE", "*"],
+            ["ALLOW", "*"],
```
SantiagoTorres: do we still need this?

demo/owner_alice/create_layout.py
```diff
@@ -36,21 +36,20 @@ def main():
...
         ],
         "product_matchrules": [
-            ["MATCH", "PRODUCT", "foo.py", "AS", "foo.py",
-                "FROM", "write-code"],
+            ["MATCH", "foo.py", "WITH", "PRODUCTS", "FROM", "write-code"],
             # FIXME: See material_matchrules above
-            ["CREATE", "*"],
+            ["ALLOW", "*"],
```
SantiagoTorres: also here

in_toto/artifact_rules.py
```diff
@@ -0,0 +1,186 @@
+"""
+<Module Name>
+  in_toto/artifact_rules.py
+
+<Author>
+  Santiago Torres-Arias <santiago@nyu.edu>
```
SantiagoTorres, line 6: considering all the refactoring done, maybe it would be good to add your name here

in_toto/verifylib.py
```diff
@@ -306,294 +308,582 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
 
-def verify_match_rule(rule, artifact_queue, artifacts, links):
-  """
-  <Purpose>
-    Verifies that target path pattern - 3rd or 5th element of rule - matches
-    at least one file in target artifacts. This might conflict with
-    general understanding of glob patterns (especially "*").
+def _equal_hash_dicts(hash_dict_a, hash_dict_b):
+  """Compare to hash_dict for equality, formats are
+  securesystemslib.formats.HASHDICT_SCHEMA. Return True or False. """
```
SantiagoTorres: Is this check being made before? or should we run this check before doing anything else.

in_toto/verifylib.py
```diff
@@ -306,294 +308,582 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+
+  if set(hash_dict_a.keys()) != set(hash_dict_b.keys()):
+    return False
 
-    Further verifies that each matched target artifact has a corresponding
-    source artifact with a matching hash in the passed artifact dictionary and
-    that this artifact in also the artifact queue.
+  for key in hash_dict_a.keys():
+    if hash_dict_a[key] != hash_dict_b[key]:
```
SantiagoTorres: I assume there are no recursive properties, but could we short_circuit to check for equality in both the value and the reference? 

I'm also not sure if this method is [strictly necessary](https://github.com/python/cpython/blob/master/Objects/dictobject.c#L2697-L2737)

in_toto/verifylib.py
```diff
@@ -306,294 +308,582 @@ def verify_all_steps_command_alignment(layout, links_dict):
...
+      source_products_queue, rule_data["pattern"])
 
-def verify_delete_rule(rule, artifact_queue):
+  for matched_product in matched_products:
+    if matched_product in source_materials_queue:
+      raise RuleVerficationError("Rule '{0}' failed, product '{1}' was found"
+          " in materials but should have been newly created."
+          .format(rule, matched_product))
+
+  return list(set(source_products_queue) - set(matched_products))
```
SantiagoTorres: I wonder if this would be easier to do using pop() (maybe deepcopying the list). This may just be a matter of style though.

SantiagoTorres (changes requested):
I have minor comments. Given that this PR is so big, I want to make another pass once these are addressed....

in_toto/artifact_rules.py
```diff
@@ -0,0 +1,186 @@
...
+        " rule must start with one of '{0}' and specify a 'pattern' as"
+        " second element.\n"
+        "Got: \n\t'{1}'".format(", ".join(ALL_RULES), rule))
+
+  rule_type = rule[0].lower()
+  pattern = rule[1]
+
+  if not isinstance(pattern, six.string_types):
+    raise securesystemslib.exceptions.FormatError("Wrong rule format,"
+        " 'pattern' (second element) must be a string.\n"
```
SantiagoTorres: I wonder if here we want to say "the pattern". It feels oddly worded here.

in_toto/artifact_rules.py
```diff
@@ -0,0 +1,186 @@
...
+        "type": rule_type,
+        "pattern": pattern,
+      }
+
+  # Type is "MATCH"
+  elif rule_type in COMPLEX_RULES:
+
+    # ... IN <source-path-prefix> WITH (MATERIALS|PRODUCTS)
+    # IN <destination-path-prefix> FROM <step>
+    if (rule_len == 10 and rule[2].lower() == "in" and
```
SantiagoTorres: I wonder if it would improve readabilty to lower() all the strings before starting comparisons. e.g.,
```
rule = [x.lower for x in rule]
```

in_toto/artifact_rules.py
```diff
@@ -0,0 +1,186 @@
...
+      dest_name = rule[5]
+
+    else:
+      raise securesystemslib.exceptions.FormatError("Wrong rule format,"
+          " match rules must have the format:\n\t"
+          " MATCH <pattern> [IN <source-path-prefix>] WITH"
+          " (MATERIALS|PRODUCTS) [IN <destination-path-prefix>] FROM <step>.\n"
+          "Got: \n\t{}".format(rule))
+
+    if not isinstance(dest_type, six.string_types) or (
```
SantiagoTorres: wouldn't it be nicer to iterate over all the rules to check they are strings first?

in_toto/models/layout.py
```diff
@@ -93,6 +93,7 @@ def dump(self, filename='root.layout'):
   def read_from_file(filename='root.layout'):
     """Static method to instantiate a new Layout object from a
     canonical JSON serialized file """
+    print filename
```
SantiagoTorres: I don't believe we want this.

test/models/test_artifact_rules.py
```diff
@@ -0,0 +1,191 @@
...
+  Nov 19, 2016
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Test artifact rule unpacking.
+
+"""
+
```
SantiagoTorres: I see this blank space in some files but not in others. Does it matter?

test/models/test_artifact_rules.py
```diff
@@ -0,0 +1,191 @@
...
+  def test_unpack_rule_not_enough_keywords(self):
+    """Test rule syntax error, too little arguments. """
+    rule = ["DELETE"]
+    with self.assertRaises(securesystemslib.exceptions.FormatError):
+      unpack_rule(rule)
+
+  def test_unpack_rule_unknown_rule_type(self):
+    """Test generic rule syntax error, too many arguments. """
+    rule = ["SUBVERT", "foo"]
```
SantiagoTorres, line 43: This should be a valid rule tbh

test/test_verifylib.py
```diff
@@ -156,121 +159,275 @@ def test_commands_do_not_align_at_all_log_warning(self):
...
 
-    rule = ["CREATE", "bar"]
-    with self.assertRaises(RuleVerficationError):
-      verify_create_rule(rule, self.artifact_queue)
+  @classmethod
+  def setUpClass(self):
+    sha_1 = ("d65165279105ca6773180500688df4bd"
+                  "c69a2c7b771752f0a46ef120b7fd8ec3")
 
+    sha_2 = ("155c693a6b7481f48626ebfc545f0523"
```
SantiagoTorres: do we want to change these names to something that doesn't hint of either algorithm?

test/test_verifylib.py
```diff
@@ -307,231 +469,280 @@ def setUp(self):
...
+  """Test verifylib.verify_item_rules(source_name, source_type, rules, links)"""
 
   def setUp(self):
-    self.item_name = "test-item"
-    self.sha256_foo = \
+    self.item_name = "item"
+    self.sha256_1 = \
         "d65165279105ca6773180500688df4bdc69a2c7b771752f0a46ef120b7fd8ec3"
-    self.sha256_bar = \
+    self.sha256_2 = \
```
SantiagoTorres, line 689: Like this, I think this name is better.

SantiagoTorres (changes requested):

in_toto/verifylib.py
lukpueh: What check are you referring to? `securesystemslib.formats. HASHDICT_SCHEMA.check_match`?

in_toto/verifylib.py
lukpueh: I had this discussion with @awwad the other day. For some reason I was convinced that dictionaries had to be compared manually, apparently I was wrong.
I'll remove the function altogether.

in_toto/verifylib.py
lukpueh: I'm not married to having the set difference operation, especially as we need to temporarily cast. Using pip would require an additional loop over the list, though. I think we can leave it for now.

in_toto/artifact_rules.py
lukpueh: List comprehension!!

lukpueh: I thought about this too, but I think we shouldn't convert the user defined parts (pattern, prefix, step).

test/models/test_artifact_rules.py
lukpueh: Yep, we [want this](https://github.com/secure-systems-lab/code-style-guidelines#documentation-strings). But I won't change this everywhere in this PR. This could be candidate for a general code cleanup PR.
> Note that most importantly, the """ that ends a multiline docstring should be on a line by itself, and preferably preceded by a blank line, e.g.:

test/models/test_artifact_rules.py
lukpueh: ?

lukpueh:
@SantiagoTorres thanks for your comments. Ready for another round?

in_toto/artifact_rules.py
SantiagoTorres: we are making a copy of it here. Or what do you mean?

in_toto/verifylib.py
SantiagoTorres: The issue is that this is not a cast, you are allocating a new set object for both lists, and then turning the resulting set object into another list. Of course this is a minor detail, but it sounds expensive/heavy.

test/models/test_artifact_rules.py
SantiagoTorres: It says preceded, I'm talking about the succedding line before the imports.

test/models/test_artifact_rules.py
SantiagoTorres: ignore this, it's a silly comment 🚀

in_toto/artifact_rules.py
lukpueh: Making a copy of what where?

I meant two things:
1. List comprehension is discouraged by our code  style guidelines
2. I think the user defined parts of the rule "path pattern", "path prefix" and "step name" should not be converted to lower, because the case might matter for those.
Hence we'd have to explicitly convert certain elements that are at varying indices depending on the rule type or whether it has optional parts in it or not.

test/models/test_artifact_rules.py
lukpueh: Ah ok. I guess I can remove that. :)

in_toto/artifact_rules.py
```diff
@@ -0,0 +1,171 @@
...
+      "source_prefix": source_prefix,
+      "dest_prefix": dest_prefix,
+      "dest_type" : dest_type.lower(),
+      "dest_name": dest_name
+    }
+
+  #else:
+  # can't reach else
+  # if it is neither in GENERIC_RULES nor in MATCH_RULES it would
+  # have already raised an exception above when checking if it is in ALL_RULES
```
SantiagoTorres, line 178: I guess this else is left for the sake of documentation, but do we really want it?

SantiagoTorres (approved):

PR #72, opened 2017-02-24
Update README.md

JustinCappos:
I think this is now obsolete, right?


lukpueh:
Forgot to take this out. Thanks, @JustinCappos !

Issue #25, opened 2016-10-27
Use "{}".format instead of "%s" % ()


lukpueh:
@SantiagoTorres made a [valid point](https://github.com/in-toto/in-toto/pull/3#discussion_r84121985) in preferring `"{}".format` over  `"%s" % ()` for string formatting.

See more info in the [Python docs](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting):

> Note: The formatting operations described here exhibit a variety of quirks that lead to a number of common errors (such as failing to display tuples and dictionaries correctly). Using the newer str.format() interface helps avoid these errors, and also provides a generally more powerful, flexible and extensible approach to formatting text.

vladimir-v-diaz:
I think the code style guidelines is sprinkled with examples like so:

```Python
print(str(foo) + " wants to get drunk at " + str(bar))
```

SantiagoTorres:
Hmmm, I don't know if `str(foo) + ...` is better memory-wise. I guess for the simple case this wouldn't be a problem.

 Also, it doesn't allow for format specifications such as:
```
 "{:^10}".format(number) # fancily center-align something.
```

vladimir-v-diaz:
I find the examples listed in the guidelines easier to read, however, format() might be more convenient to use.

One could make the case that format() requires more effort to parse since you have to scan the list of variables at the end to make sense of the string, and newcomers would have to look up documentation for special uses of format()

JustinCappos:
What is wrong with using "text "+str(foo)+" more text "+str(bar)  ?

Thanks,
Justin

SantiagoTorres:
> What is wrong with using "text "+str(foo)+" more text "+str(bar) ? 

I think the biggest issue is that + is a binary operator, and thus it can't be smart in the way you concatenate strings. From [here](https://softwareengineering.stackexchange.com/questions/304445/why-is-s-better-than-for-concatenation):

> In Python str is immutable, so the left and right string have to be copied into the new string for every pair of concatenation. If you concatenate four strings of length 10, you will be copying (10+10) + ((10+10)+10) + (((10+10)+10)+10) = 90 characters, instead of just 40 characters. And things gets quadratically worse as the number and size of the string increases. Java optimizes this case some of the times by transforming the series of concatenation to use StringBuilder, but CPython doesn't.

There's also the issue with the lack of format specifiers that I mentioned above.

[...]

PR #77, opened 2017-03-22
Enhances user feedback (#6, #25)

lukpueh:
This PR addresses various https://github.com/in-toto/in-toto/issues/6

- Adds more user feedback
- Refactors logging wrapper 
- Replaces '%' formatter with `format` (https://github.com/in-toto/in-toto/issues/25)
- Removes/replaces `Exception, e:` with `Exception as e` syntax (https://github.com/in-toto/in-toto/issues/6#issuecomment-256982234)
- Changes wording in various info and exception messages

### DIFF
--- a/in_toto/in_toto_verify.py
+++ b/in_toto/in_toto_verify.py
@@ -29,9 +29,6 @@ import argparse
 
 from in_toto import log
 from in_toto import verifylib
-from in_toto.exceptions import (SignatureVerificationError, LayoutExpiredError,
-    RuleVerficationError, ThresholdVerificationError, BadReturnValueError,
-    LinkNotFoundError)
 
 def in_toto_verify(layout_path, layout_key_paths):
   """
@@ -59,16 +56,8 @@ def in_toto_verify(layout_path, layout_key_paths):
   try:
     verifylib.in_toto_verify(layout_path, layout_key_paths)
 
-  except (SignatureVerificationError, LayoutExpiredError,
-      RuleVerficationError, ThresholdVerificationError,
-      BadReturnValueError, LinkNotFoundError) as e:
-
-    # Log exception name and message
-    log.fail_verification("{0} - {1}".format(type(e).__name__, e))
-    sys.exit(1)
-
   except Exception as e:
-    log.error("in in-toto verify - {}".format(e))
+    log.fail_verification("{0} - {1}".format(type(e).__name__, e))
     sys.exit(1)
 
 def main():
--- a/in_toto/in_toto_verify.py
+++ b/in_toto/in_toto_verify.py
@@ -27,8 +27,11 @@
 import sys
 import argparse
 
-from in_toto import verifylib
 from in_toto import log
+from in_toto import verifylib
+from in_toto.exceptions import (SignatureVerificationError, LayoutExpiredError,
+    RuleVerficationError, ThresholdVerificationError, BadReturnValueError,
+    LinkNotFoundError)
 
 def in_toto_verify(layout_path, layout_key_paths):
   """
@@ -55,6 +58,15 @@ def in_toto_verify(layout_path, layout_key_paths):
   """
   try:
     verifylib.in_toto_verify(layout_path, layout_key_paths)
+
+  except (SignatureVerificationError, LayoutExpiredError,
+      RuleVerficationError, ThresholdVerificationError,
+      BadReturnValueError, LinkNotFoundError) as e:
+
+    # Log exception name and message
+    log.fail_verification("{0} - {1}".format(type(e).__name__, e))
+    sys.exit(1)
+
   except Exception as e:
     log.error("in in-toto verify - {}".format(e))
     sys.exit(1)
@@ -86,7 +98,7 @@ def main():
 
   args = parser.parse_args()
 
-  # Turn on all the `log.doing()` in the library
+  # Turn on all the `log.info()` in the library
   if args.verbose:
     log.logging.getLogger().setLevel(log.logging.INFO)
 
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -42,7 +42,7 @@ import securesystemslib.exceptions
 if os.name == 'posix' and sys.version_info[0] < 3:
   try:
     import subprocess32 as subprocess
-  except Exception:
+  except ImportError:
     log.warn("POSIX users (Linux, BSD, etc.) are strongly encouraged to"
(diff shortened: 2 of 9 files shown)


in_toto/in_toto_verify.py
```diff
@@ -27,8 +27,8 @@
 import sys
 import argparse
 
-from in_toto import verifylib
 from in_toto import log
+from in_toto import verifylib
```
SantiagoTorres, line 31: any reason as to why this order changed? Just curious.

in_toto/in_toto_verify.py
```diff
@@ -55,8 +55,9 @@ def in_toto_verify(layout_path, layout_key_paths):
   """
   try:
     verifylib.in_toto_verify(layout_path, layout_key_paths)
+
```
SantiagoTorres, line 58: is this space needed?

in_toto/runlib.py
```diff
@@ -27,28 +27,28 @@
...
+import securesystemslib.exceptions
+
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
 # install and use the much more recent subprocess32
 if os.name == 'posix' and sys.version_info[0] < 3:
   try:
     import subprocess32 as subprocess
-  except Exception, e:
-    logging.warning("POSIX users (Linux, BSD, etc.) are strongly encouraged to"
+  except Exception:
```
SantiagoTorres: Can we use a more specific exception? (e.g., ImportError)

in_toto/util.py
```diff
@@ -124,7 +124,7 @@ def prompt_import_rsa_key_from_file(filepath):
   password = None
   try:
     import_rsa_key_from_file(filepath)
-  except securesystemslib.exceptions.CryptoError, e:
+  except securesystemslib.exceptions.CryptoError:
```
SantiagoTorres, line 127: I wonder, is the lack of password a `CryptoError` specifically. Isn't there another case that could cause this behavior and ask for a password to a key it didn't even load?

test/test_verifylib.py
```diff
@@ -936,7 +936,7 @@ def setUpClass(self):
   def tearDownClass(self):
     """Change back to initial working dir and remove temp dir. """
     os.chdir(self.working_dir)
-    log.doing(self.test_dir)
+    log.info(self.test_dir)
     #shutil.rmtree(self.test_dir)
```
SantiagoTorres: I don't know if we need this line

SantiagoTorres (changes requested):
I noticed some tangential little issues.

in_toto/in_toto_verify.py
lukpueh: Actually yes. I find ist nicer to read if imports are ordered by length. :)

in_toto/in_toto_verify.py
lukpueh: Just a personal convention I picked up recently. I add blank lines after control flow blocks even if it's compound statements.

Our [style guideline just suggests](https://github.com/secure-systems-lab/code-style-guidelines#blank-lines) to
> Use blank lines in functions to indicate logical sections and help to offset comments.

in_toto/runlib.py
lukpueh: Good idea. Will do.

in_toto/util.py
lukpueh: Correct. Loading an encrypted key without a password is just one scenario that triggers a `CryptoError`. When I wrote this I quickly wanted a user interface to load both encrypted and non-encrypted keys. 

We should probably re-factor the whole util module at some point and see if we can move functionality to  `securesystemslib.keys`. I suggest we do this in a separate PR though.

test/test_verifylib.py
lukpueh: Probably not. Will remove. :)

in_toto/util.py
SantiagoTorres: Ok sure, should we ticketize this issue?

in_toto/in_toto_verify.py
SantiagoTorres: Okidoke 👍

in_toto/in_toto_verify.py
SantiagoTorres: it's also alphabetical 💯

test/test_verifylib.py
SantiagoTorres: Wait, don't we need to uncomment instead?

lukpueh: You are right. I guess had commented the cleanup out for debugging purposes. Since it's temp directories, the os should remove them at some point anyway. Nevertheless, I commented the cleanup back in (85e607b).

in_toto/util.py
lukpueh: Done: https://github.com/in-toto/in-toto/issues/80

in_toto/verifylib.py
```diff
@@ -943,35 +960,47 @@ def verify_threshold_equality(layout, chain_link_dict):
...
     for keyid, link in six.iteritems(key_link_dict):
       keys_dict = {keyid: layout.keys.get(keyid)}
 
       # check if the authorized functionary has signed the link
       try:
         verify_link_signatures(link, keys_dict)
       except SignatureVerificationError as e:
-        raise ThresholdVerificationError("Link not signed by"
-            " authorized functionary!")
+        raise ThresholdVerificationError("Link '{0}' not signed by"
```
SantiagoTorres, line 990: I wonder if we want to say 
```
Link '{0}' *was* not signed by *an* authorized functionary!
```

Maybe we need to replace the *an* with *enough* or so.

lukpueh: Actually we already check in [verify_all_steps_signatures](https://github.com/in-toto/in-toto/blob/output-enh/in_toto/verifylib.py#L251-L254) if a signature is correct and the functionary key was authorized to sign.

Here we check again (by filename keyid) to prevent threshold circumvention by duplicating a validly signed link and changing one filename.

lukpueh: Now that I look at it I wonder why we are not doing this in `verify_all_steps_signatures`. My reasoning should be buried somewhere among my many [threshold review comments](https://github.com/in-toto/in-toto/pull/69).

SantiagoTorres: Ok, I'll opened ticket #82 to mention this. I don't think this is a blocker for this review.

PR #78, opened 2017-03-22
IDs not in link files...

JustinCappos:
I didn't find the IDs in the link files so had to change the directions to accommodate.  

Discard this PR if this is because my in-toto software is out of date.  (However, if this is so perhaps change the directions to use * in the cp commands in place of the keyIDs...

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -160,10 +160,10 @@ It's time to release our software now.
 ### Verify final product (client)
 Let's first copy all relevant files into the `final_product` that is
 our software package `demo-project.tar.gz` and the related metadata files `root.layout`,
-`clone.[Bob's keyid].link`, `update-version.[Bob's keyid].link` and `package.[Carl's keyid].link`:
+`clone.link`, `update-version.link` and `package.link`:
 ```shell
 cd ..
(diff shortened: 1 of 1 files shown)


lukpueh:
Thanks for the PR. As you assumed  there probably was a version mismatch between the demo instructions you had at hand and the code base you used for the demo.

We merged  @RohanBhirangi's key threshold PR (https://github.com/in-toto/in-toto/pull/69) yesterday. This introduced a new naming scheme for the link files.

JustinCappos:
Okay did you fix the demo to use bob.*.link, etc. in the instructions?

lukpueh:
Actually, I'm not sure why we should do this. IMHO, it is better if the copy-paste commands in the demo instructions are very verbatim (to show the concepts). Given that the demo always uses the same keys the IDs in the filenames do not change.
Also, I think in a later iteration, we will replace the `cp`s with something like `in-toto-bundle`, as you suggested in https://github.com/in-toto/in-toto/issues/79.

JustinCappos:
Okay, that makes sense.  I didn't realize the keys were always the same and
thought the directions wouldn't work when cut-pasted.  Thanks for
clarifying.

PR #81, opened 2017-03-29
Fixes dead symlink bug in `record_artifacts_as_dict`

lukpueh:
File existence is only checked on the first level of the file
system traversal. If we recurse in a directory using os.walk,
we might also record symlinks to non-existent files, which would
crash once we try reading the file.

This commit excludes non-existent files in our `os.walk` loop
and adapts the related unittest.

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -182,7 +182,11 @@ def record_artifacts_as_dict(artifacts):
         filepaths = []
         for filename in files:
           norm_filepath = os.path.normpath(os.path.join(root, filename))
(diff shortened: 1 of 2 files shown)


SantiagoTorres:
should we `log.warn` on the else maybe? 

```
else:
   log.warn("file {} appears to be a broken symlink. Skipping..." ...)
```

Or so

lukpueh:
Ha, I just had the same thought. 👍

SantiagoTorres:
LGTM 💯

Issue #36, opened 2016-11-07
Add intelligent constructors for metadata classes


lukpueh:
Currently we use default constructors for all in-toto classes. There is a "factory-like" `read` method  which has been used in the demo so far. 

It would be useful to have intelligent constructors or factories for objects that will be created manually by the project owner, i.e. `Layout`, `Step` and `Inspection`, with the following capabilities.

- ~~Take key dictionary or key list and create key dictionary from it (Layout)~~ (wontfix)
- Take expected command in list or string format (Layout)
- Add default expiration if not specified (Layout)
- ~~Take matchrules in list or string format (Step, Inspection)~~ (wontfix, would require escaping)
- Validate passed data

PR #83, opened 2017-04-03
Store all commands as list and use `shlex` to split command strings

lukpueh:
This commit modifies the metadata format so that commands are always stored as list (empty list, if no command provided). Which fixes a warnings like: ```WARNING: Run command '' differs from expected command '[]'```

Changed metadata fields were: `Step.expected_command`, `Link.command` and `Inspection.run`. In case a command is supplied.

The metadata creation "factory" (`read`, `create_link_metadata`) functions still accept non-lists as input, but turn them into lists using `shlex.split` for shell like syntax.

The logic in the metadata creation functions will be moved to the respective constructors at some point, c.f. https://github.com/in-toto/in-toto/issues/36.

### DIFF
--- a/demo/run_demo.py
+++ b/demo/run_demo.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import shlex
 import subprocess
 import argparse
 import time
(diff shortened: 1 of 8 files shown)


SantiagoTorres:
👍 thanks!

PR #84, opened 2017-04-21
Add constructors with default attribute assignments (fix #36)

### LINKED ISSUE #36: Add intelligent constructors for metadata classes

lukpueh:
This PR adds a constructor to `Signable`, `Layout`, `Step`,`Inspection` and `Link` and moves the default attribute logic from the `read` methods to the constructors.
Note: The constructors henceforth only accept keyword arguments.

Additionally, it adds model validation to the constructors of `Layout`, `Step` and `Inspection`, which guarantees that no objects of above classes are created with malformed attributes.

*Background:*
The in-toto metadata classes used to perform some default attribute assignments in their static `read` methods, i.e. methods that create the respective object from a passed Python
dictionary using the auto-constructors from the `attr` module.

The following default assignments are performed:
- Default expiration date (1 month)
- Default empty data structures, if a value was not passed
- `shlex`-splitting Step commands (`expected_command`) and Inspection commands (`run`) if passed as String

This lead to inconsistencies between objects created using `read` and objects created the default constructors.

This PR also fixes a bug in the default expiration date format (i.e. removes milliseconds from the date string).

Fixes https://github.com/in-toto/in-toto/issues/36

### DIFF
--- a/in_toto/models/common.py
+++ b/in_toto/models/common.py
@@ -68,12 +68,16 @@ class Metablock(object):
         if method[0].startswith("_validate_"):
           method[1]()
 
-@attr.s(repr=False)
+@attr.s(repr=False, init=False)
 class Signable(Metablock):
   """Objects with base class Signable can sign their payload (a canonical
   pretty printed JSON string not containing the signatures attribute) and store
   the signature (signature format: securesystemslib.formats.SIGNATURE_SCHEMA) """
-  signatures = attr.ib(default=attr.Factory(list))
+  signatures = attr.ib()
+
+  def __init__(self, **kwargs):
+    super(Signable, self).__init__()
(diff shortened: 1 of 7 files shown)


SantiagoTorres:
LGTM

Issue #82, opened 2017-03-29
Coalesce signature verification in-toto verify


SantiagoTorres:
As it is now, there are two places that perform signature verification. Following review #77, we decided that both routines could be placed in the same place, and have the second place just verify that the signature matches the keyid prefix in the filename.

lukpueh:
This actually happened in https://github.com/in-toto/in-toto/pull/85. However, the way we verify signatures is somewhat confusing. I'll close here and will open a new "revisit signature verification" ticket.

PR #85, opened 2017-04-25
fixes signature verification in verifylib

RohanBhirangi:
verify all steps signatures strictly checks if the link is signed
by the keyid mentioned in the filename

### DIFF
--- a/in_toto/exceptions.py
+++ b/in_toto/exceptions.py
@@ -25,3 +25,8 @@ class LinkNotFoundError(Error):
   """Indicates that a ran command exited with non-int or non-zero return
   value. """
   pass
+
+class AuthorizationError(Error):
+  """Indicates that a ran command exited with non-int or non-zero return
+  value. """
+  pass
--- a/in_toto/exceptions.py
+++ b/in_toto/exceptions.py
@@ -22,11 +22,9 @@ class BadReturnValueError(Error):
   pass
 
 class LinkNotFoundError(Error):
-  """Indicates that a ran command exited with non-int or non-zero return
-  value. """
+  """Indicates that a link file was not found. """
   pass
 
 class AuthorizationError(Error):
-  """Indicates that a ran command exited with non-int or non-zero return
-  value. """
+  """Indicates that the link was signed by a non-authorized functionary. """
   pass
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -250,7 +250,7 @@ def verify_all_steps_signatures(layout, chain_link_dict):
       # filename is added to the dictionary which ensures
       # that the link has been signed by that key
       if keyid in step.pubkeys:
-            keys_dict[keyid] = layout.keys[keyid]
+        keys_dict[keyid] = layout.keys[keyid]
       else:
         raise AuthorizationError("Unauthorized Key! '{0}'".format(keyid))
 
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -246,9 +246,13 @@ def verify_all_steps_signatures(layout, chain_link_dict):
       keys_dict = {}
 
       # Create the dictionary of keys for this step
-      for pubkeyid in step.pubkeys:
-          if pubkeyid == keyid:
-            keys_dict[pubkeyid] = layout.keys[pubkeyid]
+      # Only one key with the matching keyid in the
+      # filename is added to the dictionary which ensures
+      # that the link has been signed by that key
+      if keyid in step.pubkeys:
+            keys_dict[keyid] = layout.keys[keyid]
+      else:
+        raise AuthorizationError("Unauthorized Key! '{0}'".format(keyid))
 
       log.info("Verifying signature(s) for '{0}'...".format(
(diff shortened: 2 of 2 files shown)


SantiagoTorres:
Hi @RohanBhirangi , Maybe I"m missing something but I don't see this is using the filename anywhere?

RohanBhirangi:
Hey, we load the link files (step.keyid.link) into the chain_link_dict ({step:{keyid: link}}) and check whether the link has been signed by the key corresponding to the keyid.
I assume this is what we wanted.

in_toto/verifylib.py
```diff
@@ -239,16 +239,17 @@ def verify_all_steps_signatures(layout, chain_link_dict):
...
-    for pubkeyid in step.pubkeys:
-      keys_dict[pubkeyid] = layout.keys[pubkeyid]
-
     for keyid, link in six.iteritems(key_link_dict):
+      keys_dict = {}
+
+      # Create the dictionary of keys for this step
+      for pubkeyid in step.pubkeys:
+          if pubkeyid == keyid:
+            keys_dict[pubkeyid] = layout.keys[pubkeyid]
```
lukpueh: Here are some notes,
- make sure to always have consistent indentation (2 spaces per level, 4 for line continuation),
- use `if pubkeyid in step.pubkeys` instead of the loop,
- I think it makes sense to fail early (plus give user feedback) , e.g. something like:
```python
if pubkeyid in step.pubkeys:
  keys_dict[pubkeyid] = layout.keys[pubkeyid]
else:
  raise SignatureVerificationError("<some descriptive message here>")
```
- also remove the signature verification part from the [threshold verification function]( https://github.com/in-toto/in-toto/blob/develop/in_toto/verifylib.py#L986-L993) as mentioned in the related [issue](https://github.com/in-toto/in-toto/issues/82),
- try to add some comments that describe what your changes are good for,
- if you fix an issue please mention the issue in the PR.

lukpueh (commented):
Thanks for the PR, please address comments.

in_toto/verifylib.py
SantiagoTorres: I'm adding to this, where does the keyid prefix of the link filename check actually happen?  I'm finding it hard to parse.

lukpueh: The variable value `keyid` holds the (whole) keyid, determined by the filename infix. So instead of taking all keys authorized for a certain step (`step.pubkeys`) and verifying that the link file is signed by one ore more of those keys, we take the key, identified by `keyid`, check if it is authorized, and then verify if the link was signed with that key.

RohanBhirangi: > if pubkeyid in step.pubkeys:
  keys_dict[pubkeyid] = layout.keys[pubkeyid]
else:
  raise SignatureVerificationError("<some descriptive message here>")


@lukpueh  Do we need this? This will raise an error if the keyid (as in the filename) is not present in the layout. But if it isn't present in the layout, our initial load_metadata will fail.
Or maybe I'm missing something.

lukpueh: Isn't this check something you are [proposing in your PR](https://github.com/RohanBhirangi/in-toto/blob/225c0169f672fa16972ba893c67a39b149313343/in_toto/verifylib.py#L249-L250)? 

The only difference is that you would let a non-authorized key fail implicitly in the `Signable`'s `verify_signature` function due to an [empty signatures field](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/common.py#L102) or an [empty key dictionary](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/common.py#L108), and I suggest to let it fail explicitly. 

Maybe `SignatureVerificationError` is not the best exception type, maybe it should be something like `AuthorizationError`. 

In any case I'd say it does not hurt to re-check the authorization here, although it's already done in the `import_step_metadata_from_files_as_dict` function.

in_toto/exceptions.py
```diff
@@ -25,3 +25,8 @@ class LinkNotFoundError(Error):
   """Indicates that a ran command exited with non-int or non-zero return
   value. """
   pass
+
+class AuthorizationError(Error):
+  """Indicates that a ran command exited with non-int or non-zero return
+  value. """
```
lukpueh: Please update the docstrings when you copy-paste. ;)

in_toto/verifylib.py
```diff
@@ -246,9 +246,13 @@ def verify_all_steps_signatures(layout, chain_link_dict):
...
 
       # Create the dictionary of keys for this step
-      for pubkeyid in step.pubkeys:
-          if pubkeyid == keyid:
-            keys_dict[pubkeyid] = layout.keys[pubkeyid]
+      # Only one key with the matching keyid in the
+      # filename is added to the dictionary which ensures
+      # that the link has been signed by that key
+      if keyid in step.pubkeys:
+            keys_dict[keyid] = layout.keys[keyid]
```
lukpueh: Indentation!

lukpueh (changes requested):
Please fix tiny issues and we can merge.

SantiagoTorres:
this LGTM

lukpueh (approved):

PR #86, opened 2017-04-25
Enable delegation

RohanBhirangi:


SantiagoTorres:
Before I continue reviewing, could you please fix the permission issues on the committed files?

in_toto/in_toto_verify.py
```diff
@@ -54,8 +56,16 @@ def in_toto_verify(layout_path, layout_key_paths):
...
+    log.info("Verifying software supply chain...")
 
+    log.info("Reading layout...")
+    layout = in_toto.models.layout.Layout.read_from_file(layout_path)
+
+    log.info("Reading layout key(s)...")
+    layout_key_dict = in_toto.util.import_rsa_public_keys_from_files_as_dict(
+        layout_key_paths)
+
+    verifylib.in_toto_verify(layout, layout_key_dict)
```
lukpueh, line 69: Please update function docstring to reflect your changes.

in_toto/models/layout.py
```diff
@@ -152,12 +152,19 @@ def import_step_metadata_from_files_as_dict(self):
...
       key_link_dict = {}
       for keyid in step.pubkeys:
+        filename = FILENAME_FORMAT.format(step_name=step.name, keyid=keyid)
         try:
-          link = models__link.Link.read_from_file(FILENAME_FORMAT.format(
-              step_name=step.name, keyid=keyid))
+          with open(filename, 'r') as fp:
+            file = json.load(fp)
         except IOError as e:
           pass
```
lukpueh: I know this is not part of this PR, but could you please add a comment here, that explains why we `pass` on an `IOError`.

in_toto/models/layout.py
```diff
@@ -152,12 +152,19 @@ def import_step_metadata_from_files_as_dict(self):
     for step in self.steps:
       key_link_dict = {}
       for keyid in step.pubkeys:
+        filename = FILENAME_FORMAT.format(step_name=step.name, keyid=keyid)
         try:
-          link = models__link.Link.read_from_file(FILENAME_FORMAT.format(
-              step_name=step.name, keyid=keyid))
+          with open(filename, 'r') as fp:
+            file = json.load(fp)
```
lukpueh: `file` is a builtin in Python2. If you don't have a good reason to override it, please don't!

in_toto/models/layout.py
```diff
@@ -152,12 +152,19 @@ def import_step_metadata_from_files_as_dict(self):
...
         except IOError as e:
           pass
         else:
+          if file["_type"] == "Link":
+            link = models__link.Link.read(file)
+          elif file["_type"] == "layout":
+            link = Layout.read(file)
+          else:
+            raise in_toto.exceptions.LinkNotFoundError("Invalid format".format())
```
lukpueh: Consider using some vertical spacing and comments in this function.

lukpueh: Please also update function docstring.

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+  """
+  <Purpose>
+    Checks if any step has been delegated by the functionary, recurses into
+    the delegation and replaces the layout object in the chain_link_dict
+    by an equivalent link object
+
+  <Arguments>
+    layout:
+            The layout specified by the project owner against which the
+            threshold will be verified.
```
lukpueh: Copy-pasta? :) Please adapt to this function or make more general.

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+      if link._type == "layout":
+        layout_key_dict = {}
+
+        for key, key_obj in six.iteritems(layout.keys):
+
+          # Extract the key object from the passed layout
+          if key == keyid:
+            # Extract the key object of the key corresponding
+            # to the link object
+            layout_key_dict[key] = key_obj
```
lukpueh: I think you can replace lines 1079:1087 with one line of code:
```python
layout_key_dict = {keyid: layout.keys.get(keyid)}
```
and a good comment about what it means using that key.

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+
+        # Make a recursive call to in_toto_verify with the
+        # layout and the extracted key object
+        return_link = in_toto_verify(link, layout_key_dict)
+
+        # Replace the layout object in the passed chain_link_dict
+        # with the link file returned by in-toto-verify
+        key_link_dict[keyid] = return_link
```
lukpueh: I guess you could also merge above two lines, but that's a matter of taste. 
Slightly more important, I think we can come up with a better (more descriptive) name than `return_link`.
What about "summary_link"? @SantiagoTorres, any suggestions?

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+
+  return chain_link_dict
+
+def get_return_link_file(layout, reduced_chain_link_dict):
```
lukpueh: See naming discussion above.

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+  """
+  <Purpose>
+    Merges the materials of the first step (as mentioned in the layout)
+    and the products of the last step and returns a new link file which
+    defines the entire software supply chain
```
lukpueh: Could you maybe rephrase this? I wouldn't say this link *defines* the supply chain, but rather it reports what came into it and what came out of it. 
Maybe you can also summarize the many discussions we had about this here in a sentence or two.

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+
+  <Arguments>
+    layout:
+            The layout specified by the project owner against which the
+            threshold will be verified.
```
lukpueh: And another gotcha!

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+  <Exceptions>
+    None.
+
+  <Side Effects>
+    None.
+
+  """
+
+  # Create a new link object to return
+  return_link = in_toto.models.link.Link()
```
lukpueh: As said above, it's not so interesting that we will *return* this link (whereto actually?) but rather that it stands for the entire supply chain.

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+
+  for step_name, link in six.iteritems(reduced_chain_link_dict):
```
lukpueh: You don't need to loop here, you could do
`reduced_chain_link_dict[layout.steps[0].name]` and `reduced_chain_link_dict[layout.steps[-1].name]` to get first and last item instead.

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+    # Check if this step is the last step in the layout
+    if step_name == layout.steps[len(layout.steps) - 1].name:
+      # Copy the products and other properties
+      # to the return link
+      return_link._type = link._type
+      return_link.name = link.name
+      return_link.products = link.products
+      return_link.byproducts = link.byproducts
+      return_link.command = link.command
+      return_link.return_value = link.return_value
```
lukpueh: I think it would be good to describe why certain values are take from the first and others from the last step. Either inline or in the function docstring.

in_toto/verifylib.py
```diff
@@ -1135,3 +1227,8 @@ def in_toto_verify(layout_path, layout_key_paths):
 
   # We made it this far without exception that means, verification passed.
   log.pass_verification("The software product passed all verification.")
+
+  # Since the verification is successful, return a link file which defines
+  # the entire software supply chain
+  return_link = get_return_link_file(layout, reduced_chain_link_dict)
+  return return_link
```
lukpueh: Please update the function's docstrings.

lukpueh (changes requested):
Looks good overall. Please address comments and I'll give you another round of reviews.

lukpueh:
Btw. @SantiagoTorres, @JustinCappos, and I decided (after receiving some feedback at DockerCon) that *"sublayouts"* might be a better term for/instead of *"delegations"*. Can you please refactor?

lukpueh:
Two more things I forgot to mention yesterday:
- Please always write a PR description (you can find [rudimentary instructions here](https://github.com/secure-systems-lab/code-style-guidelines/issues/6)).
- Please add unit tests for your two new functions

in_toto/models/layout.py
SantiagoTorres: (To disambiguate, just don't use `file` as a variable name here)

in_toto/verifylib.py
```diff
@@ -1043,8 +1044,105 @@ def reduce_chain_links(chain_link_dict):
...
+            The layout specified by the project owner against which the
+            threshold will be verified.
+
+    chain_link_dict:
+            A dictionary of key-link pair with step names as keys. For each
+            step name, there are one or more keyids and corresponding
+            link objects.
+
+  <Exceptions>
+    raises an Exception if verification of the delegated step fails.
```
SantiagoTorres, line 1060: I wonder if we want to have a more granular Exception taxonomy.

in_toto/verifylib.py
SantiagoTorres: I imagine it'd be also necessary to check if layout_key_dict is not none in case this change happens...

in_toto/verifylib.py
SantiagoTorres: `summary_link` sounds good to me.

in_toto/verifylib.py
SantiagoTorres: Also a minor caveat: it does not return a link file.

in_toto/verifylib.py
SantiagoTorres: I'm curious, has this been tested with 3-level delegations?

in_toto/verifylib.py
lukpueh: We could, but we don't have to, the format will be checked down the road.

SantiagoTorres: Right, and it would fail with a "this layout is of the wrong format" or something of that nature. I'm usually for failing early.

in_toto/verifylib.py
RohanBhirangi: Nope, this hasn't yet been tested with 3-level delegations. I've checked only 2-level delegations.

in_toto/models/layout.py
```diff
@@ -149,20 +151,38 @@ def import_step_metadata_from_files_as_dict(self):
...
+        filename = FILENAME_FORMAT.format(step_name=step.name, keyid=keyid)
+
+        # load the link object from the file
         try:
-          link = models__link.Link.read_from_file(FILENAME_FORMAT.format(
-              step_name=step.name, keyid=keyid))
+          with open(filename, 'r') as fp:
+            link_obj = json.load(fp)
+        # pass the IOError since all functionaries are
+        # not required to have performed the step (thresholds)
```
lukpueh: Nitpick:
I think the comment is ambiguous. What about:
```
# We try to load a link for every authorized functionary, but don't fail
# if the file does not exist (authorized != required)
# FIXME: Should we really pass on IOError, or just skip inexistent links
```

in_toto/models/layout.py
```diff
@@ -149,20 +151,38 @@ def import_step_metadata_from_files_as_dict(self):
...
+          # Check whether the object is of type link or layout
+          # and load it accordingly
+          if link_obj["_type"] == "Link":
+            link = models__link.Link.read(link_obj)
+          elif link_obj["_type"] == "layout":
+            link = Layout.read(link_obj)
+          else:
+            raise in_toto.exceptions.LinkNotFoundError("Invalid format".format())
           key_link_dict[keyid] = link
+
```
lukpueh, line 183: Vertical spacing?

in_toto/verifylib.py
lukpueh: The exceptions are of regular supply chain verification granularity (recursion). Or do you think they should be wrapped/masked by a SubLayoutException?

in_toto/verifylib.py
```diff
@@ -1043,45 +1044,136 @@ def reduce_chain_links(chain_link_dict):
...
+
+      # Extract keyid and link from the passed chain_link_dict
+      # Check if the link object is of type 'layout'
+      if link._type == "layout":
+        layout_key_dict = {}
+
+        # Retrieve the entire key object for the keyid
+        # corresponding to the link
+        layout_key_dict = {keyid: layout.keys.get(keyid)}
+
```
lukpueh, line 1082: @RohanBhirangi: Didn't you want to reply to [Santiago's `None`-check comment](https://github.com/in-toto/in-toto/pull/86#discussion_r114427572) here?

in_toto/verifylib.py
```diff
@@ -1043,45 +1044,136 @@ def reduce_chain_links(chain_link_dict):
...
+    A dictionary of key-link pair with step names as keys. For each
+    step name, there are one or more keyids and corresponding
+    link objects.
+  """
+
+  for step_name, key_link_dict in six.iteritems(chain_link_dict):
+
+    for keyid, link in six.iteritems(key_link_dict):
+
+      # Extract keyid and link from the passed chain_link_dict
```
lukpueh: This comment does not make sense here.

in_toto/verifylib.py
```diff
@@ -1043,45 +1044,136 @@ def reduce_chain_links(chain_link_dict):
...
+      # Check if the link object is of type 'layout'
```
lukpueh: This comment [isn't really necessary](https://github.com/secure-systems-lab/code-style-guidelines#comments).

in_toto/verifylib.py
```diff
@@ -1043,45 +1044,136 @@ def reduce_chain_links(chain_link_dict):
...
+  summary_link._type = first_step_link._type
+  summary_link.name = first_step_link.name
+
+  # Copy products, byproducts and command from the last step link
+  # as these properties would be used by the subsequent steps
+  summary_link.products = last_step_link.products
+  summary_link.byproducts = last_step_link.byproducts
+  summary_link.command = last_step_link.command
+  summary_link.return_value = last_step_link.return_value
+
```
lukpueh, line 1137: [Ping](https://github.com/in-toto/in-toto/pull/86/files/97c28e84fa3929b5165657a0df9123aab89ea719#r114184488)

in_toto/verifylib.py
```diff
@@ -1043,45 +1044,136 @@ def reduce_chain_links(chain_link_dict):
...
+  """
+
+  for step_name, key_link_dict in six.iteritems(chain_link_dict):
+
+    for keyid, link in six.iteritems(key_link_dict):
+
+      # Extract keyid and link from the passed chain_link_dict
+      # Check if the link object is of type 'layout'
+      if link._type == "layout":
+        layout_key_dict = {}
```
lukpueh: I think it would be nice to add a `log.info` here, saying that the step (layout) with `<name>` will be verified as sublayout.

in_toto/verifylib.py
```diff
@@ -1113,13 +1199,16 @@ def in_toto_verify(layout_path, layout_key_paths):
   log.info("Verifying link metadata signatures...")
   verify_all_steps_signatures(layout, chain_link_dict)
 
+  log.info("Verifying sublayouts...")
+  chain_link_dict = verify_sublayouts(layout, chain_link_dict)
```
lukpueh, line 1205: Please add this to the docstring actions above. Please also explain why sublayouts are verified at this place (after verify signatures and before the rest), you can explain this either in the docstring or as comment.

lukpueh (changes requested):
Thanks for your changes. Please address comments and add unittests!

in_toto/verifylib.py
SantiagoTorres: Hmm, I don't see a reason to have a specific delegation type. What do you think?

lukpueh: Me neither. I think it's fine like this. One thing that bugs me more is the inconsistent usage of  the `<Exception>` in function docstrings. Sometimes we only list exceptions that are actually raised in the very function, and sometimes we also raise exceptions that get raised in called functions (like here). I'm not sure which one is better.

in_toto/verifylib.py
RohanBhirangi: @SantiagoTorres I’m not sure if we really need it. The layout_key_dict would be None if we are not able to fetch the key object from the layout. But if the key object is not present in the layout we wouldn’t reach this step. Because we verify the layout signatures and step signatures before this step where we retrieve the key object from the layout. So if the key object is not present, we would fail early.

in_toto/verifylib.py
RohanBhirangi: I've tested this with 3-level delegations. Works fine! :)

in_toto/verifylib.py
```diff
@@ -1070,11 +1070,11 @@ def verify_sublayouts(layout, chain_link_dict):
 
   for step_name, key_link_dict in six.iteritems(chain_link_dict):
 
+    # Extract keyid and link from the passed chain_link_dict
```
lukpueh: This comment states the obvious (i.e. something that can be easily gathered from the code); and contradicts the code (which iterates over keyid, link pairs in a key_link_dict). Both bad commenting practices. Please re-phrase.

in_toto/verifylib.py
```diff
@@ -1039,45 +1040,142 @@ def reduce_chain_links(chain_link_dict):
...
+            verified later on.
+        6.  Verify alignment of defined (Step) and reported (Link) commands
             NOTE: Won't raise exception on mismatch
-        8.  Execute Inspection commands
+        7.  Execute Inspection commands
             NOTE: Inspections, similar to Steps executed with 'in-toto-run',
             will record materials before and products after command execution.
             For now it records everything in the current working directory.
-        9.  Verify rules defined in each Step's material_matchrules and
+        8.  Verify rules defined in each Step's material_matchrules and
```
lukpueh: You're missing the threshold verification step here.

Btw. I think "verify threshold equality" is not an ideal name, because you are not verifying the equality of thresholds but rather "link equality for a step with threshold greater one". You don't have to fix this here, but you could write a ticket and/or PR.

lukpueh (changes requested):
Please fix two minor comment problems and write add unit tests.

lukpueh:
Will be merged with https://github.com/in-toto/in-toto/pull/94. Thanks for contributions, @RohanBhirangi and reviews, @SantiagoTorres!

PR #89, opened 2017-05-04
Change LICENSE's copyright ownership.

baloian:
Hi @lukpueh and @SantiagoTorres , I just would like you to know that about two weeks ago
during developers meeting we decided (actually Justin decided) that all our lab's projects copyright
ownership should be New York University.
So, my responsibility is to check and if it needs modify LICENSE file.

- In this PR I just changed ownership from Santiago Torres to New York University.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 Santiago Torres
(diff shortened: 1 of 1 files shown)


SantiagoTorres:
💯 Thanks here too!

These licenses are auto-generated, so I imagine they may all be wrong...

PR #92, opened 2017-06-09
Add opensuse demo

ShikherVerma:
Adds a demo-opensuse directory with project owner and functionary keys and instructions on running the demo which showcase how in-toto can be used with opensuse software supply chain.


ShikherVerma:
@SantiagoTorres @lukpueh there is a really weird thing happening. The RPM contains `connman.spec`.  
The [original file](https://github.com/ShikherVerma/connman/blob/master/connman.spec) has the hash `419e3afa9add01181265a3f1cf71d0e831e0dfa2722fd3d6fbf5d64b780e738c` but if you go through the demo and `unrpm` the final `connman-1.30-1.1.src.rpm` which is available for download to the client, and recompute the hash it comes out to be different `a3c0b32450350b6a65f98af32d83d2fedf5473dbb63a0b6fb277d7c4bfc0a785`.

SantiagoTorres:
@ShikherVerma Thanks for submitting. Can you give me the RPM and the original file so I can check this out? I'm incredibly curious to see why this happens.

ShikherVerma:
These are the [original file](https://github.com/ShikherVerma/connman/blob/master/connman.spec) and built [src rpm](http://download.opensuse.org/repositories/home:/shikher/openSUSE_Factory/src/)
I took a `diff` and it seems the one in src.rpm has changelog appended to it and release number incremented. This is harmless but I think its a bad practise to have the shipped spec file different from the source one. Any malicious change to the spec file would also go unnoticed since the hashes don't match anyway.

demo-opensuse/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM opensuse
+MAINTAINER Shikher Verma
```
SantiagoTorres: MAINTAINER is [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated). Could you change to labels instead?

demo-opensuse/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN cd home/
+RUN zypper --non-interactive install git python-pip gcc python-devel libffi-devel libopenssl-devel vim build sudo osc wget
+RUN zypper --non-interactive install -t pattern devel_C_C++
```
SantiagoTorres: Is this required for the final functionary? Maybe this could be a cool place to try out [`multi-stage builds`](https://docs.docker.com/engine/userguide/eng-image/multistage-build/#before-multi-stage-builds) to save on the final container size.

SantiagoTorres: If not, you may want to remove the packages manually after you install pycrypto or whatever is pulling all of those C/C++ compilation dependencies.

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
+# in-toto openSUSE demo
+In this demo, we will use in-toto to secure a software supply chain for OpenSUSE. Bob is a developer for a project, Carl tests the software before packaging, and Alice oversees the project. So, using in-toto's names for the parties, Alice is the project owner - she creates and signs the software supply chain layout with her private key - and Bob and Carl are project functionaries - they carry out the steps of the software supply chain as defined in the layout.
```
SantiagoTorres: you may want to change the dashes for emdashes here for aesthetic.

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+For the sake of demonstrating in-toto, we will have you run all parts of the software supply chain. That is, you will perform the commands on behalf of Alice, Bob and Carl as well as the client who verifies the final product.
+
+# Setup
+Install docker
+https://docs.docker.com/engine/installation/
+
+Get in-toto
+```
+git clone https://github.com/in-toto/in-toto.git
```
SantiagoTorres: At this point I'm a little confused. Do we need in-toto locally if we are going to use the container?

demo-opensuse/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN (cd home/ && git clone https://github.com/in-toto/in-toto.git)
+RUN (cd home/in-toto/ && pip install --upgrade pip setuptools virtualenv)
+RUN (cd home/in-toto/ && virtualenv -p /usr/bin/python2.7 in-toto)
+RUN (cd home/in-toto && source in-toto/bin/activate && pip install -r requirements.txt)
+RUN (cd home/in-toto && source in-toto/bin/activate && pip install -e .)
```
SantiagoTorres: I wonder if we need a venv on the container. I have a feeling that we can install these system-wide given that the container is just serving the purpose of running in-toto commands. It also let's us drop the source venv.

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+```
+Run the docker image
+```
+docker run --privileged -i -t in-toto-demo-opensuse
+```
+
+# Change into the demo directoy and you are ready to start
+```
+cd home/in-toto/
+source in-toto/bin/activate/
```
SantiagoTorres: I wonder if we want to have this in the .bashrc or profile so the user doesn't have to do this.

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+# └── owner_alice
+#     ├── alice
+#     ├── alice.pub
+#     └── create_layout.py
+```
+
+### Define software supply chain layout (Alice)
+First, we will create a new package on open build server and we will need to define the software supply chain layout. To simplify this process, we provide a script that generates a simple layout for the purpose of the demo.
+
+In this software supply chain layout, we have Alice, who is the project owner that creates the layout, Bob, who clones the project's repo and performs some pre-packaging editing (update changelog), Carl, then tests the build and verifies that its fit to ship, Carl then commits the package which triggers open build server to build the RPMs.
```
SantiagoTorres: I suggest we split this long sentence into a series of shorter, simpler ones.

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+First, signup for an account on opensuse build service at
+https://build.opensuse.org/ICSLogin/
```
SantiagoTorres: this may overcomplicate this, but is this service opensource? could we host this inside the container itself?

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+Export your username as environment variable so that you can copy paste the demo commands
```
SantiagoTorres: You are *setting* the env variable, not *exporting* on the command below (unless the zsh behavior is different from bash)

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+osc meta prj -e home:$username
+```
+
+This will open a template xml in your favourite editor. For this demo uncomment the first one.
+```
+<repository name="openSUSE_Factory">
+   <path project="openSUSE:Factory" repository="standard" />
+   <arch>x86_64</arch>
+   <arch>i586</arch>
+ </repository>
```
SantiagoTorres, line 92: I'm thinking that these operations should also be performed by a functionary (maybe bob?)

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+`clone` on his behalf, that is we use in-toto to clone the project repo from GitHub and
+record metadata for what we do. Execute the following commands to change to Bob's
+directory and perform the step.
+
+```
+cd ../functionary_bob
+mv ../owner_alice/home:$username/ ./
+cd home:$username
+in-toto-record --step-name clone --key ../bob start
+git clone https://github.com/shikherverma/connman.git connman-src
```
SantiagoTorres: Is there a reason we need to clone from the fork and not upstream?

SantiagoTorres: Also, now I see why we need the C++ toolchain 💯

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+mv connman-src/* connman/
+rm -r connman-src/
+in-toto-record --step-name clone --key ../bob stop --products connman/_service connman/connman-1.30.tar.gz connman/connman-1.30.tar.sign connman/connman-rpmlintrc connman/connman.changes connman/connman.keyring connman/connman.spec
```
SantiagoTorres: Another alternative would be to already have a `git` listener with the repository cloned so we don't have to access the network.

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+Bob has done his work and can send over the sources to Carl.
+```
+cd ..
+mv home:$username/ ../functionary_carl/
+```
+
+### Test (Carl)
+Now, we will perform Carl’s `test` step by executing the following commands to change to Carl's directory.
+
+```
```
SantiagoTorres: I wonder if we want to set bash syntax in this and the previous examples.

demo-opensuse/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+This command will verify that
+ 1. the layout has not expired,
+ 2. was signed with Alice’s private key, and that according to the definitions in the layout
+ 3. each step was performed and signed by the authorized functionary
+ 4. the recorded materials and products align with the matchrules and
+ 5. the inspection `unpack` finds what it expects.
+
+From it, you will see the meaningful output `PASSING` and a return value
+of `0`, that indicates verification worked out well:
```
SantiagoTorres: s/worked out well/verification was successful/ or so.

SantiagoTorres (changes requested):
Looking great! Thanks for sending this over 👍

demo-opensuse/Dockerfile
ShikherVerma: Yeah we can drop the venv.

demo-opensuse/README.md
ShikherVerma: Not really. The docker file is in the repo so users can directly download just the dockerfile and start from there. Should I put a link to docker file instead ?

demo-opensuse/README.md
ShikherVerma: Yes thats better than asking them to cd into it.

demo-opensuse/README.md
ShikherVerma: We can host our own open build server http://openbuildservice.org/download/ but I though going through the opensuse one would give a better feeling of how in-toto integrated into opensuse software supply chain.

demo-opensuse/README.md
ShikherVerma: you are right. its setting not exporting. typo error.

demo-opensuse/README.md
ShikherVerma: the fork is just https://build.opensuse.org/package/show/network:utilities/connman# put in a git repo. Getting it from build.opensuse.org also dowloaded `osc` metadata which we then had to remove. Using a fork makes the demo cleaner.

demo-opensuse/README.md
ShikherVerma: Yeah we can do that. but the demo already depends on network due to integrating openSUSE's build server. Also we would need the files in the Container then.

demo-opensuse/README.md
ShikherVerma: okay

demo-opensuse/Dockerfile
ShikherVerma: yup these are required during the demo. can't drop them.

demo-opensuse/README.md
ShikherVerma: moved to bob

ShikherVerma:
rebased from `develop`

demo-opensuse/README.md
SantiagoTorres: I thought I had mentioned that we may want to deploy that in the container too.

demo-opensuse/README.md
```diff
@@ -0,0 +1,242 @@
+# in-toto openSUSE demo
+In this demo, we will use in-toto to secure a software supply chain for OpenSUSE. Bob is a developer for a project, Carl tests the software before packaging, and Alice oversees the project. So, using in-toto's names for the parties, Alice is the project owner — she creates and signs the software supply chain layout with her private key — and Bob and Carl are project functionaries — they carry out the steps of the software supply chain as defined in the layout.
+
+For the sake of demonstrating in-toto, we will have you run all parts of the software supply chain. That is, you will perform the commands on behalf of Alice, Bob and Carl as well as the client who verifies the final product.
+
+# Setup
+Install docker
+https://docs.docker.com/engine/installation/
+
+Get in-toto
```
SantiagoTorres: Can't we replace all of these with "do a docker pull of the image at in-toto/opensuse-demo" or so?

demo-opensuse/README.md
```diff
@@ -0,0 +1,242 @@
...
+# Change into the demo directory and you are ready to start
+```shell
+cd demo-opensuse/
+```
+
+Inside the demo-opensuse directory you will find four directories: `owner_alice`,
+`functionary_bob`, `functionary_carl` and `final_product`. Alice, Bob and Carl
+already have RSA keys in each of their directories. This is what you see:
+```shell
+tree  # If you don't have tree, try 'find .' instead
```
SantiagoTorres, line 35: I'm wondering if it's a good idea to pre-install tree in the container if we are going to ask them to run this command.

demo-opensuse/README.md
SantiagoTorres: Are the rules for bob's step containing a `CREATE template.xml` or so?

demo-opensuse/README.md
```diff
@@ -0,0 +1,242 @@
...
+```shell
+osc up
+```
+
+Now, we will take the role of the functionary Bob and perform the step
+`clone` on his behalf, that is we use in-toto to clone the project repo from GitHub and
+record metadata for what we do.
+```shell
+in-toto-record --step-name clone --key ../bob start
+git clone https://github.com/shikherverma/connman.git connman-src
```
SantiagoTorres: ping on this. Do we really need to clone from your fork?

demo-opensuse/owner_alice/create_layout.py
```diff
@@ -0,0 +1,99 @@
...
+        ],
+        "product_matchrules": [
+            ["MATCH", "connman-1.30-1.1.src.rpm", "WITH", "PRODUCTS", "FROM", "package"],
+            ["MATCH", "connman-1.30.tar.gz", "WITH", "PRODUCTS", "IN", "connman", "FROM", "clone"],
+            ["MATCH", "_service", "WITH", "PRODUCTS", "IN", "connman", "FROM", "clone"],
+            ["MATCH", "connman-1.30.tar.sign", "WITH", "PRODUCTS", "IN", "connman", "FROM", "clone"],
+            ["MATCH", "connman-rpmlintrc", "WITH", "PRODUCTS", "IN", "connman", "FROM", "clone"],
+            ["MATCH", "connman.changes", "WITH", "PRODUCTS", "IN", "connman", "FROM", "update-changelog"],
+            ["MATCH", "connman.keyring", "WITH", "PRODUCTS", "IN", "connman", "FROM", "clone"],
+            # FIXME: very weird that hash doesn't match for this file
```
SantiagoTorres: Now that we know why this happens, maybe it's worth mentioning it instead.

demo-opensuse/owner_alice/create_layout.py
```diff
@@ -0,0 +1,99 @@
...
+        "expected_command": "",
+        "threshold": 1,
+      }],
+    "inspect": [{
+        "name": "unpack",
+        "material_matchrules": [
+            ["MATCH", "connman-1.30-1.1.src.rpm", "WITH", "PRODUCTS", "FROM", "package"],
+            # FIXME: If the routine running inspections would gather the
+            # materials/products to record from the rules we wouldn't have to
+            # ALLOW other files that we aren't interested in.
```
SantiagoTorres, line 78: @lukpueh thoughts on this situation?

demo-opensuse/owner_alice/create_layout.py
```diff
@@ -0,0 +1,99 @@
...
+        "product_matchrules": [],
+        "pubkeys": [key_carl["keyid"]],
+        "expected_command": "osc build openSUSE_Factory x86_64 connman/connman.spec",
+        "threshold": 1,
+      },{
+        "name": "package",
+        "material_matchrules": [
+            ["MATCH", "connman/_service", "WITH", "PRODUCTS", "FROM", "clone"],
+            ["MATCH", "connman/connman-1.30.tar.gz", "WITH", "PRODUCTS", "FROM", "clone"],
+            ["MATCH", "connman/connman-1.30.tar.sign", "WITH", "PRODUCTS", "FROM", "clone"],
```
SantiagoTorres, line 59: I'm wondering if we can do *anything* with this detached signature (e.g., check that the hash inside matches the hash reported by connman in this link)

SantiagoTorres (changes requested):
Thanks! Here are some other little details. It's getting there! 👍

demo-opensuse/README.md
ShikherVerma: will add a rule for it.

demo-opensuse/Offline-demo.md
```diff
@@ -0,0 +1,157 @@
+# in-toto offline openSUSE demo
+You could try this demo if you don't have or don't want to create an OpenSUSE build service account.
```
SantiagoTorres: s/could/can/

Maybe elaborate that you simulate what the OSC server does in the background by doing rpmbuild.

demo-opensuse/Offline-demo.md
```diff
@@ -0,0 +1,157 @@
...
+echo $?
+# should output 0
+```
+
+### Wrapping up
+Congratulations! You have completed the in-toto opensuse demo!
+
+This exercise shows a very simple case in how in-toto can protect the different steps within the software supply chain. More complex software supply chains that contain more steps can be created in a similar way. You can read more about what in-toto protects against and how to use it on [in-toto's Github page](https://in-toto.github.io/).
+
+### Clean up
```
SantiagoTorres: Do we need this in this case? the contianer will be "dropped" once you leave.

demo-opensuse/README.md
```diff
@@ -46,21 +47,15 @@ tree  # If you don't have tree, try 'find .' instead
...
-In this software supply chain layout, we have Alice, Bob and Carl. Alice is the project owner that creates the root layout. Bob, is the developer who clones the project's repo and performs some pre-packaging editing. Carl is the builder then tests the build and verifies that its fit to ship. Carl then commits the package which triggers open build server to build the RPMs.
-
-First, signup for an account on opensuse build service at
-https://build.opensuse.org/ICSLogin/
-
-Set your username as environment variable so that you can copy paste the demo commands
-```shell
-username=<username>
-```
+In this software supply chain layout, we have Alice, Bob and Carl. Alice is the project owner that creates the root layout. Bob, is the developer who clones the project's repo and performs some pre-packaging editing. Carl is the builder who then tests the build and verifies that its fit to ship. Carl then commits the package which triggers open build server to build the RPMs.
```
SantiagoTorres: Three people participate in this software supply chain, Alice, Bob and Carl. Alice is the project owner that creates the root layout. Bob is the developer, and he clones the project's repo and performs some pre-packaging edits. Carl then tests the sources and verifies that its fit to ship. Carl then commits the package which triggers open build server to build the RPMs.

demo-opensuse/Offline-demo.md
```diff
@@ -0,0 +1,157 @@
...
+#     ├── alice.pub
+#     ├── create_layout.py
+#     ├── create_layout_offline.py
+#     └── verify-signature.sh
+```
+
+### Define software supply chain layout (Alice)
+We will define the software supply chain layout. To simplify this process, we provide a script that generates a simple layout for the purpose of the demo.
+
+In this software supply chain layout, we have Alice, Bob and Carl. Alice is the project owner that creates the root layout. Bob, is the developer who clones the project's repo and performs some pre-packaging editing. Carl is the builder who then tests the build and verifies that its fit to ship. Carl then builds the RPM.
```
SantiagoTorres: > Like in the on-line example, Alice, Bob and Carl will carry out the steps in this software supply chain. Alice is the project owner that creates the root layout. Bob, is the developer who clones the project's repo and performs some pre-packaging edits. Carl then builds the sources and verifies that its fit to ship. Carl then packages the built binary into an RPM.

SantiagoTorres:
Thanks! 👍

ShikherVerma:
yaay

PR #95, opened 2017-06-13
Add in_toto_mock

### LINKED ISSUE #93: Add `in-toto-mock` command line utility
in-toto steps are usually executed using the [`in-toto-run`](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py) command line utility, whose functionality is implemented in the according library function [`runlib.in_toto_run`](https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L330-L403).

Since this command requires a signing key and a bunch of command line arguments, it would be helpful to have a stripped down variant of that command that can be used to easily mock `in-toto-run`'s metadata generation aspect, without the need to specify a key and knowing all the command line arguments. 

This is particularly interesting for in-toto's [layout tool](https://github.com/in-toto/layout-web-tool), where we ask un-experienced in-toto users to (mock) run their supply chain, so that we can create a layout using the (mock) link metadata.

Here is an example usage of the envisioned command:
```
in-toto-mock -n <step name> -- <step command>
```
The resulting link metadata `<step name>.link-mock` is not signed and records materials and products at the current working directory where the command is executed.

I think `in-toto-run`'s `--verbose` and `--record-byproducts` flags should be on by default.

UPDATE:
As mentioned above `in-toto-mock` most likely will be used to reverse generate in-toto layouts with an initial set of custom artifact rules. Therefor it might be helpful to store the full path of the current working directory in resulting mock link files.
I think for now we should only store it when running `in-toto-mock` but make it so that we can easily opt in later to also store it when using `in-toto-run`.

ShikherVerma:
* **What does this PR do?**
Adds in_toto_mock command, which is a stripped down version of in_toto_run.

* **Are there points in the code the reviewer needs to double check?**
No, its basically a rip off of [in-toto-run](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py). Opened both files in [Meld](http://meldmerge.org/) and copy pasted shamelessly :stuck_out_tongue: 

@lukpueh 

* **Does this PR meet the acceptance criteria?**

   - Documentation created/updated (e.g., did you update the docstrings?)
      * README.md requires editing.

   - All builds are passing (did you run the tests locally?)
      * Yes

   - Code follows the [style guide](https://github.com/secure-systems-lab/code-style-guidelines)
      * Followed manually. Is there some code formatter you are using ? like [yapf](https://github.com/google/yapf).

* **What are the relevant issue numbers fixed (add one line for each)?**
Fixed #93 

* **Steps Required**
- [x] Add in-toto-mock and tests
- [x]  Fix code formatting issues
- [x] Check possible error in tests. In [test_in_toto_run#L106](https://github.com/in-toto/in-toto/blob/develop/test/test_in_toto_run.py#L106) it seems to be a minor copy pasting error, I guess it should be `in-toto-run` instead of `in-toto-record`, [test_in_toto_run#L148](https://github.com/in-toto/in-toto/blob/develop/test/test_in_toto_run.py#L148) again should be run instead of record. I have left it in _in_toto_mock test too, if its an error I'll rebase and edit that commit.
- [ ] Update README.md

### DIFF
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -15,7 +15,8 @@
 <Purpose>
   Stripped down variant of in-toto-run command that can be used to mock metadata
   generation of in-toto-run, without the need to specify a key and knowing all
-  the command line arguments.
+  the command line arguments. Generated MockLink is unsigned and includes working
+  directory, byproducts and used current directory as material and products.
 
   Provides a command line interface which takes any link command of the software
   supply chain as input and generates mock in_toto metadata.
@@ -36,20 +37,16 @@ import os
 import sys
 import argparse
 from in_toto import (util, runlib, log)
-from in_toto.models.link import Link
-from in_toto.models.link import MOCK_FILENAME_FORMAT
 
 def in_toto_mock(name, link_cmd_args):
   """
   <Purpose>
-    Calls runlib.in_toto_run with current directory as material_list and
-    product_list. Without signing key and record_byproducts as True.
-    Writes the returned link to disk. And catches exceptions.
+    Calls runlib.in_toto_mock with name and link_cmd_args.
 
   <Arguments>
     name:
-            A unique name to relate link metadata with a step defined in the
-            layout.
+            A unique name to relate mock link metadata with a step defined
+            in the layout.
     link_cmd_args:
             A list where the first element is a command and the remaining
             elements are arguments passed to that command.
@@ -58,7 +55,6 @@ def in_toto_mock(name, link_cmd_args):
     SystemExit if any exception occurs
 
   <Side Effects>
-    Writes newly created mock link metadata file to disk "<name>.link-mock"
     Calls sys.exit(1) if an exception is raised
 
   <Returns>
@@ -66,12 +62,7 @@ def in_toto_mock(name, link_cmd_args):
   """
 
   try:
-    mock_link = runlib.in_toto_run(name, material_list=['.'],
-        product_list=['.'], link_cmd_args=link_cmd_args, record_byproducts=True)
-    mock_fn = MOCK_FILENAME_FORMAT.format(step_name=name)
-    log.info("Storing mock link metadata to '{}'...".format(mock_fn))
-    mock_link.dump(filename=mock_fn)
-
+    runlib.in_toto_mock(name, link_cmd_args)
   except Exception as e:
     log.error("in toto mock - {}".format(e))
     sys.exit(1)
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -27,8 +27,7 @@
 
   Example Usage
   ```
-  in-toto-mock --name write-code --materials . --products . --by-products \
-      -- touch foo.py
+  in-toto-mock --name write-code -- touch foo.py
   ```
 
 """
@@ -40,50 +39,39 @@ from in_toto import (util, runlib, log)
 from in_toto.models.link import Link
 from in_toto.models.link import MOCK_FILENAME_FORMAT
 
-def in_toto_mock(name, material_list, product_list, link_cmd_args,
-    record_byproducts):
+def in_toto_mock(name, link_cmd_args):
   """
   <Purpose>
-    Calls runlib.in_toto_run without signing key and writes the returned link
-    to disk. And catches exceptions.
+    Calls runlib.in_toto_run with current directory as material_list and
+    product_list. Without signing key and record_byproducts as True.
+    Writes the returned link to disk. And catches exceptions.
 
   <Arguments>
     name:
             A unique name to relate link metadata with a step defined in the
             layout.
-    material_list:
-            List of file or directory paths that should be recorded as
-            materials.
-    product_list:
-            List of file or directory paths that should be recorded as
-            products.
     link_cmd_args:
             A list where the first element is a command and the remaining
             elements are arguments passed to that command.
-    record_byproducts:
-            A bool that specifies whether to redirect standard output and
-            and standard error to a temporary file which is returned to the
-            caller (True) or not (False).
 
   <Exceptions>
     SystemExit if any exception occurs
 
   <Side Effects>
+    Writes newly created mock link metadata file to disk "<name>.link-mock"
     Calls sys.exit(1) if an exception is raised
 
   <Returns>
     None.
   """
 
-  """Runs passed command, storing its materials, products and by-products into
-  mock link metadata file. The mock link metadata file is stored to disk. """
-
   try:
-    mock_link = runlib.in_toto_run(name, material_list, product_list,
-        link_cmd_args, key=False, record_byproducts=record_byproducts)
+    mock_link = runlib.in_toto_run(name, material_list=['.'],
+        product_list=['.'], link_cmd_args=link_cmd_args, record_byproducts=True)
     mock_fn = MOCK_FILENAME_FORMAT.format(step_name=name)
     log.info("Storing mock link metadata to '{}'...".format(mock_fn))
     mock_link.dump(filename=mock_fn)
+
   except Exception as e:
     log.error("in toto mock - {}".format(e))
     sys.exit(1)
@@ -98,10 +86,7 @@ def main():
 
   parser.usage = ("\n"
       "%(prog)s  --name <unique step name>\n{0}"
-               "[--materials <filepath>[ <filepath> ...]]\n{0}"
-               "[--products <filepath>[ <filepath> ...]]\n{0}"
-               "[--record-byproducts]\n{0}"
-               "[--verbose] -- <cmd> [args]\n\n"
+               " -- <cmd> [args]\n\n"
                .format(lpad))
 
   in_toto_args = parser.add_argument_group("in-toto options")
@@ -109,19 +94,6 @@ def main():
   # FIXME: Do we limit the allowed characters for the name?
   in_toto_args.add_argument("-n", "--name", type=str, required=True,
       help="Unique name for link metadata")
-
-  in_toto_args.add_argument("-m", "--materials", type=str, required=False,
-      nargs='+', help="Files to record before link command execution")
-
-  in_toto_args.add_argument("-p", "--products", type=str, required=False,
-      nargs='+', help="Files to record after link command execution")
-
-  in_toto_args.add_argument("-b", "--record-byproducts",
-      help="If set redirects stdout/stderr and stores to link metadata",
-      dest="record_byproducts", default=False, action="store_true")
-
-  in_toto_args.add_argument("-v", "--verbose", dest="verbose",
-      help="Verbose execution.", default=False, action="store_true")
 
   # FIXME: This is not yet ideal.
   # What should we do with tokens like > or ; ?
@@ -131,11 +103,9 @@ def main():
   args = parser.parse_args()
 
   # Turn on all the `log.info()` in the library
-  if args.verbose:
-    log.logging.getLogger().setLevel(log.logging.INFO)
+  log.logging.getLogger().setLevel(log.logging.INFO)
 
-  in_toto_mock(args.name, args.materials, args.products,
-      args.link_cmd, args.record_byproducts)
+  in_toto_mock(args.name, args.link_cmd)
 
 if __name__ == "__main__":
   main()
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python
+"""
+<Program Name>
+  in_toto_mock.py
+
+<Author>
+  Shikher Verma <root@shikherverma.com>
+
+<Started>
+  June 12, 2017
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Stripped down variant of in-toto-run command that can be used to mock metadata
+  generation of in-toto-run, without the need to specify a key and knowing all
+  the command line arguments.
+
+  Provides a command line interface which takes any link command of the software
+  supply chain as input and generates mock in_toto metadata.
+
+  in_toto_mock options are separated from the command to be executed by
+  a double dash.
+
+  The implementation of the tasks can be found in runlib.
+
+  Example Usage
+  ```
+  in-toto-mock --name write-code --materials . --products . --by-products \
+      -- touch foo.py
+  ```
+
+"""
+
+import os
+import sys
+import argparse
+from in_toto import (util, runlib, log)
+from in_toto.models.link import Link
+from in_toto.models.link import MOCK_FILENAME_FORMAT
+
+def in_toto_mock(name, material_list, product_list, link_cmd_args,
+    record_byproducts):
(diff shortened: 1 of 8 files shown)


ShikherVerma:
Amend commit. Add test to fail `in_toto_mock`.

in_toto/in_toto_mock.py
```diff
@@ -0,0 +1,111 @@
...
+            layout.
+    link_cmd_args:
+            A list where the first element is a command and the remaining
+            elements are arguments passed to that command.
+
+  <Exceptions>
+    SystemExit if any exception occurs
+
+  <Side Effects>
+    Calls sys.exit(1) if an exception is raised
```
lukpueh, line 58: I'd say, dumping the link file to disk is definitely a side effect.

in_toto/in_toto_mock.py
```diff
@@ -0,0 +1,111 @@
...
+
+  <Returns>
+    None.
+  """
+
+  """Runs passed command, storing by-products and current directory as materials
+  and products. The mock link metadata file generated is stored to disk. """
```
lukpueh: You don't need two types of docstrings.

in_toto/in_toto_mock.py
```diff
@@ -0,0 +1,111 @@
...
+  try:
+    mock_link = runlib.in_toto_run(name, material_list=['.'],
+        product_list=['.'], link_cmd_args=link_cmd_args, record_byproducts=True)
+    mock_fn = MOCK_FILENAME_FORMAT.format(step_name=name)
+    log.info("Storing mock link metadata to '{}'...".format(mock_fn))
+    mock_link.dump(filename=mock_fn)
```
lukpueh: I'd add a blank line here for readability.

lukpueh (changes requested):
Awesome! Exactly what I had in mind. Please address minor suggestions and this is good to merge.

SantiagoTorres:
Awesome PR description 👍

lukpueh:
Please read the issue update in https://github.com/in-toto/in-toto/issues/93 and add the requested functionality.

vladimir-v-diaz:
If there is anything in the PR summary that works well, we should consider using it for the lab's [pull request template](https://github.com/blog/2111-issue-and-pull-request-templates)

test/test_in_toto_mock.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python
+
+"""
+<Program Name>
+  test_in_toto_mock.py
+
+<Author>
+  Shikher Verma <root@shikherverma.com>
```
SantiagoTorres, line 8: This username 🎉

in_toto/in_toto_mock.py
ShikherVerma: Yeah :sweat_smile: definitely.

ShikherVerma: too many side effects. lets port everything to Haskell :stuck_out_tongue:

in_toto/in_toto_mock.py
ShikherVerma: Removing the second one.

test/test_in_toto_mock.py
ShikherVerma: mah domain mah username :laughing:

ShikherVerma:
@lukpueh @SantiagoTorres Check possible error in tests. In [test_in_toto_run#L106](https://github.com/in-toto/in-toto/blob/develop/test/test_in_toto_run.py#L106) it seems to be a minor copy pasting error, I guess it should be `in-toto-run` instead of `in-toto-record`, [test_in_toto_run#L148](https://github.com/in-toto/in-toto/blob/develop/test/test_in_toto_run.py#L148) again should be run instead of record. I have left it in _in_toto_mock test too, if its an error I'll rebase and edit that commit.

lukpueh:
[test_in_toto_run#L106](https://github.com/in-toto/in-toto/blob/develop/test/test_in_toto_run.py#L106) is not really an error, just a dubious test case. :)
But you are right, that line has no business there.
 [test_in_toto_run#L148](https://github.com/in-toto/in-toto/blob/develop/test/test_in_toto_run.py#L148) seems to be indeed a copy paste comment bug. 

Mind fixing both?

ShikherVerma:
fixing them.

in_toto/in_toto_mock.py
ShikherVerma: I added a blank line but this comment is not being marked outdated by github

ShikherVerma:
oh I broke some tests :(

lukpueh:
Yep, probably related to the field you added to links. We use some [demo link files](https://github.com/in-toto/in-toto/tree/develop/test/demo_files) for testing that don't have those fields yet.

in_toto/runlib.py
```diff
@@ -315,20 +318,26 @@ def create_link_metadata(link_name, materials_dict=None, products_dict=None,
...
+
   link_dict = {
     "name" : link_name,
     "materials" : materials_dict,
     "products" : products_dict,
     "command" : link_cmd_args,
     "byproducts" : byproducts,
-    "return_value" : return_value
+    "return_value" : return_value,
+    "working_directory" : working_directory
```
ShikherVerma: @lukpueh so should I update those test files or update this line to add working_directory only for link-mock and not keep empty field in link.

lukpueh:
Actually, I think I'd prefer to not have the `cwd` field in regular link files. Maybe we should have a subclass of `Link`, e.g. `MockLink` that contains that field.

ShikherVerma:
hmm I can add that line conditionally. lets keep a single class until there are more differences between them. What do you think ?

in_toto/runlib.py
lukpueh: Updating this line won't be enough, you'll still have the (empty) field in the JSON.

ShikherVerma: I can make it conditional in link.py. Only add working_directory if its not empty.

lukpueh: If you mean to conditionally assign the attribute in `Link`, that would not change anything, each object will sill always have that attribute, and in consequence the (empty) field in the JSON.

You could ignore the field when converting the class to JSON. But (if you want to do it properly) you'd have to write something like [`attr`'s as_dict](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/common.py#L40-L41) function which currently does this job. 

Which on the other hand wouldn't be too bad, because [it's not the huge](https://github.com/python-attrs/attrs/blob/1b419e304b01bbbcb1d0d7b0515efdbc7de73588/src/attr/_funcs.py#L10-L67) and above all that function is pretty much the only thing that currently keeps us from removing that dependency.

in_toto/models/mock_link.py
```diff
@@ -0,0 +1,97 @@
...
+  See LICENSE for licensing information.
+
+<Purpose>
+  Provides a class for mock link metadata which is information gathered when a
+  step of the supply chain is run as mock.
+"""
+
+import attr
+import json
+from . import link
```
lukpueh: Please don't use relative imports but rather `import in_toto.models.link`.

in_toto/models/mock_link.py
```diff
@@ -0,0 +1,97 @@
...
+MOCK_FILENAME_FORMAT = "{step_name}.link-mock"
+
+@attr.s(repr=False, init=False)
+class MockLink(link.Link):
+  """
+  A mock link is the metadata representation of a supply chain step mocked
+  by a functionary.
+
+  Mock links are recorded and stored to a file when a functionary
+  wraps a command with in_toto-run.
```
lukpueh: Rather `in-toto-mock`, right?

in_toto/models/mock_link.py
```diff
@@ -0,0 +1,97 @@
...
+  def __init__(self, **kwargs):
+    super(MockLink, self).__init__(**kwargs)
+
+    self._type = "MockLink"
+    self.name = kwargs.get("name")
+    self.materials = kwargs.get("materials", {})
+    self.products = kwargs.get("products", {})
+    self.byproducts = kwargs.get("byproducts", {})
+    self.command = kwargs.get("command", [])
+    self.return_value = kwargs.get("return_value", None)
```
lukpueh: Actually, you can let the inherited attributes be assigned in the parent constructor.

in_toto/runlib.py
```diff
@@ -327,6 +328,67 @@ def create_link_metadata(link_name, materials_dict=None, products_dict=None,
...
+
+  mock_link_dict = {
+    "name" : name,
+    "materials" : materials_dict,
+    "products" : products_dict,
+    "command" : link_cmd_args,
+    "byproducts" : byproducts,
+    "return_value" : return_value,
+    "working_directory" : os.getcwd()
+  }
```
lukpueh: You could still use `create_link_metadata` and add `cwd` upon return.

test/test_in_toto_mock.py
```diff
@@ -0,0 +1,103 @@
...
+      pass
+
+  def test_main_required_args(self):
+    """Test CLI command with required arguments. """
+
+    args = [ "in_toto_mock.py", "--name", self.test_step, "--", "echo", "test"]
+    with patch.object(sys, 'argv', args):
+      in_toto_mock_main()
+
+    self.assertTrue(os.path.exists(self.test_link))
```
lukpueh, line 76: You maybe want to clean up (remove the file) after running the test. So it does not interfere with later calls to `path.exists`. (Also below).

in_toto/in_toto_mock.py
```diff
@@ -0,0 +1,101 @@
...
+<Started>
+  June 12, 2017
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Stripped down variant of in-toto-run command that can be used to mock metadata
+  generation of in-toto-run, without the need to specify a key and knowing all
+  the command line arguments.
```
lukpueh: Maybe you want to mention how it differs from `in-toto-run` (what defaults it uses).

lukpueh (changes requested):
Looking good! Please address some minor comments and we are good to merge..

in_toto/runlib.py
ShikherVerma: I tried that first but `create_link_metadata` doesn't return a dictionary it returns a link. I tried adding option to get back mock link but it made the function ugly. I could change it to return dictionary but that would break other parts of code.

lukpueh: Right, my bad. Maybe you could check if we should get rid of `create_link_metadata` altogether?  IIRC, the constructors should do pretty much the same work.

test/test_in_toto_mock.py
ShikherVerma: I think its being removed somehow. (no idea how :stuck_out_tongue:,) I tested it. Put a `    self.assertTrue(os.path.exists(self.test_link))` in the beginning of next test and it failed. Its not being removed manually in `test_in_toto_run.py` too. There is `tearDown` function that gets called.

in_toto/runlib.py
ShikherVerma: Yes I think we should get rid of the create_link_metadata function. Its adding more lines rather than reducing. Will submit a separate pr for that.

SantiagoTorres: Followup here. We are rebasing this on #100 right?

test/test_in_toto_mock.py
SantiagoTorres: Yeah, the `TearDown` should be taking care of it, but be wary of doing testing the way you did, as it is still assuming that tests will run in a certain order.

ShikherVerma: oh I see. I checked though, added print statements to verify that the tests are being run in order.

in_toto/runlib.py
ShikherVerma: We can use the constructor for MockLink too. I guess I should cherry pick commit here itself since it impacts in-toto-mock also.

SantiagoTorres: > We can use the constructor for MockLink too. 

Yeah, I'd imagine make the constructor a simple call to super + setting up the cwd would work perfectly 👍

ShikherVerma:
I can't understand why coverage decreased :/ need help

ShikherVerma:
Coverage decreased as loc reduced.

ShikherVerma:
I guess only updating the `README.md` is left now.

lukpueh:
Is it correct to assume that you do "a lot of" history re-writing? If that's the case please don't while being under review. It makes re-reviews harder. :)

in_toto/in_toto_mock.py
lukpueh: So, you might want to list it in the side effects section?

ShikherVerma:
Yeah generally rewrite history to minimize small small commits. Won't do it while in review.

lukpueh:
For me small commits that show me that/ and how you addressed my comments are helpful. But before you PR you can do all the history cosmetics you like. :)

in_toto/models/mock_link.py
```diff
@@ -0,0 +1,63 @@
...
+  def __init__(self, **kwargs):
+    super(MockLink, self).__init__(**kwargs)
+
+    self._type = "MockLink"
+    self.working_directory = kwargs.get("working_directory", None)
+
+
+  def dump(self):
+    """
+    Write pretty printed JSON represented of self to a file. with filename.
```
lukpueh: -`.`

in_toto/models/mock_link.py
```diff
@@ -0,0 +1,63 @@
...
+
+<Started>
+  June, 2017
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Provides a class for mock link metadata which is information gathered when a
+  step of the supply chain is run as mock.
```
lukpueh: Do you mind adding a little note that summarizes the discussion we had in the PR (why we need a separate MockLink class, alternatives, ...)

in_toto/runlib.py
```diff
@@ -265,66 +266,57 @@ def execute_link(link_cmd_args, record_byproducts):
...
-    "materials" : materials_dict,
-    "products" : products_dict,
-    "command" : link_cmd_args,
-    "byproducts" : byproducts,
-    "return_value" : return_value
-  }
+  log.info("Creating mock link metadata...")
+  mock_link = in_toto.models.mock_link.MockLink(name=name, materials=materials_dict,
+    products=products_dict, command=link_cmd_args, byproducts=byproducts,
+    return_value=return_value, working_directory=os.getcwd())
```
lukpueh: Please add a blank line here

lukpueh (changes requested):
LGTM, please address requested changes (minor) and we can merge.

in_toto/in_toto_mock.py
ShikherVerma: not sure what you mean

ShikherVerma: fixed all other problems

SantiagoTorres:
Thanks! 💯

ShikherVerma:
yaay

Issue #97, opened 2017-06-13
Add `in-toto-sign` command line utility


lukpueh:
in-toto `Layouts` and `Links` are both [signable](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/common.py#L72).

Let's add a command line utility to handle in-toto metadata signatures. Here are some functions I consider helpful:

- add signature, while keeping existing signature(s)
- replace existing signature(s) with new signature
- verify signature(s)

Feel free to add your own ideas.

ShikherVerma:
Maybe it could also create a link from a link-mock file ? would that be a use case?

ShikherVerma:
Can't assign the issue to me. Need discussion more discussion about the CLI syntax before implementing.

lukpueh:
Good thought. I wonder though if it's really necessary. Do you have something in mind where this could be convenient?

ShikherVerma:
Might be useful if you want to do the *link json generation and signing separately*. Like if some super paranoid guy keeps his private gpg key on a disconnected raspberry pi and only signs stuff on it ( @adarshaj i'm looking at you) or uses a [yubikey](https://www.yubico.com/why-yubico/how-yubikey-works/) and doesn't have it while generating the link json.

lukpueh:
That definitely makes sense, but note that you can use `in-toto-run` without a key, which produces an unsigned regular `.link` file, which in turn could be signed using e.g. the here proposed command line tool.

ShikherVerma:
Oh I see. Then this extra `link-mock to link` functionality is not required.

What would be the syntax for in-toto-sign cli.
``` bash
# Add signature
$in-toto-sign --add-sign --key bob --link building.link

# Replace signature
$in-toto-sign --replace-sign --key muck --link step.[bob-key-prefix].link

# for verification
$in-toto-sign --verify --key bob.pub --link bobthebuilder.[bob-key-prefix].link
```
For Add signature and replace signature whose key prefix goes in the link file name ? should we have key prefix if link is singed by more than 1 person?

ShikherVerma:
What if we try to replace sign on a link which is signed by more than one person. then do remove all signatures or we add option for which signature we want to remove ?

isachit:
Shikher are you on it?

isachit:
like already started the programming? If yes, I can send you whatever I have done on this until now, if not let me know too.

ShikherVerma:
@isachit you can take up this issue :+1:

[...]

PR #99, opened 2017-06-14
In-toto-sign

isachit:
Created the CLI tool to handle the following in context of signing `Link` and `Layout` file:
- add signature, while keeping existing signature(s)
- replace existing signature(s) with new signature
- verify signature(s)

**Issue Referenced** - #97 

Please note the progress:
- [x] Write the script
- [x] Add Exceptions
- [x] Add Documentation
- [x] Write the unit tests and run them
- [x] Fix any other issues, and add other functionalities.

@lukpueh and @SantiagoTorres  can you have a look at it, whenever possible and please let me know if it was what you wanted.

UPDATE 1: in-toto-sign now supports multiple signatures(both multiple signing, and verification of the multiple signatures)
UPDATE 2: users can now specify the destination or target filename of the output of in-toto-sign. In case no destination is specified, the source file name would be taken into consideration by in-toto-sign while dumping.


in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+<Purpose>
+  A CLI tool for adding, replacing, verifying signatures in link
+  metadata files. It takes the following inputs - path to the signable
+  object, and path to the key file. Also there are two optional inputs
+  based on which it decides whether to sign the file with the replacement
+  of the existing signature, or sign it but without replacement and then
+  appends the new signature in the file. Further in each of the two cases,
+  depending on the arguments it then dumps the file either by infixing
+  the keyID i.e. <filename>.<8-initial-characters-from-the-keyID>.link
+  or simpy <filename>.link
```
lukpueh: simply

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+            signable_object.dump()
+            sys.exit(0)
+
+          else:
+            signable_object = add_replace(args.signablepath, args.key)
+            signable_object.dump(key = rsa_key)
+            sys.exit(0)
+
+      except Exception as e:
+        print('The following error occured while signing', e)
```
lukpueh: occurred (also fix below)

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    # append the signature into the file
+    signable_object.signatures.append(signature)
+
+    return signable_object
+
+
+def add_replace(link, key):
+    """
+    <Purpose>
+      Replaces all the exisiting signature with the new signature,
```
lukpueh: existing

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+
+    """
+    args = parse_args()
+    rsa_key = in_toto.util.import_rsa_key_from_file(args.key)
+
+    if args.verbose:
+        log.logging.getLogger.setLevel(log.logging.INFO)
+
+    if args.operator == 'sign':
+
```
lukpueh: Please remove blank line

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+      try:
+        if not args.replaceall:
+
```
lukpueh: Please remove blank line

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+            signable_object.dump()
+            sys.exit(0)
+
+          else:
+            signable_object = add_sign(args.signablepath, args.key)
+            signable_object.dump(key = rsa_key)
+            sys.exit(0)
+
+        else:
+
```
lukpueh: Please remove blank line

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+            signable_object = add_replace(args.signablepath, args.key)
+            signable_object.dump(key = rsa_key)
+            sys.exit(0)
+
+      except Exception as e:
+        print('The following error occured while signing', e)
+        sys.exit(2)
+
+    elif args.operator == 'verify':
+
```
lukpueh: Please remove blank line

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    <Exceptions>
+      None
+
+    <Returns>
+      An object containing the contents of the link file after
+      adding the signature which replaces the old signatures
+    """
+
+    # reading the link file and making an object
```
lukpueh: Please start comments with upper case

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    signable_object = link_import.read_from_file(link)
+
+    # import rsa key from the filepath
+    rsa_key = in_toto.util.import_rsa_key_from_file(key)
```
lukpueh: This won't work if the private key is encrypted. You can use `util.prompt_import_rsa_key_from_file` instead. Also below.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # check if the key corresponds to the correct format
+    securesystemslib.formats.KEY_SCHEMA.check_match(rsa_key)
+
+    # core working
```
lukpueh: This is not a very useful comment.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    rsa_key = in_toto.util.import_rsa_key_from_file(key)
+
+    # checking if the given key follows the format
+    securesystemslib.formats.KEY_SCHEMA.check_match(rsa_key)
+
+    # create the signature using the key, and the link file
+    signature = securesystemslib.keys.create_signature(rsa_key, signable_object.payload)
+
+    # append the signature into the file
+    signable_object.signatures.append(signature)
```
lukpueh: Why not use `Signable`'s `sign` method?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    # import rsa key from the filepath
+    rsa_key = in_toto.util.import_rsa_key_from_file(key)
+
+    # check if the key corresponds to the correct format
+    securesystemslib.formats.KEY_SCHEMA.check_match(rsa_key)
+
+    # core working
+    signature = securesystemslib.keys.create_signature(rsa_key, signable_object.payload)
+    signable_object.signatures = []
+    signable_object.signatures.append(signature)
```
lukpueh: Why not use `Signable`'s `sign` method (after having removed existing signatures)?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+      None
+
+    <Exceptions>
+      None
+
+    <Returns>
+      Parsed arguments (args object)
+    """
+    parser = argparse.ArgumentParser(
+        description="in-toto-sign : Signs etc etc")
```
lukpueh: Please make more descriptive.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+      key_pub - public key to be used to verification
+
+    <Exceptions>
+      Raises SignatureVerificationError
+        - 'Invalid Signature' : when the verification fails
+        - 'Signature Key Not Found' : when KeyError occurs
+        - 'No Signatures Found' - when no signature exists
+
+    <Returns>
+      Boolean True when the verification is success
```
lukpueh: This function does not really have to be a boolean function. Either it raises an exception (something went wrong) or it doesn't (the signature was successfully verified/was correct).

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+                    "[--replaceall]\n{0}"
+                    "[--infixkeyid]\n{0}"
+                    "<path/to/signable>"
+                    "[--verbose]\n\n"
+                    .format(lpad))
+
+    in_toto_args = parser.add_argument_group("in-toto-sign options")
+
+    in_toto_args.add_argument("-k", "--key", type=str, required=True,
+                              help="Path to private key to sign link metadata (PEM)")
```
lukpueh: Could also be a public key, right?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    First calls parse_args to parse the arguments, and then calls either
+    add_sign or add_replace depending upon the arguments. Based on the
+    arguments it then dumps the corresponding file.
+
+    """
+    args = parse_args()
+    rsa_key = in_toto.util.import_rsa_key_from_file(args.key)
+
+    if args.verbose:
+        log.logging.getLogger.setLevel(log.logging.INFO)
```
lukpueh: 2 space indentation

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+import in_toto.log
+import in_toto.util
+from in_toto.models.common import Signable as signable_object
+from in_toto.models.link import Link as link_import
+import securesystemslib.exceptions
+import securesystemslib.keys
+import securesystemslib.formats
+
+
+def add_sign(link, key):
```
lukpueh: two space indentation (also below)

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+
+    in_toto_args = parser.add_argument_group("in-toto-sign options")
+
+    in_toto_args.add_argument("-k", "--key", type=str, required=True,
+                              help="Path to private key to sign link metadata (PEM)")
+
+    in_toto_args.add_argument("operator", type=str, choices=['sign', 'verify'],
+                              help="sign or verify")
+
+    in_toto_args.add_argument("-r", "--replaceall", action="store_true",
```
lukpueh: Change to `--replace-sig`

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+                              help = "Whether to replace all the old signatures"
+                              "or not")
+
+    in_toto_args.add_argument("-i", "--infixkeyid", action="store_true",
```
lukpueh: Change to `--infix`

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    in_toto_args = parser.add_argument_group("in-toto-sign options")
+
+    in_toto_args.add_argument("-k", "--key", type=str, required=True,
+                              help="Path to private key to sign link metadata (PEM)")
+
+    in_toto_args.add_argument("operator", type=str, choices=['sign', 'verify'],
+                              help="sign or verify")
+
+    in_toto_args.add_argument("-r", "--replaceall", action="store_true",
+                              help = "Whether to replace all the old signatures"
```
lukpueh: Please remove whitespaces before and after assignment operator when using keyword arguments consistently throughout the script.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+                              "or not")
+
+    in_toto_args.add_argument("-i", "--infixkeyid", action="store_true",
+                              help = "whether to infix keyID in the file name"
+                              "or not")
```
lukpueh: Please add a little more detail.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    lpad = (len(parser.prog) + 1) * " "
+
+    parser.usage = ("\n"
+                    "--key <signing key> | <verifying key>\n{0}"
+                    "<sign | verify>\n{0}"
+                    "[--replaceall]\n{0}"
+                    "[--infixkeyid]\n{0}"
+                    "<path/to/signable>"
+                    "[--verbose]\n\n"
+                    .format(lpad))
```
lukpueh: Please review formatting here. Should look more like:

```shell
usage:
in_toto_sign.py (sign | verify) --key <filepath>
                [--replace-sig]
                [--infix]
                [--verbose]
                <path/to/signable>
```

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+
+    in_toto_args = parser.add_argument_group("in-toto-sign options")
+
+    in_toto_args.add_argument("-k", "--key", type=str, required=True,
+                              help="Path to private key to sign link metadata (PEM)")
+
+    in_toto_args.add_argument("operator", type=str, choices=['sign', 'verify'],
+                              help="sign or verify")
```
lukpueh: If you are really motivated and want to learn `argparse` you could use [`subcommands`](https://docs.python.org/2.7/library/argparse.html#sub-commands) for this.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+
+          if not args.infixkeyid:
+            signable_object = add_replace(args.signablepath, args.key)
+            signable_object.dump()
+            sys.exit(0)
+
+          else:
+            signable_object = add_replace(args.signablepath, args.key)
+            signable_object.dump(key = rsa_key)
+            sys.exit(0)
```
lukpueh: You could simplify this like so (pseudocode)
```python
try:
  if sign:
    if replace:
       signed_obj = sign_replace(...)

    else:
        signed_obj = sign_add(...)

    if infix:
        signed_obj.dump_with_infix()

    else:
        signed_obj.dump_without_infix()

  else:
     ...
except ExceptionA:
  ...
except ExceptionB:
  ...
...
```


You see what I mean?

lukpueh (changes requested):
Thanks for the PR. Works as expected. Most of my comments are actually style considerations. 

Please also add the `console_script` to `setup.py` and  write unit tests. :)

in_toto/in_toto_sign.py
SantiagoTorres: I think there is also merit in trying to simplify the description too.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+    # create the signature using the key, and the link file
+    signature = securesystemslib.keys.create_signature(rsa_key, signable_object.payload)
+
+    # append the signature into the file
+    signable_object.signatures.append(signature)
+
+    return signable_object
+
+
+def add_replace(link, key):
```
SantiagoTorres: I'm not sure if this is the best method name. To me it reads as "adding a replace" rather than "replacing any existing signatures"

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+        print('The following error occured while signing', e)
+        sys.exit(2)
+
+    elif args.operator == 'verify':
+
+      try:
+        if verify_sign(args.signablepath, args.key):
+          sys.exit(0)
+        else:
+          sys.exit(1)
```
SantiagoTorres: This will never hit. You either return a `True` or raise an `Exception`, so the condition can never be false.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,265 @@
...
+
+      except Exception as e:
+        print('The following error occured while verification', e)
+        sys.exit(3)
```
SantiagoTorres: I"m not sure we are documenting these error codes are we? Is it worth having it something different on each case?

SantiagoTorres (commented):
Looking good! Thanks for the contrib 👍

in_toto/in_toto_sign.py
isachit: Okay using it.

in_toto/in_toto_sign.py
isachit: Alright! Since I am using link.verify_signatures(key_dict), in the function verify_signatures, exceptions are already raised. So what I can do is that, either it returns nothing, the verification was successful, or it catches the exception which occurred while running and prints it . What do you say?

in_toto/in_toto_sign.py
isachit: Yes. Will change the description.

in_toto/in_toto_sign.py
isachit: Okay! Yes yes! Changed it!

in_toto/in_toto_sign.py
isachit: replace_sig? What do you say?

in_toto/in_toto_sign.py
isachit: changed it.

in_toto/in_toto_sign.py
SantiagoTorres: sounds good 👍

in_toto/in_toto_sign.py
isachit: alright. changed it.

test/test_in_toto_sign.py
```diff
@@ -0,0 +1,190 @@
...
+    demo_files = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "demo_files")
+
+    # Create and change into temporary directory
+    self.test_dir = os.path.realpath(tempfile.mkdtemp())
+    os.chdir(self.test_dir)
+
+    # Copy demo files to temp dir
+    for file in os.listdir(demo_files):
+      shutil.copy(os.path.join(demo_files, file), self.test_dir)
```
lukpueh: `file` is reserved in python. Please use a different name.

test/test_in_toto_sign.py
```diff
@@ -0,0 +1,190 @@
...
+
+    # Load layout template
+    layout_template = Layout.read_from_file("demo.layout.template")
+
+    # Store layout paths to be used in tests
+    self.layout_single_signed_path = "single-signed.layout"
+    self.layout_double_signed_path = "double-signed.layout"
```
lukpueh: You don't need this. It's never used

test/test_in_toto_sign.py
```diff
@@ -0,0 +1,190 @@
...
+  def test_in_toto_sign_add_sign(self):
+    """in_toto_sign_add_sign run through. """
+    add_sign(self.layout_single_signed_path, self.alice_path_pvt)
+
+  def test_in_toto_sign_replace_sign(self):
+    """in_toto_sign_replace_sign run through. """
+    replace_sign(self.layout_single_signed_path, self.alice_path_pvt)
+
+  def test_in_toto_sign_verify_sign(self):
+    """in_toto_sign_verify_sign run through. """
```
lukpueh: This actually does not "run through", right?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,239 @@
...
+
+      else:
+        signable_object.dump()
+        sys.exit(0)
+
+    else:
+      verify_sign(args.signablepath, args.key)
+
+  except Exception as e:
+    print('The following error occurred', e)
```
lukpueh: Could you use `log.error` here instead?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,239 @@
...
+      if args.infix:
+        signable_object.dump(key=rsa_key)
+        sys.exit(0)
+
+      else:
+        signable_object.dump()
+        sys.exit(0)
+
+    else:
+      verify_sign(args.signablepath, args.key)
```
lukpueh: Could you put this in a separate `try/except` block (not inside this try) and log with `log.fail_verification` on `SignatureVerificationError`, and `log.error` on all other exceptions?

Maybe it would also be nice to log with `log.pass_verification` if the verification passes?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,239 @@
...
+                  "<path/to/signable>\n\n"
+                  .format(lpad))
+
+  in_toto_args = parser.add_argument_group("in-toto-sign options")
+
+  in_toto_args.add_argument("operator", type=str, choices=['sign', 'verify'],
+                            help="Sign or Verify")
+
+  in_toto_args.add_argument("-k", "--key", type=str, required=True,
+                            help="Path to the key ")
```
lukpueh: Do you mind changing this to accept multiple public keys? Currently, `verify` would fail when the file is signed by multiple keys. (Hint: [`nargs='+'`](https://docs.python.org/2.7/library/argparse.html#nargs)) And don't forget to change the usage message. ;)

lukpueh (changes requested):
Almost there.

test/test_in_toto_sign.py
isachit: I guess file_object should be okay.

lukpueh: IMHO something like "file_path" is better, because `listdir` does not return file objects but paths. ;)

isachit: Okay!

in_toto/in_toto_sign.py
isachit: Sure. But, while verifying, in case of multiple signatures, the public keys need to be input in the same order as it was signed by the functionaries. That is, for example, the file was first signed by A, then by B (without replacement), then by C (without replacement). When verification takes place, then, pub key of A, followed by B, followed by C, should be specified in the arguments. Right? Or? 

Also there will be multiple cases, depending upon if any one of the functionaries in the supply chain have replaced all the existing signatures or not. What do you say?

test/test_in_toto_sign.py
isachit: Done.

test/test_in_toto_sign.py
isachit: Done.

in_toto/in_toto_sign.py
isachit: Done.

in_toto/in_toto_sign.py
isachit: Done.

in_toto/in_toto_sign.py
isachit: Ah okay. Seems like for verification, the function I have used, takes care of that. It takes input a dictionary of filepaths to various public keys and verifies them. Okay cool.

isachit: screen shot 2017-06-29 at 6 08 57 pm
Any idea why this happens? hm.. Strange..

SantiagoTorres: it's taking the three strings as part of --keys. An easy way out would be to change the parameters to be
```bash
in_toto_sign.py verify <link> --keys key1 [[key2]...] 
```
Or something a litte more git-like
```bash
in_toto_sign verify key1 [[key2]...] -- <path-to-link-file>
```

But I don't know how hard would it be to have this on argparse. Maybe the former is easier.

in_toto/in_toto_sign.py
isachit: Well, as on here, [common.py](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/common.py) ,It raises SignatureVerificationError on all the cases, no matter if it is due to KeyError, or due to invalid signature or something else.

in_toto/in_toto_sign.py
isachit: Alright! Thank you so much!

in_toto/in_toto_sign.py
lukpueh: Yes, `Signable.verify_signatures` raises `SignatureVerificaitonError` whenever something goes wrong.

But your `verify_sign` does not only call `verify_signatures`, it also [loads keys and the file to be verified](https://github.com/isachit/in-toto/blob/c71ccd5152211cb4decab0a25d40371457e977e5/in_toto/in_toto_sign.py#L137-L139). If something goes wrong while doing that I think your tool should `log.error`.

isachit: Ah okay. Cool. :) Added it.

in_toto/in_toto_sign.py
isachit: Hey @lukpueh , on another note, when asking the user for the target, do he needs to specify a filename? or a path along with the filename to which the file should be dumped?

isachit: For example:

```--destination package_lukas``` 

OR

``` abc/def/lukas/files/important/package_lukas```

**Comment by isachit** [comment]
Why is coverage decreasing. :|

File: in_toto/in_toto_sign.py [hunk 34]
```diff
@@ -134,10 +137,28 @@ def verify_sign(link, key_pub):
...
+          signable_object = layout_import.read_from_file(file_path)
+          return signable_object
+
+        elif 'Link' in line:
+          found = 1
+          signable_object = link_import.read_from_file(file_path)
+          return signable_object
+    if found==-1:
+      raise Exception("Invalid file")
+  file_object.close()
```
lukpueh: We talked about this. Here is what you should do:
Read the file from disk using `open`, try to parse the JSON into a Python dictionary, check if that dict has a key called `_type`, and instantiate the object accordingly using `Layout.read(python_dict)` or `Link.read(python_dict)`.

lukpueh: Btw. you don't have to close a file explicitly when opened in a `with` block. From the [python docs](https://docs.python.org/2.7/tutorial/inputoutput.html#methods-of-file-objects):

```python
>>> with open('workfile', 'r') as f:
...     read_data = f.read()
>>> f.closed
True
```

in_toto/in_toto_sign.py
```diff
@@ -207,33 +234,87 @@ def main():
   arguments it then dumps the corresponding file.
   """
   args = parse_args()
-  rsa_key = in_toto.util.import_rsa_key_from_file(args.key)
+  length = len(args.keys)
+  rsa_key = in_toto.util.import_rsa_key_from_file(args.keys[length-1])
+  FILENAME_FORMAT_LAYOUT = "{file_name}.{keyid:.8}.layout"
+  FILENAME_FORMAT_LINK = "{file_name}.{keyid:.8}.link"
+  index=args.signablepath.rfind('/')
+  source_file_name=args.signablepath[index+1:]
```
lukpueh: You could use [`os.path.basename`](https://docs.python.org/2/library/os.path.html#os.path.basename) instead.

in_toto/in_toto_sign.py
```diff
@@ -207,33 +234,87 @@ def main():
   arguments it then dumps the corresponding file.
   """
   args = parse_args()
-  rsa_key = in_toto.util.import_rsa_key_from_file(args.key)
+  length = len(args.keys)
+  rsa_key = in_toto.util.import_rsa_key_from_file(args.keys[length-1])
+  FILENAME_FORMAT_LAYOUT = "{file_name}.{keyid:.8}.layout"
```
lukpueh: I am not sure if we ever need a keyid-infixed layout. But you can leave it.

in_toto/in_toto_sign.py
```diff
@@ -207,33 +234,87 @@ def main():
...
+  FILENAME_FORMAT_LINK = "{file_name}.{keyid:.8}.link"
```
lukpueh: You should import and use the existing  [`FILENAME_FORMAT`](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/link.py#L26) instead.

in_toto/in_toto_sign.py
```diff
@@ -207,33 +234,87 @@ def main():
   arguments it then dumps the corresponding file.
   """
   args = parse_args()
-  rsa_key = in_toto.util.import_rsa_key_from_file(args.key)
+  length = len(args.keys)
+  rsa_key = in_toto.util.import_rsa_key_from_file(args.keys[length-1])
```
lukpueh: This comes a little unexpected. I wonder if it wouldn't be better to separate `sign` and `verify` after all, e.g. using a subparser as suggested earlier.

in_toto/in_toto_sign.py
lukpueh: Also, try to stick to the codestyle guidelines (whitespace before and after assignment operator).

in_toto/in_toto_sign.py
```diff
@@ -207,33 +234,87 @@ def main():
...
+        if not args.destination:
+          if signable_object._type=='layout':
+            fn = FILENAME_FORMAT_LAYOUT.format(file_name=source_file_name,
+              keyid=rsa_key['keyid'])
+            signable_object.dump(fn)
+            sys.exit(0)
+
+          else:
+            signable_object.dump(key=rsa_key)
+            sys.exit(0)
```
lukpueh: Didn't we agree that if no `destination` is specified, you will `dump` overwrite the file? That's not what's happening here.

lukpueh (changes requested):
The command line arguments here are getting a little confusing. I think we should re-discuss the exact usage of this script. Let me write a follow-up comment that explains what I have in mind before you address any of the comments I wrote in this review.

As a general note, please make sure that unittest coverage does not decrease with your code additions. If you take a look at [`coveralls`' report](https://coveralls.io/builds/12234311/source?filename=in_toto%2Fin_toto_sign.py) you'll see that there are a bunch of lines with red background. These lines are not executed by any unit test, so you would have to add unit tests that trigger that behavior.

in_toto/in_toto_sign.py
vladimir-v-diaz: > As a general note, please make sure that unittest coverage does not decrease with your code additions.

Have you tried using coverage's `--fail-under`?  This should make it so that if coverage falls below a certain threshold, a build failure occurs.
```
coverage report -m --fail-under 97
```

I wonder if you can give this option a decimal number...

in_toto/in_toto_sign.py
isachit: Hm. But isn't it just for the link files? I was using it, but then I thought, to make it different for link and layout.

in_toto/in_toto_sign.py
isachit: @lukpueh Sure. But, what if he specifies, that the keyID must be infixed. True, ```signable_object.dump(key=rsa_key) ``` does not overwrites the file because
1. It takes into account when keyID must be infixed in the file name.
2. Even when it infixes the keyID in the original source file name, it doesn't dump overwrites it since the new file is dumped at a different location than the original source file.

So I guess, the only change that needs to be done, is to dump it at the original location, and overwrite the existing file when `-i` is not specified, and in the other case, remove the existing file, and dump the one with the infixed keyID in it.  Is it what you mean?

isachit: Sure. I will add the tests for those. @vladimir-v-diaz  Thanks!

in_toto/in_toto_sign.py
isachit: Done.

lukpueh:
Let's take a step back and consider what we actually want here. `in-toto-sign` has two subcommands, `sign` and `verify`. Let's look at each of their options/arguments individually (omitting `-v` because that's straight forward).

**sign**
- `<path/to/input/signable>`
   Path to signable file, must be a link or layout in json format
- `--keys <private key path>[ <private key path>, ...]`
  Create a signature using each key and add each signature to the signable
- `[--replace-sig]` (optional)
  If specified, remove existing signatures before adding new signatures
- `[--destination] <path/to/output/signable>` (optional)
  If specified dump signed link or layout to `<path/to/output/signable>` else dump to `<path/to/input/signable>`
- `[--infix]` (optional)
  Here we have to make some decisions. First, which keyid should we use as infix if the resulting file will have multiple signatures, either because a signature was added to an exiting signature or multiple signatures were added? And second, where should we dump the file to? Let's keep in mind that the keyid infix actually only makes sense for link files signed by one key. I propose the following solutions:
  - Make `--destination` and `--infix` mutually exclusive (fail if both are specified)
  - Ignore `--infix` if the passed signable is a layout (warn user using `log.warn`)
  - Use the key of the last signature that was added as infix (warn user using `log.warn`)
  - Dump the resulting signable to <link_name>.<first 8 digits of keyid>.link in the current working directory (inform user using `log.info`)
  
**verify** (should be straight forward)
- `<path/to/signable>`
- `--keys <public key path>[ <public key path>, ...]`

Let me know would you think about it (especially the `--infix` part)!

isachit:
@lukpueh  Please check afb69df77721fdbe0a7e76447010c670bc400673 
I have implemented it according to your last comment and as per our discussion.

isachit:
@lukpueh  Do you know why the travis CI build isn't happening?

lukpueh:
Yes, because there are conflicts in `setup.py`, because we have merged your other PR that has also modified that file.

isachit:
Okay. Well that's nuff. Still -0.2% :| Honestly @coveralls ? Are you kidding me?

isachit:
@lukpueh  Sorry to bug you again, but you need to see this.

vladimir-v-diaz:
@isachit 
You don't appear to have a test case for line 114 of common.py:
https://coveralls.io/builds/12299985/source?filename=in_toto%2Fmodels%2Fcommon.py

isachit:
@vladimir-v-diaz  Thanks maaan! :D Trying to write a test for it.

isachit:
Yaaye!! :D

isachit:
@lukpueh  I guess its done. Just history rewriting remains. Right?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,326 @@
...
+  <Returns>
+    None
+  """
+  signable_object = check_file_type_and_return_object(file_path)
+  key_pub_dict = in_toto.util.import_rsa_public_keys_from_files_as_dict(
+      key_pub)
+  signable_object.verify_signatures(key_pub_dict)
+
+
+def check_file_type_and_return_object(file_path):
```
lukpueh: Can you write a short docstring here?

lukpueh, line 148: Can you write a short docstring here?

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,326 @@
...
+  lpad = (len(parser.prog) + 1) * " "
+
+  parser.usage = ("\n"
+                  "(sign | verify) <path/to/signable>\n{0}"
+                  "[--replace-sig]\n{0}"
+                  "[--infix]\n{0}"
+                  "[--verbose]\n{0}"
+                  "[--destination]\n{0}"
+                  "--keys <filepath/filepaths>\n\n"
+                  .format(lpad))
```
lukpueh: Let's just remove the custom usage message (Line 185-195). The default usage message that argparse provides is good enough and more accurate.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,326 @@
...
+  if args.verbose:
+    log.logging.getLogger().setLevel(log.logging.INFO)
+
+  if args.subparser_name == 'sign':
+    try:
+      if args.destination and args.infix:
+        log.error('Please specify one and only one out of infix and '
+                  'destination!')
+        sys.exit(1)
+      else:
```
lukpueh: You only get to this line if above condition does not match. So you can remove the else here, which makes the code less nested.

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,326 @@
...
+        if args.replace_sig:
+          signable_object = replace_sign(args.signablepath, args.keys)
+        else:
+          signable_object = add_sign(args.signablepath, args.keys)
+
+        if signable_object._type == 'layout':
```
lukpueh: `os.path.join` only really makes sense if you pass it two arguments, also you should only warn if `--infix` is specified and in general you could make this a little slicker. What do you think about the following?

```python
if args.infix: 
  log.warn('Ignoring option --infix for layouts...')

if args.destination:
  path = args.destination

else:
  path = args.signablepath

log.info("Dumping layout to '{}' ...".format(path))
signable_object.dump(path)
sys.exit(0)
```

in_toto/in_toto_sign.py
```diff
@@ -0,0 +1,326 @@
...
+            log.info('Dumping in the current working directory...')
+            signable_object.dump(source_file_name)
+            sys.exit(0)
+          else:
+            file_name = os.path.join(args.destination)
+            signable_object.dump(file_name)
+            log.info('Dumping at ' + str(file_name) + '...')
+            sys.exit(0)
+
+        if signable_object._type == 'Link':
```
lukpueh: Here another suggestion for the link dumping code.

```python
if args.destination:
  path = args.destination

elif args.infix:
  # We will dump the file name to CWD as
  # `<link name>.<keyid of last key passed>.link`
  path = FILENAME_FORMAT_IMPORT.format(
    step_name=signable_object.name, keyid=args.keys[-1]["keyid"])

  if len(args.keys) > 1:
    log.warn("Using last key in the list of passed keys for infix...")

else:
  path = args.signablepath

log.info("Dumping link to '{}'' ...".format(path))
signable_object.dump(path)
sys.exit(0)
```

lukpueh (changes requested):
Awesome coverage! Overall, looking good. Here a couple of final minor suggestions.

in_toto/in_toto_sign.py
lukpueh: I guess we could even take it a step further and write the last three lines only once for layouts and links. :)
e.g.:
```python
log.info("Dumping {0} to '{1}'' ...".format(signable_object._type.lower(), path))
signable_object.dump(path)
sys.exit(0)
```

in_toto/in_toto_sign.py
isachit: Done.

in_toto/in_toto_sign.py
isachit: Ahh okay.

isachit:
+0.1% Yaaayyee!!

lukpueh:
Thanks @isachit!

PR #100, opened 2017-06-19
Remove create_link_metadata

ShikherVerma:
* **What does this PR do?**
Removes `create_link_metadata` function as it adds more lines of code than removing.

@lukpueh

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -265,67 +265,6 @@ def execute_link(link_cmd_args, record_byproducts):
 
   return {"stdout": stdout_str, "stderr": stderr_str}, return_value
 
-def create_link_metadata(link_name, materials_dict=None, products_dict=None,
-    link_cmd_args=None, byproducts=None, return_value=None):
-  """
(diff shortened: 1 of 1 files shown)


ShikherVerma:
@lukpueh the decrease in coverage is due to decrease in loc.

in_toto/runlib.py
```diff
@@ -389,8 +328,15 @@ def in_toto_run(name, material_list, product_list,
...
-      link_cmd_args, byproducts, return_value)
+  link_dict = {
+    "name" : name,
+    "materials" : materials_dict,
+    "products" : products_dict,
+    "command" : link_cmd_args,
+    "byproducts" : byproducts,
+    "return_value" : return_value
+  }
+  link = in_toto.models.link.Link.read(link_dict)
```
SantiagoTorres, line 339: wouldn't it be more convenient to just use the constructor with all the parameters from above here?

lukpueh: Yep, that's what I was implying earlier.

Issue #96, opened 2017-06-13
Add `in-toto-keygen` utility


lukpueh:
Currently in-toto provides some [utility functions](https://github.com/in-toto/in-toto/blob/develop/in_toto/util.py) to create (password encrypted) RSA key pairs programatically.

It would be nice to additionally have a command line utility, e.g. `in-toto-keygen`, for key creation (like a stripped down `ssh-keygen`).

Keep in mind that we might want to [drop the util module](https://github.com/in-toto/in-toto/issues/80) at some point and move its (updated) functionality to [securesystemslib/keys](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/keys.py).

isachit:
On it.

lukpueh:
Implemented with https://github.com/in-toto/in-toto/pull/102

PR #102, opened 2017-06-29
in-toto-keygen: a command line utility for generating keys

isachit:
A CLI tool for creating key files, and dumping them with filename , filename.pub for private and public keys respectively.
General Usage:
`python in_toto_keygen.py [-p] <filename>`

**Issue Referenced** - #96 

Please note the progress:
- [x] Write the script
- [x] Add Exceptions
- [x] Add Documentation
- [x] Write the unit tests and run them
- [ ] Fix any other issues, and add other functionalities.

@lukpueh and @SantiagoTorres.

UPDATE 1: Added @lukpueh in the author field since he is the one who actually wrote the two functions I have used.
UPDATE 2: It's pretty much in place. Just removing the unnecessary history remains.


isachit:
I wonder, why aren't @SantiagoTorres  and @lukpueh  in the participants section. :| :/ And also, The travis CI build, along with coveralls. Hm.. Fishy.

isachit:
Oh yeeesss!!!! Finally!! :D

test/test_in_toto_keygen.py
```diff
@@ -0,0 +1,99 @@
...
+
+  def test_prompt_password(self):
+      """Call password prompt. """
+      password = "123456"
+      with patch("getpass.getpass", return_value=password):
+          self.assertEqual(prompt_password(), password)
+
+
+if __name__ == '__main__':
+  unittest.main(buffer=True)
```
SantiagoTorres: Can you end this line with a newline? This usually confuses linecount algorithms and causes off-by-ones on diffing

test/test_in_toto_keygen.py
```diff
@@ -0,0 +1,99 @@
+"""
```
SantiagoTorres, line 1: In addition to the existing tests, I'm wondering if it'd be nice to have:

1. Create a keypair, and load the public and private keys
2. Create a keypair with different length parameters and load it

SantiagoTorres (commented):
Looking good! 👍

test/test_in_toto_keygen.py
isachit: Alright. On it. :)

test/test_in_toto_keygen.py
isachit: Okay!

isachit:
@SantiagoTorres  I guess, this is what you suggested right? or?

in_toto/in_toto_keygen.py
```diff
@@ -14,7 +14,9 @@
 
 <Purpose>
   A CLI tool for creating key files, and dumping them with <filename>,
-  <filename>.pub for private and public keys respectively.
+  <filename>.pub for private and public keys respectively. It also takes an
+  integer as an input, which specifies the length of the RSA key to be
+  generated. By default it is set as 3072.
```
SantiagoTorres: Yeah this is good.

SantiagoTorres (commented):
Looks good to me. @lukpueh what do you think?

in_toto/in_toto_keygen.py
```diff
@@ -0,0 +1,170 @@
...
+  try:
+    if args.prompt:
+      prompt_generate_and_write_rsa_keypair(args.name, args.bits)
+      sys.exit(0)
+    else:
+      generate_and_write_rsa_keypair(args.name, args.bits, password=None)
+      sys.exit(0)
+
+  except Exception as e:
+    print('The following error occurred', e)
```
lukpueh: Please use `log.error` instead of print.

in_toto/in_toto_keygen.py
```diff
@@ -0,0 +1,170 @@
...
+                            "private key with the same before storing")
+
+  in_toto_args.add_argument("name", type=str,
+                            help="The filename of the resulting key files",
+                            metavar="<filename>")
+
+  in_toto_args.add_argument("bits", nargs= "?", default=3072, type=int,
+                            help="The key size, or key length, of the RSA "
+                            "key.  'bits' must be 2048, or greater, and a "
+                            "multiple of 256.", metavar="<bits>")
```
lukpueh: Latter is only [partially true](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/keys.py#L211-L214) also see the [default crypto library](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/settings.py#L49).
I think you can just drop the info here.

lukpueh (changes requested):
Great work!

 I wonder though why you copied over the code from `in_toto.util`. I thought we had agreed that you will try to use `securesystemslib.keys` directly, or if that's not possible, just use `in_toto.util`.

I'd rather not have the same code twice, especially since I'd like to move that particular code to `securesystemslib.keys` altogether. 

👍  for adding the key length parameter. 
I'd welcome if you changed `in_toto.util. generate_and_write_rsa_keypair` to accept an optional bits parameter and then use that function here.

lukpueh:
Great work! Thanks for the PR, @isachit. Out of curiosity, why did you call the branch *nonuke*?

isachit:
@SantiagoTorres I see what you did there! :D :P Haha. There were some issues with rebasing and removing the commit history not relevant to this PR. So @SantiagoTorres  helped me with that, and while doing it, he made the branch nonuke.

SantiagoTorres:
[¯\\_(ツ)_/¯](http://i.imgur.com/7A2QTxW.gif)

PR #107, opened 2017-07-18
Refactoring artifact rules field name

isachit:
The names of the artifact rule properties of in-toto steps and inspections have been updated in the implementation. They are now called ```expected_materials``` and ```expected_products```.

Issue Referenced: #106

### DIFF
--- a/demo/owner_alice/create_layout.py
+++ b/demo/owner_alice/create_layout.py
@@ -17,25 +17,27 @@ def main():
     },
     "steps": [{
         "name": "clone",
-        "material_matchrules": [],
-        "product_matchrules": [["CREATE", "demo-project/foo.py"]],
+        "expected_materials": [],
+        "expected_products": [["CREATE", "demo-project/foo.py"]],
         "pubkeys": [key_bob["keyid"]],
         "expected_command": "git clone https://github.com/in-toto/demo-project.git",
         "threshold": 1,
       },{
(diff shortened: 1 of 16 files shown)


lukpueh:
@isachit, thanks for the PR I re-worded your commit messages and added another commit. @SantiagoTorres, do you want to review? I guess we can squash before merging.

demo/owner_alice/create_layout.py
```diff
@@ -53,9 +55,9 @@ def main():
             ["ALLOW", "alice.pub"],
             ["ALLOW", "root.layout"],
         ],
-        "product_matchrules": [
+        "expected_products": [
             ["MATCH", "demo-project/foo.py", "WITH", "PRODUCTS", "FROM", "update-version"],
-            # FIXME: See material_matchrules above
+            # FIXME: See expected_materials above
```
SantiagoTorres, line 59: Minor: is this `FIXME` and the one above it still relevant?

SantiagoTorres:
Feel free to merge regardless of the comment. This has my LGTM.

demo/owner_alice/create_layout.py
lukpueh: Actually this one is still relevant, because it refers to the `FIXME` in line 51, which we have not fixed yet. But the one in 29 has been fixed. :)

PR #110, opened 2017-09-01
Add header blocks to `setup.py` and `runtests.py`

lukpueh:


setup.py
```diff
@@ -1,6 +1,33 @@
...
+    - Python2.7 (www.python.org)
+    - OpenSSL (www.openssl.org)
+    - git (git-scm.com)
+    - pip (pip.pypa.io)
+    - virtualenvs - optional but strongly recommended!
+      (http://docs.python-guide.org/en/latest/dev/virtualenvs/)
+
+  # Installation from GitHub
+    ```
+    pip install git+git://github.com/in-toto/in-toto@develop
```
SantiagoTorres: interesting that `git+git` exists as a protocol. I think I'd rather opt for `git+https`.

setup.py
```diff
@@ -1,6 +1,33 @@
...
+<Started>
+  May 23, 2016
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  setup.py script to install in-toto framework and in-toto scripts
+
+  # System Dependencies
```
SantiagoTorres, line 19: Our *system* doesn't depend on git, or pip per se. I think this is from the readme, but we could probably call these "development requirements" rather than "system dependencies"? What do you think?

setup.py
lukpueh: Will change

setup.py
lukpueh: Makes sense to not list those under "system dependencies". What about the oxymoronic "optional requirements"? Because at least pip is not really a "development requirement" either.

setup.py
SantiagoTorres: Just a note, I wasn't being pedantic, apparently you can do git+git :P

setup.py
SantiagoTorres: I think probably going with debian's `recommended tools` for pip/git, and the others as actual `requirements`. What do you think?

setup.py
lukpueh: I think https is a good choice. 
Btw. here's a good [overview of `git`'s protocols](https://git-scm.com/book/id/v2/Git-on-the-Server-The-Protocols) which are [all supported by `pip`](https://pip.pypa.io/en/stable/reference/pip_install/#git).

setup.py
lukpueh: Sounds good to me, I'd vote for adding `virtualenvs` to recommended tools as well

SantiagoTorres: sounds good 👍

lukpueh:
Thanks for reviewing, @SantiagoTorres!

Issue #87, opened 2017-05-02
Clean up unit-test code.


SantiagoTorres:
Some of the unit test code needs to be refactored and cleaned up.

This issue is left up ambiguous intentionally, as it will be discussed in an upcoming hackathon.

lukpueh:
From a note in a23e3570374339433af04dd168b5b06feff9f2ba:
>[the test keys'] keyids, as they appear e.g. in the link filenames [write-code.c8650e01.link and package.2dc02526.link] and in "demo.layout.template", are actually outdated. I.e. if these keys ("bob" and "carl") are imported from PEM format into securesystemlib's rsa key format and hashed to generate the keyid, that keyid would be different.
The reason is that securesystemlib's key format has changed to to include a field "keyid_hash_algorithms" when creating the hash/keyid.

This should be taken care of when fixing this issue

gvjkumar2004:
I would like to work on this issue !

adityasaky:
Go ahead!

ashutosh887:
@adityasaky Let me work on this now

Ayush9026:
@adityasaky sir i want to solve this issue.

lukpueh:
> @adityasaky sir i want to solve this issue.

Thanks, @Ayush9026, contributions are very much appreciated! Do you have any specific refactorings in mind? From what I see this issue requires some discussion about the scope first.

tannerjones4075:
Is this issue still relevant or should it be closed?

lukpueh:
The scope of this issue is quite unclear and probably too broad. Let's close and resubmit, if we have something more specific in mind.

PR #111, opened 2017-09-01
Fix wrong arguments in in_toto_record tests

lukpueh:
The functions `in_toto_record_start` and `in_toto_record_stop` expect a list of file paths/ directory paths as third argument. This commit fixes 3 unittests that passed strings instead of lists.

While this modifies those specific unittests to behave as expected, two important problems remain:

- Unittests should focus on what they are testing, i.e. if a unittest contains logic that can pass or fail, maybe we shouldn't include that logic (fix with https://github.com/in-toto/in-toto/issues/87)
- We should employ more defensive programming, i.e. validate arguments to functions (do we want a "make-code-more-defensive" ticket?)

### DIFF
--- a/test/test_in_toto_record.py
+++ b/test/test_in_toto_record.py
@@ -126,18 +126,18 @@ class TestInTotoRecordTool(unittest.TestCase):
 
   def test_in_toto_record_start_stop(self):
     """in_toto_record_start/stop run through. """
-    in_toto_record_start("test-step", self.key, self.test_artifact)
-    in_toto_record_stop("test-step", self.key, self.test_artifact)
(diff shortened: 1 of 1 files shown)


SantiagoTorres:
>  Unittests should focus on what they are testing, i.e. if a unittest contains logic that can pass or fail, maybe we shouldn't include that logic (fix with #87)

True. I don't recall the testing codebase having a lot of this, does it? 

>  We should employ more defensive programming, i.e. validate arguments to functions (do we want a "make-code-more-defensive" ticket?)

Hmm, I think this philosophy should be in the code guidelines for the lab. Sure, let's open a ticket for this...

PR #112, opened 2017-09-01
Add readme Layout field (fix #108)

### LINKED ISSUE #108: Layout model is missing the README field.
As of the [0.9 version of the in-toto specification](https://github.com/in-toto/docs/blob/v0.9/in-toto-spec.pdf), the Layout metadata should include an optional README field.

We need to update the Layout model to reflect this change.

lukpueh:
Add `readme` field to in-toto Layout model (fix in-toto/in-toto#108)

With version 0.9 of the in-toto specification, the Layout metadata has a `readme` field for an optional human-readable description of the supply chain.

Note that while this change allows to load existing layouts (without readme field, defaulting to an empty string), it will break any existing signatures.

The PR also adds a data type validator (must be string) and a unittest for that validator.

### DIFF
--- a/test/models/test_layout.py
+++ b/test/models/test_layout.py
@@ -126,7 +126,6 @@ class TestLayoutValidator(unittest.TestCase):
       self.layout.steps = [test_step]
       self.layout.validate()
 
-
     test_step = Step(name="this-is-a-step")
     test_step.expected_materials = [["CREATE", "foo"]]
     test_step.threshold = 1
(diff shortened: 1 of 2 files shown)


test/models/test_layout.py
```diff
@@ -180,5 +189,7 @@ def test_import_step_metadata_wrong_type(self):
     # Clean up
     os.remove(link_path)
 
+
+
```
SantiagoTorres: I think we overshooted this whitespace. Other than that, it LGTM.

SantiagoTorres:
LGTM

PR #119, opened 2017-09-20
Fix test keys

lukpueh:
See detailed commit messages for more information.
Note: We should think about regression tests at some point.


SantiagoTorres:
>  Note: We should think about regression tests at some point.

👍

SantiagoTorres (approved):
Good commit descriptions and conceptual separation. Thanks!

PR #120, opened 2017-09-21
Fix metadata format

### LINKED ISSUE #98: Fix metadata signatures mismatch between specs and implementation
We have been silently accepting a minor metadata format mismatch between specs and implementation. I think it is time to fix that by choosing one or the other.

Reason for the format as specified: Adopted from TUF
Reason for the format as implemented: `is-a` relationship seemed  (from an OO perspective)  more intuitive and easier to handle than `has-a` realtionship.

**Implementation**
```python
{
  "_type": "layout" or "link",
  ... # rest of the respective properties,
  "signatures" : [
    { "keyid" : KEYID, "method" : METHOD, "sig" : SIGNATURE },
  , ... ]
}
```

**Specification**

```python
{
  "signed" : ROLE, # layout or link metadata object [**]
  "signatures" : [
    { "keyid" : KEYID, "method" : METHOD, "sig" : SIGNATURE },
  , ... ]
}

# [**] Btw. I find the term `ROLE` for the metadata type really confusing. 
# While it makes sense for TUF, # where each role corresponds to 
# an individual metadata format, it does not for in-toto, which does 
# not have the some notion of roles.
```

SantiagoTorres:
The current state of the in-toto classes do not match with the description in the specification (closes #98) 

Please review commit by commit, so as to make historical sense **and** to simplify your review process.

You will probably notice some OOP abuse, and other horrendous things (I apologize to the language puritans out there in advance). These are the things that I don't pledge allegiance to, but I included anyway (I was hoping to ticketize/discuss them):

- The abuse of OOP (also known as, "**why is there a mixin in my diff**")
- Is that a custom __getattr__ implementation?
- The wonky adapter pattern in the constructor
- The if-else clauses in the constructor (specially for links) so as to be too smart around the code throwing various types at it.

Most of these were added to introduce a "working copy" in under a 300 line diff. I hope we can slowly pour in upgrades and cleanups to get rid of the mixin, the "custom builtins" and the "why does this code try handle 3 different things at the same time" instances.


in_toto/models/common.py
```diff
@@ -32,18 +32,85 @@
...
+
+    signature = securesystemslib.keys.create_signature(key, repr(self.signed))
+    self.signatures.append(signature)
+
+  """ FIXME: This is mostly sintactic sugar and backwards compatibility stuff
+  """
+  def __getattr__(self, item):
+    # if the item was malformed, then return none
+    if item == 'signatures' or item == 'signed':
+        return None
```
lukpueh: Fix 4 space indentation

in_toto/models/common.py
```diff
@@ -32,18 +32,85 @@
...
+  def sign(self, key):
+    """Signs the canonical JSON representation of itself (without the
+    signatures property) and adds the signatures to its signature properties."""
+
+    securesystemslib.formats.KEY_SCHEMA.check_match(key)
+
+    signature = securesystemslib.keys.create_signature(key, repr(self.signed))
+    self.signatures.append(signature)
+
+  """ FIXME: This is mostly sintactic sugar and backwards compatibility stuff
```
lukpueh: typo: sintactic

in_toto/models/common.py
```diff
@@ -32,18 +32,85 @@
...
+  """
+
+  def __init__(self, **kwargs):
+
+    self.signatures = []
+    if "signatures" in kwargs:
+      self.signatures = kwargs['signatures']
+      del kwargs['signatures']
+
+    signable = self.get_signable()
```
lukpueh, line 53: Should we define `get_signable` in the base class `Metablock` as well? Something like:

```python
def get_signable(self):
  raise NotImplementedError()
```

in_toto/models/common.py
```diff
@@ -32,18 +32,85 @@
...
+        self.signed = signable(**kwargs['signed'])
+    else:
+        self.signed = signable()
+
   """Objects with base class Metablock have a __repr__ method
   that returns a canonical pretty printed JSON string and can be dumped to a
   file """
   def __repr__(self):
-    return canonicaljson.encode_pretty_printed_json(attr.asdict(self))
+    return '{{"signed": {}, "signatures": {}}}'.format(self.signed,
```
lukpueh: Mind adding a comment about double braces and replacement fields and?

in_toto/models/common.py
```diff
@@ -32,18 +32,85 @@
...
+    signatures property) and adds the signatures to its signature properties."""
+
+    securesystemslib.formats.KEY_SCHEMA.check_match(key)
+
+    signature = securesystemslib.keys.create_signature(key, repr(self.signed))
+    self.signatures.append(signature)
+
+  """ FIXME: This is mostly sintactic sugar and backwards compatibility stuff
+  """
+  def __getattr__(self, item):
```
lukpueh: Please provide a docstring with more details

in_toto/models/layout.py
```diff
@@ -275,7 +295,7 @@ def _validate_steps_and_inspections(self):
       names_seen.add(inspection.name)
 
 @attr.s(repr=False, init=False)
-class Step(models__common.Metablock):
+class Step(models__common.Signable):
```
lukpueh: Step is not "signable". Please replace base class `Signable` with `ValidationMixin`.

in_toto/models/layout.py
```diff
@@ -383,7 +403,7 @@ def _validate_expected_command(self):
 
 
 @attr.s(repr=False, init=False)
-class Inspection(models__common.Metablock):
+class Inspection(models__common.Signable):
```
lukpueh: Inspection is not "signable". Please replace base class `Signable` with `ValidationMixin`.

in_toto/models/common.py
```diff
@@ -32,18 +32,85 @@
...
+      del kwargs['signatures']
+
+    signable = self.get_signable()
+    if 'signed' in kwargs:
+      if isinstance(kwargs['signed'], Signable):
+        self.signed = kwargs['signed']
+      else:
+        self.signed = signable(**kwargs['signed'])
+    else:
+        self.signed = signable()
```
lukpueh: Fix 4 space indentation

in_toto/models/common.py
lukpueh: I wonder if we shouldn't make the effort and fix all the unit tests, etc. now at a time where we don't really care for backwards compatibility?
Also, overriding `__getattr__` is just part of the story, as can be seen in  https://github.com/in-toto/in-toto/pull/120/commits/c2f703085eeadd4a7d8dcc5f26a8307ed404f977

in_toto/models/common.py
```diff
@@ -69,47 +136,16 @@ def validate(self):
           method[1]()
 
 @attr.s(repr=False, init=False)
-class Signable(Metablock):
+class Signable(ValidationMixin):
```
lukpueh: Class docstring and header block need to be updated as well.

in_toto/models/common.py
```diff
@@ -69,47 +136,16 @@ def validate(self):
...
-        key = keys_dict[keyid]
-      except KeyError:
-        raise SignatureVerificationError(
-            "Signature key not found, key id is '{0}'".format(keyid))
-      if not securesystemslib.keys.verify_signature(
-          key, signature, self.payload):
-        raise SignatureVerificationError("Invalid signature")
 
+  """Objects with base class Signable have a __repr__ method
+  that returns a canonical pretty printed JSON string and can be dumped to a
```
lukpueh: I'd remove the "and can be dumped to a file".

in_toto/models/layout.py
```diff
@@ -77,31 +77,10 @@ class Layout(models__common.Signable):
...
-    self.inspect = kwargs.get("inspect", [])
-    self.keys = kwargs.get("keys", {})
-    self.readme = kwargs.get("readme", "")
 
-    # Assign a default expiration (on month) if not specified
-    self.expires = kwargs.get("expires", (datetime.today() +
-        relativedelta(months=1)).strftime("%Y-%m-%dT%H:%M:%SZ"))
-
-    self.validate()
+  def get_signable(self):
```
lukpueh, line 72: This needs a docstring

in_toto/models/common.py
```diff
@@ -32,18 +32,85 @@
...
+@attr.s(repr=False, init=False)
 class Metablock(object):
+  """ This object holds the in-toto metablock data structure. This includes
+  the fields "signed" and "signatures", i.e., what was signed and the
+  signatures. Other convenience classes will inherit this class to provide
+  serialization and signing capabilities to in-toto metadata.
+  """
+
+  def __init__(self, **kwargs):
+
```
lukpueh: Please add a docstring that explains your adaptor wizardry.

in_toto/models/layout.py
```diff
@@ -47,8 +47,8 @@
 # import validators
 import in_toto.models.common as models__common
 import in_toto.models.link as models__link
-@attr.s(repr=False, init=False)
-class Layout(models__common.Signable):
+
+class Layout(models__common.Metablock):
```
lukpueh, line 51: Doesn't below docstring belong to `_Layout` now? I'd rather explain here how and why `Layout` holds a `_Layout` in its `signed` property (inherited from the `Metablock` base class).

in_toto/models/layout.py
```diff
@@ -120,15 +99,21 @@ def read_from_file(filename='root.layout'):
...
     """Static method to instantiate a new Layout from a Python dictionary """
+    signatures = data.get('signatures', [])
     steps = []
-    for step_data in data.get("steps"):
+
+    if 'signed' not in data:
+        data['signed'] = _Layout()
+
+    for step_data in data['signed'].get("steps"):
+      # FIXME: is this really needed?
```
lukpueh: I think so yes, since we don't want to assign a list of step dictionaries to the `steps` attribute but a list of `Step` objects.

in_toto/models/layout.py
```diff
@@ -203,6 +193,36 @@ def import_step_metadata_from_files_as_dict(self):
 
     return step_link_dict
 
+
+@attr.s(repr=False, init=False)
+class _Layout(models__common.Signable):
```
lukpueh: What about a more descriptive name, like e.g. `LayoutSignable`, or `LayoutPayload`? What do you think?

in_toto/models/link.py
```diff
@@ -26,8 +26,7 @@
 FILENAME_FORMAT = "{step_name}.{keyid:.8}.link"
 UNFINISHED_FILENAME_FORMAT = ".{step_name}.{keyid:.8}.link-unfinished"
 
-@attr.s(repr=False, init=False)
-class Link(models__common.Signable):
+class Link(models__common.Metablock):
```
lukpueh, line 29: See comment about `Layout` vs. `_Layout` docstrings. IMO same applies here.

in_toto/models/link.py
```diff
@@ -60,25 +59,22 @@ class Link(models__common.Signable):
...
+    if 'signed' not in kwargs:
+      new_kwargs = {}
+      new_kwargs['signed'] = {}
+      new_kwargs['signed'].update(kwargs)
+      kwargs = new_kwargs
 
+    super(Link, self).__init__(**kwargs)
+
+  def get_signable(self):
+    """ this method is used by the base class's constructor to obtain the
```
lukpueh: We usually start docstrings with a capital letter.

lukpueh (changes requested):
Thanks for the PR, @SantiagoTorres! I think it's an elegant solution.

As a few general notes:
- I think the validator mixin is alright, but I would probably disassociate it from `Signable` and use it independently
- I think we should get rid of the custom `getattr` now, as it does not really help with backwards compatibility
- The manifestations of your *wonky adapter pattern* need some more comments, and if I understood your correctly in that it's just a temporary solution, I would also add some inline `TODO`s or so.

lukpueh:
Also note that the PR breaks the [demo](https://github.com/in-toto/in-toto/tree/develop/demo) (and the [opensuse-demo](https://github.com/in-toto/in-toto/tree/develop/demo-opensuse)).

lukpueh:
Just spotted a couple of more indentation mismatches (4 instead of 2 spaces). Please fix.

in_toto/models/common.py
SantiagoTorres: Yup, sounds  like a good idea.

in_toto/models/layout.py
SantiagoTorres: That's probably clearer in the long term. I adopted this approach because of the `_` prefix, but it's probably not that clear/useful. Will change.

in_toto/models/common.py
SantiagoTorres: Yeah, I had a version in which I also overwrote `__setattr__` but it felt particularly insulting.

I can do a bulk replace of these changes all throughout the tests, but I probably want to separate the bulk rename of the rest of the modules to a separate PR.

in_toto/models/layout.py
```diff
@@ -120,15 +99,21 @@ def read_from_file(filename='root.layout'):
   @staticmethod
   def read(data):
     """Static method to instantiate a new Layout from a Python dictionary """
+    signatures = data.get('signatures', [])
     steps = []
-    for step_data in data.get("steps"):
+
+    if 'signed' not in data:
+        data['signed'] = _Layout()
```
lukpueh: I think you could leave this out (which would also prevent coverage decline), as you do the same thing in `common.Metablock`'s constructor anyway, right? https://github.com/in-toto/in-toto/blob/c1e0cf122f2e1faf6583784517a23cb752a0fc77/in_toto/models/common.py#L56-L57


You could wrap the lines below in a `if "signed" in data:`, or something like `data.get("signed", {}).get("steps")` which I like but others might find ugly.

in_toto/models/layout.py
```diff
@@ -177,12 +162,17 @@ def import_step_metadata_from_files_as_dict(self):
           pass
 
         else:
+
+          if 'signed' not in link_obj:
+            raise in_toto.exceptions.LinkNotFoundError("Invalid format. '{0}'"
+                " is not a valid in-toto metadata file.".format(filename))
```
lukpueh: Can you add a unittest for this line?

SantiagoTorres: interesting you proposed adding `data.get("signed", {}).get("steps")` above but not here. Any reasoning behind this?

I'm reluctant to add a UT on this line as this feels like code that should be refactored (why don't we have a schema run on this?) why do we use schema anyway?

in_toto/models/layout.py
SantiagoTorres: Sure will do, but I'm less worried about losing 0.01% of coverage than having an increasingly convoluted constructor.

 I do think that this is bringing up the bad coupling patterns that we have on our constructors though. This spaghetti code is just reflecting the fact that we'd rather use .read and inline the json rather than instantiating metadata objects all throughout the code.

in_toto/models/layout.py
lukpueh: Only negligible reasoning really. :) But yes, I thought about proposing the same solution here as well. What do you think?

The reason why we don't schema validate her is because we don't have link schema validation yet (which we should add https://github.com/in-toto/in-toto/issues/122).

lukpueh (approved):

lukpueh:
Thanks @SantiagoTorres!

PR #121, opened 2017-09-26
Add optional argument flag for no command

### LINKED ISSUE #10: Allow no-command (signed-off-by) steps
@JustinCappos suggested that there might be steps that are not related to a command or any file modifications, e.g. for code reviews.

We should add  "no command" options (probably only to the `qa` page) to accommodate such needs.  Also we probably need to add a `no-command` flag to `in-toto-run`.

ShikherVerma:
* **What does this PR do?**
Adds `--no-command` argument flag to `in-toto-run` for steps which don't have a command such as code review or quality assurance.

* **Are there points in the code the reviewer needs to double check?**
Commands like `in-toto-run --step-name name --key bob --no-command -- cmd` work even though `--no-command` specifies that there will be no `cmd`.  The above will run by ignoring the `cmd` if both `cmd` and `--no-command` are given.

@lukpueh 

* **What are the relevant issue numbers fixed?**
Partially fixes [layout-web-tool#10](https://github.com/in-toto/layout-web-tool/issues/10)

### DIFF
--- a/in_toto/in_toto_run.py
+++ b/in_toto/in_toto_run.py
@@ -98,6 +98,7 @@ def main():
                "[--materials <filepath>[ <filepath> ...]]\n{0}"
                "[--products <filepath>[ <filepath> ...]]\n{0}"
                "[--record-byproducts]\n{0}"
+               "[--no-command]\n{0}"
                "[--verbose] -- <cmd> [args]\n\n"
                .format(lpad))
 
@@ -120,12 +121,16 @@ def main():
       help="If set redirects stdout/stderr and stores to link metadata",
       dest="record_byproducts", default=False, action="store_true")
 
+  in_toto_args.add_argument("-x", "--no-command",
+      help="Set if step does not have a command",
+      dest="no_command", default=False, action="store_true")
+
   in_toto_args.add_argument("-v", "--verbose", dest="verbose",
       help="Verbose execution.", default=False, action="store_true")
 
   # FIXME: This is not yet ideal.
   # What should we do with tokens like > or ; ?
-  in_toto_args.add_argument("link_cmd", nargs="+",
+  in_toto_args.add_argument("link_cmd", nargs="*",
     help="Link command to be executed with options and arguments")
 
   args = parser.parse_args()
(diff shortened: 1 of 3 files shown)


SantiagoTorres:
This is probably a cleaner way to do it, but `in-toto-run ... -- :` exists 😛

in_toto/in_toto_run.py
```diff
@@ -142,7 +147,13 @@ def main():
     log.error("in load key - {}".format(args.key))
     sys.exit(1)
 
-  in_toto_run(args.step_name, args.materials, args.products,
+  if args.no_command:
+    in_toto_run(args.step_name, {}, {}, [], key, args.record_byproducts)
```
lukpueh: We still want to be able to record the files someone is looking at (signing off). Please pass `args.materials` and `args.products` as 2nd and 3rd argument respectively.

in_toto/runlib.py
```diff
@@ -374,8 +374,11 @@ def in_toto_run(name, material_list, product_list,
     log.info("Recording materials '{}'...".format(", ".join(material_list)))
   materials_dict = record_artifacts_as_dict(material_list)
 
-  log.info("Running command '{}'...".format(" ".join(link_cmd_args)))
-  byproducts, return_value = execute_link(link_cmd_args, record_byproducts)
+  if link_cmd_args:
+    log.info("Running command '{}'...".format(" ".join(link_cmd_args)))
+    byproducts, return_value = execute_link(link_cmd_args, record_byproducts)
+  else:
+    byproducts, return_value = 0, 0
```
lukpueh: Not sure what `byproducts` and `return_value` should be if no command was executed. I wouldn't use `0` though, because that's a actually common return value.

In `in_toto_record`, where we also generate a Link without having executed a command, we do `byproducts={}` and `return_value=None`. What do you think?

lukpueh (changes requested):
Great PR description, great commit messages, great patch! Thanks, @ShikherVerma! Please consider my comments an we are good to merge.

in_toto/runlib.py
ShikherVerma: I think `byproduct={}` and `return_value=None` is better. I'll switch to that.

PR #123, opened 2017-09-28
Add link validation (fix #122)

### LINKED ISSUE #122: Add Link schema validation methods
We need [schema validation](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/common.py#L47-L69) for Links similar to what we do for [Layouts](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/layout.py#L206-L275), [Steps](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/layout.py#L337-L381) and [Inspections](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/layout.py#L432-L460).

lukpueh:
This PR adds validation methods to `link.LinkSignable` that check the format of the corresponding fields. (closes #122)

By calling `self.validate()` on a`LinkSignable` object  all fields' formats are checked.

This PR also
- adds validation to `LinkSignable`'s constructor,
- adapts existing unit tests where schema validation was less strict, and
- adds new unit tests for the new validation methods.


in_toto/models/link.py
```diff
@@ -141,3 +141,59 @@ def __init__(self, **kwargs):
     self.byproducts = kwargs.get("byproducts", {})
     self.command = kwargs.get("command", [])
     self.return_value = kwargs.get("return_value", None)
+
+    self.validate()
+
```
SantiagoTorres, line 152: I think you need another space here, or less space on the ones below

in_toto/models/link.py
```diff
@@ -141,3 +141,59 @@ def __init__(self, **kwargs):
...
+      raise securesystemslib.exceptions.FormatError(
+          "Invalid Link: field `_type` must be set to 'link', got: {}"
+          .format(self._type))
+
+
+  def _validate_materials(self):
+    """Private method to check that `materials` is a `dict` of `HASHDICTs`."""
+    if not isinstance(self.materials, dict):
+      raise securesystemslib.exceptions.FormatError(
+          "Invalid Link: field `materials` must be of type list, got: {}"
```
SantiagoTorres, line 166: I wonder if we can coalesce this to a format string (probably with our own exception class) and just do:
```
raise InvalidLinkException(field='materials', got=type(self.materials))
```

Here and elsewhere.

test/models/test_link.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python
+"""
+<Program Name>
+  test_link.signed.py
+
+<Author>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+
+<Started>
+  Sep 28, 2011
```
SantiagoTorres: I don't think this is the right date

test/models/test_link.py
```diff
@@ -0,0 +1,124 @@
...
+    with self.assertRaises(FormatError):
+      test_link.signed.validate()
+
+
+  def test_validate_return_value(self):
+    """Test `return_value` field. Must be either an `int` or `None` """
+    test_link = Link()
+
+    # Good return_value 1
+    test_link.signed.retunr_value = 1
```
SantiagoTorres: I don't think this is checking what it should check.

SantiagoTorres (changes requested):
Four really minor details. I think once these are addressed this is ready to merge.

in_toto/models/link.py
lukpueh: We could and should, but let's do this elsewhere. Care for a "revisit exception taxonomy" ticket?

lukpueh:
Thanks for the comments, @SantiagoTorres. Please merge when ready!

in_toto/models/link.py
lukpueh: FYI: https://github.com/in-toto/in-toto/issues/126

SantiagoTorres (approved):

PR #124, opened 2017-09-29
Fix README introduction

### LINKED ISSUE #114: Readme could use a bit more explanation about in-toto workflow
The [readme](https://github.com/in-toto/in-toto/blob/develop/README.md#in-toto--) explains in-toto's general purpose briefly, then starts explaining what a layout contains, without providing the few sentences that I think would help a reader understand the in-toto workflow.

In particular, I think that the last paragraph of the first section of the readme (*sorry, can't link line numbers in .md in github*) would be better with a little bit more detail. For example, I might change it to something like this:

> ### In-Toto Workflow
> in-toto requires a **project layout** that specifies the functionaries and the tasks they are supposed to perform. The project designer (((or whatever term is preferred))) creates this project layout as a sequence of individual steps, referred to as **links**, designating functionaries to be trusted for each step. The functionaries in charge of each link produce link metadata once they've completed the work in their link, and provide this metadata to (etc.....). The chain of link metadata travels with the final product, and can be validated manually by the final client or via automated tooling in, for example, a package manager.

lukpueh:
Following suggestions from @awwad and @aaaaalbert this PR aims at clarifying in-toto terminology and workflow of the README introduction section. (closes #114)

This PR also updates the installation instructions and fixes indentation in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,11 @@ pip install git+git://github.com/in-toto/in-toto.git
 
 The in-toto software supply chain layout consists of the following parts:
  - **expiration date**
+ - **readme** (an optional description of the supply chain)
  - **functionary keys** (public keys, used to verify link metadata signatures)
  - **signatures** (one or more layout signatures created with the project owner key(s))
- - **software supply chain steps** correspond to steps carried out by a functionary as part of the software supply chain. The steps defined in the layout list the functionaries who are authorized to carry out the step (by key id). Steps require a unique name to associate them (upon verification) with link metadata that is created when a functionary carries out the step using the `in-toto` tools.
-Additionally, steps must have material and product rules which define the files a step is supposed to operate on. Material and product rules are described in the section below.
+ - **software supply chain steps**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,11 +2,12 @@
 
 Protecting the software supply chain integrity
 
-in-toto guarantees that the end-user (or client) is able to verify that the entire development life cycle has been conducted as per the specified layout and that each of the functionaries (eg. developers) have performed the specified tasks and there haven't been any malicious changes in the files.
+in-toto provides a framework to verify that the entire supply chain of a software product has been conducted as planned, by authorized personnel only, and that there hasn't been any tampering with the product in transit.
```
awwad: 👍

I'm not sure if one conducts a supply chain, though. Could you call a supply chain a sequence of events, or is it more like the machinery that executes those? If the latter case, the wording is weird.

README.md
```diff
@@ -2,11 +2,12 @@
...
 
-in-toto requires a project layout that specifies the functionaries and the tasks they are supposed to perform.
-After each functionary performs his or her task a link metadata is generated.
-This metadata is used to verify the intermediate and final products with the project layout.
+in-toto requires a **project owner** to create a **layout**. A layout lists the sequence of **steps** of the software supply chain, and the **functionaries** authorized to perform these steps.
+When a functionary performs a step in-toto gathers information about the used command and the related files and stores it to a **link** metadata file. As a consequence link files provide the required evidence to establish a continuous chain that can be validated against the steps defined in the layout.
```
awwad: stores it **in**, not **to** (saves or writes it **to**, if you want)

I like the listing of and emphasis on the conceptual bricks of the system.

awwad (commented):
It's clearer. (:

Issue #118, opened 2017-09-14
Fix in-toto-sign args/usage message


lukpueh:
As pointed out by @awwad https://github.com/in-toto/layout-web-tool/pull/14#discussion_r137356278 the usage message of `in-toto-sign` is inaccurate.

The short option form `-k KEYS` does not actually work with the indicated whitespace.

Also, I think we should change the subparser names. `in-toto-sign sign ...` sounds stupid.

lukpueh:
Fixed with #125

PR #125, opened 2017-10-03
Fix in-toto-sign args (#118) + huge refactor 

lukpueh:
This commit started as fix for in-toto/in-toto#118 but ended up as bigger refactor of the signing and signature verification command line tool. IMHO the structure makes more sense now.

It's probably easiest to just take a look at the code as it is now and not think too much about the diff.

The fix:
- fix whitespace problem with `-k KEYS [KEYS ...]` by replacing positional argument `signablepath` with required optional argument `-f FILE` (c.f. [python issue9338](https://bugs.python.org/issue9338) and [stackoverflow](https://stackoverflow.com/questions/26985650/argparse-do-not-catch-positional-arguments-with-nargs))
- fix funny `in-toto-sign sign` idiom by removing subparsers, making signing the default action and adding a `--verify` option

The refactor:
- Reduce redundancy by merging signing functionality
- Declutter huge nested control flow in `main()`
- Reduce docstrings to minimum
- Add early argument checking for conflicting args
- replace `-d, --destination` with `-o, --output`
- replace `-i, --infix` with `-x, --infix`

This PR also updates the unit tests to conform with the interface changes and removes tests for internally called helper functions and tests the interface more thoroughly instead.
It also removes big parts of test case setup as this can be done as part of the tests.

### DIFF
--- a/in_toto/in_toto_sign.py
+++ b/in_toto/in_toto_sign.py
@@ -32,6 +32,7 @@
 
     - verify signatures
 
+
   Usage:
   ```
   in-toto-sign [-h] -f FILE -k KEYS [KEYS ...] [-o OUTPUT] [-x] [-r] [-v]
@@ -166,10 +167,10 @@ def _load_metadata(file_path):
     with open(file_path, "r") as fp:
       file_object = json.load(fp)
 
-    if file_object["signed"].get("_type", {}) == "link":
+    if file_object.get("signed", {}).get("_type") == "link":
       return Link.read(file_object)
 
-    elif file_object["signed"].get("_type", {}) == "layout":
+    elif file_object.get("signed", {}).get("_type") == "layout":
       return Layout.read(file_object)
 
     else:
--- a/in_toto/in_toto_sign.py
+++ b/in_toto/in_toto_sign.py
@@ -4,6 +4,7 @@
 
 <Author>
   Sachit Malik <i.sachitmalik@gmail.com>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
 
 <Started>
   June 13, 2017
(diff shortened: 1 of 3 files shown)


in_toto/in_toto_sign.py
```diff
@@ -18,11 +18,19 @@
...
+    - write the signed file to a path consisting of
+      the value from the metadata's `name` field, the first 8 characters
+      of the signing key's id as infix and '.link' as extension,
+      e.g.:  "package.c1ae1e51.link"
+      Note:
+        The naming scheme is used to distinguish Link files of steps that are
+        required to be carried out by a threshold of functionaries.
+        This option is only available for Link metadata.
+        If multiple keys are passed for signing, the short key id of the last
+        key in the arguments list is used as infix.
```
awwad: Minor: This is a lot of detail for filename naming explanation, but I don't see where else I'd put it, so perhaps here is the best place for it. Is any explanation required as to the formats of keys or how the signing is done? Is there some place you can refer to that specifies the expected format of key objects or something?

awwad:
When a step is expected to be signed off by multiple functionaries, they're generally expected to produce their own, separate link files, which they sign separately. Since one wouldn't expect multiple signatures on a .link file (except in some edge cases, maybe), going out of your way to support that for this tool doesn't make sense, especially if it costs you having to give up one-to-one mapping in the output filenames (using the last key in a set of keys in the filename). So this tool should either not take multiple keys when signing links (but still take multiple keys when signing layouts) or, when provided multiple keys for signing links, produce one link file for each key provided, using that key as the infix in the filename, restoring one-to-one mapping.

With that problem settled, the infixing behavior for filenames should probably be the default (default 
'package.key12345.link' instead of 'package.link'; then you won't need the -x flag, and you'll be less likely to mix files up. If someone wants to produce 'package.link', they can always specify the output filename.

Also, in our chat, we talked about default behavior, and you noted that you thought the default behavior when an output file already exists (replace signature vs add signature) should be consistent for layouts and links, and that the replace functionality should be the default, and that add should not be enabled for links.

in_toto/in_toto_sign.py
```diff
@@ -16,38 +16,37 @@
...
   Examples:
   ```
-  # Sign Layout with two keys and write to specified path
-  in-toto-sign -f unsigned.layout -k priv_key1 priv_key2 -o root.layout
+  # Append two signatures to layout file and write to passed path
+  in-toto-sign -f unsigned.layout -k priv_key1 priv_key2 -o root.layout -a
 
   # Sign Link and use passed key's short id as filename infix
-  in-toto-sign -f package.c1ae1e51.link -k priv_key -x
+  in-toto-sign -f package.c1ae1e51.link -k priv_key
```
awwad: I was briefly confused about -f and -o reading this. Consider:
```
# Sign Link and write to default output filename (using key's short id as infix)
```
or
```
# Sign Link data from specified file. Since -o is not specified, write to
# default output filename, using the short id for priv_key as a filename infix
# (in place of "c1ae1e51")
```

Alternatively, you could add an example that uses -o:

```
# Re-sign Link and write output to a specific filename.
in-toto-sign -f package.old_sig.link -k priv_key -o package.new_sig.link
```

It could also just be obvious enough already. (:

in_toto/in_toto_sign.py
```diff
@@ -16,38 +16,37 @@
...
+    - replace (default) or add signature(s):
+        - Layout metadata can be signed by multiple keys at once,
+        - Link metadata can only be signed by one key at a time
+    - write signed metadata to a specified path:
+      if no output path is specified,
+        - layout metadata is written to the input file,
+        - link metadata is written to:
+         "<step name>.<short signing key id>.link"
 
     - verify signatures
```
awwad: The behavior of in-toto's link signature verification was a little surprising to me (but maybe I've been using TUF too long), so it's probably worth noting here (or perhaps in the _verify_metadata docstring). You probably also want this clarified deeper in, in verifylib.py, and possibly other places:

**When verifying a Link, if there are *additional* signatures on that Link beyond what is required by the Layout, an AuthorizationError("Unauthorized Key! ...") exception is raised, even if all the required signatures are provided.**

(As a side note, it's worth having thought a bit about whether or not this is the ideal behavior, if you haven't. There are integration edge cases where you could imagine wanting extra signatures - supporting outdated Layouts controlled by other distributors, using Link files tangentially for other purposes.)

awwad (commented):
LGTM otherwise. I haven't carefully reviewed the tests, FWIW.

lukpueh:
Thanks for your valuable comments @awwad! I adopted one suggestion (376fae822b14d4333da72c4e16047cffffff2960) and will address the other one with this ticket https://github.com/in-toto/in-toto/issues/135.

PR #134, opened 2017-10-16
Fix comments and unused variable in `verify_threshold_constraints`

lukpueh:
Removing signature verification in verifylib's `verify_threshold_constraints` was requested in
#82 and delivered in #85.

However, the PR forgot to update a comment and remove an obsolete variable assignment, which is made up for here.

This PR further updates the function's docstring trying to better convey its purpose.

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -984,10 +984,8 @@ def verify_threshold_constraints(layout, chain_link_dict):
     reference_keyid = key_link_dict.keys()[0]
     reference_link = key_link_dict[reference_key]
 
-    # Iterate over all links to check their signature and
-    # compare their properties with a reference_link
+    # Iterate over all links to compare their properties with a reference_link
     for keyid, link in six.iteritems(key_link_dict):
-      keys_dict = {keyid: layout.keys.get(keyid)}
 
       # compare their properties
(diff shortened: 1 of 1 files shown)


in_toto/verifylib.py
```diff
@@ -984,10 +985,8 @@ def verify_threshold_constraints(layout, chain_link_dict):
     reference_keyid = key_link_dict.keys()[0]
     reference_link = key_link_dict[reference_key]
 
-    # Iterate over all links to check their signature and
-    # compare their properties with a reference_link
+    # Iterate over all links to compare their properties with a reference_link
     for keyid, link in six.iteritems(key_link_dict):
-      keys_dict = {keyid: layout.keys.get(keyid)}
```
SantiagoTorres, line 990: Hmm, this change is confusing to me. Can you elaborate on what is this doing exactly?

lukpueh: This change removes a variable that became obsolete with the removal of signature verification inside `verify_threshold_constraints` in #85 more specifically 3deb4bbb97e5f0828c855759030c83f191a145c4.

SantiagoTorres: Oh ok. Thanks for answering!

Issue #136, opened 2017-10-16
Revisit threshold verification


lukpueh:
Verifying step thresholds, i.e. enforcing a specified  amount of functionaries, to each provide a link file corresponding to a given step, happens in parts explicitly and in parts implicitly in various places throughout the final product verification, i.e. when:

1. [importing step metadata](https://github.com/in-toto/in-toto/blob/1ae56cd767e60982cb273edb77e582cc821bd7f9/in_toto/models/layout.py#L171-L173)
2. [verifying step signatures](https://github.com/in-toto/in-toto/blob/1ae56cd767e60982cb273edb77e582cc821bd7f9/in_toto/verifylib.py#L250-L257)
3. [verifying threshold constraints](https://github.com/in-toto/in-toto/blob/1ae56cd767e60982cb273edb77e582cc821bd7f9/in_toto/verifylib.py#L969-L972)

(1) Tries to load a link file for every functionary authorized to carry out a step by looking for link files with the respective functionaries' keyid as filename infix. E.g. if a layout authorizes functionaries with keyids `A, B, C` and short keyids (`A', B', C'`) to carry out a step `foo` this function will look for `foo.A'.link`, `foo.B'.link`, and `foo.C'.link`.

Before returning the populated data structure of "keyid - link objects" pairs per step name, the function also checks if there are enough links with distinct keyid'ed names as specified in the corresponding step's threshold and raises `LinkNotFoundError` if not.

However, this function does not verify if the actual signatures correspond with the keyids in the filenames, and even less if those signatures are valid.

This happens in (2), which iterates over the "keyid - link objects" pairs as imported in (1) and verifies each link's (could also be a sublayout) signature with the key for that keyid.

Finally, (3) re-verifies if the data structure returned from (1) (which in the meanwhile has [replaced sub-layouts with summary links](https://github.com/in-toto/in-toto/blob/1ae56cd767e60982cb273edb77e582cc821bd7f9/in_toto/verifylib.py#L1206)) has enough links per step to meet each step's threshold and then verifies if all links corresponding to a given step list the same materials/products.

I think we should try to make the above described "threshold verification process" more transparent. Maybe we can address this issue while dealing with #135.

lukpueh:
Fixed with #170

Issue #135, opened 2017-10-16
Revisit signature verification


lukpueh:
The signature verification method of `Link` and `Layout` objects currently fails if:
- the passed key dictionary does not match the [`KEYDICT_SCHEMA`](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/formats.py#L383-L386),
- the object has no signature,
- there is a signature for which no key is found in the passed key dictionary, which means that
  multiple signatures have to be verified at once (*), 
- and finally if any of the verified signatures is actually broken

@awwad has pointed out that this (*) might be [unexpected behavior](https://github.com/in-toto/in-toto/pull/125#discussion_r144058807).

Moreover, like this links and sublayouts that carry multiple signatures from different keys would always fail, because we always [pass only one key to the signature verification function](https://github.com/in-toto/in-toto/blob/develop/in_toto/verifylib.py#L250-L257).

I think we should change this behavior and while we are at it we should better separate the following concerns:
- the actual cryptographic signature verification
- which keys are authorized to sign
  - for links and sublayouts this is defined in the corresponding step's `pubkeys` field
  - for layouts we have to assume that the key(s) passed to the verification tool are legitimate
- which keys are required to sign
  - for links this is assessed using the corresponding step's `threshold` and `pubkeys` fields, i.e. there must be `step.threshold` links and each one must have a signature, created with a distinct key from the list of `step.pubkeys`
  - for layouts I suggest we require at least one key, and if multiple keys are passed, each one of them must have a valid signature
  - for sublayouts this is somewhat harder. Maybe we should, [when recursing into (sub)layout verification](https://github.com/in-toto/in-toto/blob/develop/in_toto/verifylib.py#L1073-L1084), pass a flag to disable regular layout signature verification? Alternatively, we could pass only the keys for which the sublayout carries signatures, as long as these keys are authorized in the corresponding step's `pubkeys` filed 

*Note: Please fix comments/docstrings (also in in `in_toto_sign.py`) when dealing with this issue*

PR #139, opened 2017-10-18
Remove Link's return value field and store it as part of byproducts

### LINKED ISSUE #133: Adopt move return value inside byproduct dictionary in link metadata
The [current specification](https://github.com/in-toto/docs/blob/69c29b41f4c1fe064b5d77d61d4b268fc171bcaa/in-toto-spec.md#44-file-formats-namekeyid-prefixlink) lists the return value as part of the opaque byproduct dictionary and not as separate field. 

This change should be propagated to the [in-toto inspection command line tools](https://github.com/in-toto/in-toto-inspections).

lukpueh:
Adopting specification metadata format changes (closes #133)

The PR removes the `Link` field `return_value` and modifies `runlib` to store the return value of an executed step or inspection command in the opaque byproducts dictionary instead.

The PR also renames the `in-toto-run` command line option `--record-byproducts` to `--record-streams`, which seems more accurate, as it toggles whether stdout and stderr should be redirected and stored as part of byproducts, while the new `return-value` field inside the byproducts dictionary is always set, independently of this option.


lukpueh:
Let's ignore the decrease in coverage. It does not seem like it is actually related to the contents of the PR but rather due to some funny arithmetic. Tests for the file that caused the decrease should be dealt with when fixing  #135 and #136.

PR #140, opened 2017-10-18
Remove implicit `DISALLOW *` from rule verification

### LINKED ISSUE #131: Adopt replace implicit DISALLOW with ALLOW
In https://github.com/in-toto/docs/pull/2 the implicit `DISALLOW *` at the end of each list of artifact rules (`expected_materials`, `expected_products`) was changed to an implicit `ALLOW *`.

Adopting the change might be as easy as removing [this check in the rule verification routine](https://github.com/in-toto/in-toto/blob/develop/in_toto/verifylib.py#L894-L899).

The rationale for this spec change will be discussed in https://github.com/in-toto/docs/issues/4.

lukpueh:
Closes #131 

Until now every recorded artifact had to be consumed by a rule. That is, when verifying the batch of rules for a certain artifact type (material or product) reported for a step or inspection, the verification routine would keep a temporary queue of all artifacts of that type for that step or inspection and remove them from the queue each time an artifact was matched by a rule's artifact pattern.

If any artifacts remained in the queue after verifying all rules (for that artifact type of that step or inspection) a RuleVerficationError was raised.

This behavior amounted to an implicit `DISALLOW *` at the end of a given `expected_materials` or `expected_products` list.

With this change the implicit `DISALLOW *` is removed, i.e. replaced with an implicit `ALLOW *`.

However a layout designer can still add an explicit `DISALLOW *` at the end of a given rule list to fail if corresponding artifacts are not matched by any prior rule's pattern.

The rationale for this change can be found in in-toto/docs#4.

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -891,13 +891,6 @@ def verify_item_rules(source_name, source_type, rules, links):
       elif source_type == "products":
         source_artifacts_queue = source_products_queue
 
-  # All artifacts have to be consumed by a rule. If we have applied all rules
-  # of a list and there are still artifacts in the queue, we fail.
-  if source_artifacts_queue:
-    raise RuleVerficationError("{0} '{1}' were not authorized by any rule"
-        " of item '{2}'".format(source_type, source_artifacts_queue,
-            source_name))
-
 
 def verify_all_item_rules(items, links):
   """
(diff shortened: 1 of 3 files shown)


lukpueh:
Regarding coverage see https://github.com/in-toto/in-toto/pull/139#issuecomment-337649147

PR #141, opened 2017-10-18
Reorder final product verification routines

### LINKED ISSUE #132: Adopt order of final product verification routines
At some point the specification changed the order of the final product verification routines to only run the inspection commands after having successfully verified all steps' artifact rules.

IIRC, we also talked about only running an inspection command after having verified that inspection's material rules. However, the [current specification](https://github.com/in-toto/docs/blob/69c29b41f4c1fe064b5d77d61d4b268fc171bcaa/in-toto-spec.md#52-verifying-the-final-product) does not indicate that.

Note that this change has implications on the availability of matching targets to a given MATCH rule. So far we were able to match any material or product of any step or inspection, because at the time of rule verification, inspection had been executed and therefore, all link files were available.

Maybe this is something we should discuss/document in https://github.com/in-toto/docs/issues/4?

The relevant code can be found in [`verfylib.in_toto_verify`](https://github.com/in-toto/in-toto/blob/develop/in_toto/verifylib.py#L1162-L1170)

lukpueh:
Closes #132 

Postpone execution of Inspection commands to after having verified expected artifacts of Steps.

This guarantees that Inspection commands don't run on compromised target files, which would be a surface for attacks.

As a consequence `MATCH` rules defined for Steps won't be able to reference Inspection artifacts, because the required Inspection link metadata only gets generated when running the Inspections.

However, `MATCH` rules defined for Inspection are still able to reference Step artifacts.

This PR also updates and fixes a couple of minor issues in the long docstring of `verifylib.in_toto_verify`, which describes the final product verification routine.

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1169,7 +1169,7 @@ def in_toto_verify(layout, layout_key_dict):
         8.  Verify rules defined in each Step's expected_materials and
             expected_products field
             NOTE: At this point no Inspection link metadata is available,
-            hence (MATCH) rulescannot reference materials or products of
+            hence (MATCH) rules cannot reference materials or products of
             Inspections.
             Verifying Steps' artifact rules before executing Inspections
             guarantees that Inspection commands don't run on compromised
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1143,33 +1143,45 @@ def in_toto_verify(layout, layout_key_dict):
   <Purpose>
     Does entire in-toto supply chain verification of a final product
     by performing the following actions:
+
         1.  Verify layout signature(s)
+
         2.  Verify layout expiration
+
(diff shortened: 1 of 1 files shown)


in_toto/verifylib.py
```diff
@@ -1143,33 +1143,45 @@ def in_toto_verify(layout, layout_key_dict):
...
         6.  Verify alignment of defined (Step) and reported (Link) commands
             NOTE: Won't raise exception on mismatch
-        7.  Execute Inspection commands
+
+        7.  Verify threshold constraints
+
+        8.  Verify rules defined in each Step's expected_materials and
+            expected_products field
+            NOTE: At this point no Inspection link metadata is available,
+            hence (MATCH) rulescannot reference materials or products of
```
SantiagoTorres: space on rulescannot*

SantiagoTorres (commented):
can you update that typo?

PR #142, opened 2017-10-24
Refactor metadata model

lukpueh:
This is another attempt to fix the implementation-specification `has-a` vs. `is-a` metadata format mismatch as documented in #98.

Santiago's solution for this issue, turned out to be slightly over-engineered, which in all fairness he already hinted at in the PR's description (see #120). This PR simplifies the metadata model by removing the adapter pattern and custom `__getattr__` implementation. 

What remains is a container class for signed metadata (`Metablock`) with a `signed` and a `signatures` attribute and methods for JSON serialization and deserialization, reading and writing to disk, and creating and verifying signatures.
The `signed` attribute of `Metablock` holds either `Link` or `Layout` objects, which contain the actual metadata information. The `signatures` attribute holds the cryptographic signatures computed over the canonical json representation of the `Link` or `Layout` contained in the `signed` attribute.

It wasn't feasible to create small, self-contained commits that leave the system in a stable state. However, the commit messages can be used to incrementally understand the modifications.

I also used the chance for an overall review of the system and to make small updates and fixes, especially regarding comments and docstrings in various places.


in_toto/models/metadata.py
```diff
@@ -0,0 +1,207 @@
...
+  def __repr__(self):
+    """Returns a JSON string representation of the object."""
+    # the double {{'s is the escape sequence for an individual {. We wrap this
+    # under a format string to avoid encoding to json twice (which turns a json
+    # string into a string and so on...
+    # FIXME:
+    # We are mixing 3 JSON string formats here: The value of "signed" is
+    # "pretty printed canonical json", the value of "signatures" is
+    # "canonical json" and the container is just "json".
+    # Is this really what we want?
```
lukpueh, line 57: @SantiagoTorres, what do you think? I'd say, either we make it all (pretty printed) canonical JSON or none of it.
The only really important thing is that the JSON string representation of the object held in the `signed` attribute (Link or Layout) is deterministic when we create or verify signatures.

in_toto/runlib.py
```diff
@@ -31,12 +31,15 @@
 
 import in_toto.settings
 from in_toto import log
-from in_toto.models.link import (UNFINISHED_FILENAME_FORMAT, FILENAME_FORMAT)
+from in_toto.models.link import (UNFINISHED_FILENAME_FORMAT, FILENAME_FORMAT,
+    FILENAME_FORMAT_SHORT)
```
awwad, line 35: Remember that when you `from ... import` a string, it is copied from the source module at import time. Thus, if FILENAME_FORMAT is changed in any module after this module is imported the first time, the value in this module and the value in `in_toto.models.link` (or any module that uses a normal import of `in_toto.models.link`) will not match. Testing is one context in which such a pattern might be used, as would be any kind of initial configuration setup at a higher level than this code. (This is true not just of strings, but of anything immutable or any variable you reassign later.)

It doesn't necessarily mean you shouldn't do this, but it's worth keeping in mind. (I've been burned by this.)

in_toto/models/layout.py
```diff
@@ -222,6 +101,25 @@ def __init__(self, **kwargs):
 
     self.validate()
 
+
+  @staticmethod
+  def read(data):
+    """Static method to instantiate a new Layout from a Python dictionary """
+    signatures = data.get('signatures', [])
```
awwad: The signatures line here looks like it's not used - which is good, since it seems like it's none of class Layout's business. So this line was meant to be deleted and was missed, I think?

in_toto/models/layout.py
```diff
@@ -222,6 +101,25 @@ def __init__(self, **kwargs):
 
     self.validate()
 
+
+  @staticmethod
+  def read(data):
+    """Static method to instantiate a new Layout from a Python dictionary """
```
awwad: I think it would be good to have a bit more docstring here explaining what data looks like.

Is it like this?
`{"signed": {"steps": ..., "inspect": ..., ...}, "signatures": [...]}`

Or like this?
`{"steps": ..., "inspect": ..., ... }`

The latter, right?

in_toto/models/layout.py
```diff
@@ -294,7 +192,7 @@ def _validate_steps_and_inspections(self):
       names_seen.add(inspection.name)
 
 @attr.s(repr=False, init=False)
```
awwad, line 195: Recognizing and looking into [attrs](http://www.attrs.org/en/stable/api.html) took a little while. I like it, but it's probably worth adding an inline comment. What do you think? Something like `# automatically provide common class methods like __eq__()` 

Also, I admit that it took me a moment to not parse "attr.s" as "attr.string". (use of "serious business" alias attr.attrs might be better).

in_toto/models/layout.py
```diff
@@ -294,7 +192,7 @@ def _validate_steps_and_inspections(self):
...
-class Step(models__common.ValidationMixin):
+class Step(ValidationMixin):
```
awwad, line 196: I thought about whether or not this being the only mix-in - and the only class that Step, Inspection, and Signable inherit from - meant that it would be more readable to just make ValidationMixin into a generic in-toto class parent class, something like `InTotoObject`... but I dunno.

The kicker, though, is that the entire ValidationMixin class is four lines (define, for, if, and one function call), and only used in three classes, so maybe you should just put validate() in each class and save the reader from having to contemplate any subclassing or mix-ins or bouncing back and forth if they're reading the class or looking for a validate() method.

In fact, even class Signable only has two lines and is only inherited by two classes, so you could strip **all** class inheritance from in-toto and it would cost you a few of lines of code duplication but make the code more readable and less bouncy and entirely get rid of module `common.py`. (You might end up with about the same number of lines after removal of super calls and imports and such.)

Just a thought.

in_toto/models/metadata.py
awwad: Remember to make this an issue for later.

in_toto/models/common.py
```diff
@@ -15,140 +15,44 @@
...
     for method in inspect.getmembers(self, predicate=inspect.ismethod):
-        if method[0].startswith("_validate_"):
-          method[1]()
+      if method[0].startswith("_validate_"):
+        method[1]()
+
+
 
 @attr.s(repr=False, init=False)
 class Signable(ValidationMixin):
```
awwad, line 58: This doesn't directly relate to this PR, but I'm observing it at this code state and this PR does touch data representation, so:

Since `__repr__` is cryptographically important for class `Signable`, you should add a clear note to `Signable.__repr__()` that wards off minor tinkering. (Something like `# Please note that the returned string from this function is the content signed by ... so changes affect what data is signed, break backward compatibility with old data, and any example data`)

awwad (commented):
Looks good to me. Feedback consists of architectural thoughts and some minor edits.

JustinCappos:
On Thu, Nov 2, 2017 at 3:14 PM, Sebastien Awwad <notifications@github.com>
wrote:

> *@awwad* commented on this pull request.
>
> Looks good to me. Feedback consists of architectural thoughts and some
> minor edits.
> ------------------------------
>
> In in_toto/runlib.py
> <https://github.com/in-toto/in-toto/pull/142#discussion_r148342777>:
>
> > @@ -31,12 +31,15 @@
>
>  import in_toto.settings
>  from in_toto import log
> -from in_toto.models.link import (UNFINISHED_FILENAME_FORMAT, FILENAME_FORMAT)
> +from in_toto.models.link import (UNFINISHED_FILENAME_FORMAT, FILENAME_FORMAT,
> +    FILENAME_FORMAT_SHORT)
>
> Remember that when you from ... import a string, it is copied from the
> source module at import time. Thus, if FILENAME_FORMAT is changed in any
> module after this module is imported the first time, the value in this
> module and the value in in_toto.models.link (or any module that uses a
> normal import of in_toto.models.link) will not match. Testing is one
> context in which such a pattern might be used, as would be any kind of
> initial configuration setup at a higher level than this code. (This is true
> not just of strings, but of anything immutable or any variable you reassign
> later.)
>
> It doesn't necessarily mean you shouldn't do this, but it's worth keeping
> in mind. (I've been burned by this.)
>

This is one of the main reasons why we recommend just importing the module
and referring to objects / variables under there.

https://github.com/secure-systems-lab/code-style-guidelines#imports

in_toto/models/metadata.py
lukpueh: https://github.com/in-toto/in-toto/issues/144

in_toto/models/layout.py
lukpueh: I had the exact same thoughts, but concluded that the PR was already big enough. However, in the grand scheme of things we'll probably remove OOP altogether. But for now I think the inheritance cases are easy enough to grasp and the alternative of having duplicate code is not ideal either.

in_toto/models/layout.py
lukpueh: I think I'll leave it be for now. But I'll keep your comments regarding `attr` in the back of my mind.

in_toto/runlib.py
lukpueh: I'll fix this with https://github.com/in-toto/in-toto/issues/145

lukpueh:
Thanks for your comments, @awwad! I addressed some of them directly and ticketized others.

Issue #117, opened 2017-09-12
Revisit verbosity (stdout/stderr)


lukpueh:
In https://github.com/in-toto/in-toto/issues/6 we discussed and fixed many initial user feedback issues.
In the course of a layout tool user testing session, it was noted that when looking at stdout/stderr of the used tool (`in-toto-mock`) it was unclear if the command was executed properly or not.

Here are some general user feedback related items open for discussion and fixing:
- Make better use of verbosity levels
- Reduce the output if `-v` flag is not specified
- Make clear if a command failed or not
- Colorize output
- Do something about `subprocess32` warning (seems to confuse users)

SantiagoTorres:
> - Make clear if a command failed or not

imho, this goes against UNIX design principles. I'd suggest to not say anything if the command succeeds.

lukpueh:
I think the *rule of silence* is more flexible than that. What about three verbosity levels? e.g.:
(1) quiet: no output at all
(2) reasonable: a oneliner if the command succeeds, plus warnings and errors
(3) verbose: print everything (our current default)

And we could configure the levels e.g. 
- `(default)`, `-v`, `-vv`, or
- `-q`, `(default)`, `-v`

lukpueh:
Related issue:
The current default verbosity also inundates the unit test output (c.f. @vladimir-v-diaz's [comment](https://github.com/in-toto/in-toto/pull/163#issuecomment-355582253) on an unrelated PR). We should reduce it to a meaningful minimum.

lukpueh:
Related issue:
It would be good to review existing log statements for their appropriateness in content, type (info/warning/debug) and location (c.f. https://github.com/in-toto/in-toto/pull/167#discussion_r160725540).

PR #143, opened 2017-10-26
Add user settings module

### LINKED ISSUE #137: Provide alternatives to `settings.py`
Changing settings in the [settings file](https://github.com/in-toto/in-toto/blob/develop/in_toto/settings.py), which usually lives in some python site-packages directory, is cumbersome. 

We should provide alternative ways to configure these settings, like e.g. environment variables, rcfiles, and command line arguments (maybe all of them, in reverse order of precedence).

@SantiagoTorres  suggested to take a look at the [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) for inspiration. 

Currently there are three in-toto related settings, and three [securesystemslib settings](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/settings.py).

lukpueh:
The newly added module provides methods to read environment variables and rcfiles and to write them to the `in_toto.settings` module, i.e. override them (fix #137).

The PR also adds a call of the settings parsing and overriding method to in-toto's `__init__.py`, which should be the first thing that's executed when using in_toto. 
Any code that accesses `in_toto.settings` thereafter receives the values defined by the user.

More information about how the user sets the settings can be found in the [header docstring of `user_settings.py`](https://github.com/in-toto/in-toto/blob/4390aeae19fd9e8cee69790acd76510298c812ec/in_toto/user_settings.py#L15-L48).


in_toto/__init__.py
```diff
@@ -0,0 +1,3 @@
+from in_toto.user_settings import set_settings
```
vladimir-v-diaz: Our code style guidelines advises against importing items from a module: https://github.com/secure-systems-lab/code-style-guidelines#imports

What's wrong with the following?

``` Python
import in_toto.user_settings

in_toto.user_settings_setting.set_settings()
```

in_toto/__init__.py
```diff
@@ -0,0 +1,3 @@
...
+# Override settings in settings.py
```
vladimir-v-diaz: Can you put some blank lines between import statements and the body of the module?

in_toto/__init__.py
```diff
@@ -0,0 +1,3 @@
...
+set_settings()
```
vladimir-v-diaz: It seems wrong to perform actions when a project or module is imported.

Also, are you careful to catch exceptions or errors when your project is imported?  It would be strange to get an error about a missing or corrupted config file when attempting to import the project.

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Provides methods to read environment variables and rcfiles and to write
+  them to the in_toto.settings module (override them).
+
+  Notes:
+  - Variable values are converted to lists if they contain colons
```
vladimir-v-diaz: Should this be "environment values", so as not to confuse it with a Python variable?

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+  - Variable names are converted to upper case
```
vladimir-v-diaz: To be precise, do you mean that in-toto settings are always upper case?

Can you be a bit more precise here, perhaps providing an example?  Are `in-toto.settings.SOME_SETTING` settings always upper case regardless of the way that environments variables are written (i.e., upper case, lower case, or mixed case) ?

in_toto/user_settings.py
vladimir-v-diaz: In a comment, you should explain why this convention is used.   It is arbitrarily chosen, is it to ease parsing, to avoid parsing issues, or because multiple values are set this way with environment variables?

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+  - Environment variables need to be prefixed with "IN_TOTO_". The prefix is
+    removed when the variable is written to the settings module,
+    e.g. the envvar...
+    IN_TOTO_ARTIFACT_EXCLUDES="*.link:.gitignore"
+
+    becomes the setting ...
+    in_toto.settings.ARTIFACT_EXCLUDES = ["*.link", ".gitignore"]
+
+  - RCfile values follow the format expected by Python's builtin
+    ConfigParser. The format requires at least one section, sections however
```
vladimir-v-diaz: Do you need to revise this comment?  "The format requires at least one section, sections however are ignored"

Section headers are ignored?  Except that one section header is required?

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+"""
+import os
+import ConfigParser
+import in_toto.settings as SETTINGS_MODULE
+
+ENV_PREFIX = "IN_TOTO_"
+USER = os.path.expanduser("~")
+
+RC_PATHS = [
+  os.path.join("/etc", "in_toto", "config"),
```
vladimir-v-diaz: Can you list these in the order of precedence used in the docstring?  It helps readability, and a reviewer to verify that an RC path isn't missing.

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+  os.path.join("/etc", "in_totorc"),
+  os.path.join(USER,  ".config", "in_toto", "config"),
+  os.path.join(USER, ".config", "in_toto"),
+  os.path.join(USER, ".in_toto", "config"),
+  os.path.join(USER, ".in_totorc"),
+  ".in_totorc"
```
vladimir-v-diaz: Why is this ".in_totorc" written on a separate line, unlike the others?

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+
+"""
+import os
+import ConfigParser
+import in_toto.settings as SETTINGS_MODULE
+
+ENV_PREFIX = "IN_TOTO_"
+USER = os.path.expanduser("~")
+
+RC_PATHS = [
```
vladimir-v-diaz: How come this list doesn't include the `.in_totorc` RC file in the current working directory?  See line 40.

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+        len(name) > len(ENV_PREFIX)):
+      stripped_name = name[len(ENV_PREFIX):]
+
+      env_dict[stripped_name] = _colon_split(value)
+
+  return env_dict
+
+
+def get_rc():
+  """Load RCfiles from the usual places and return a dictionary of
```
vladimir-v-diaz: The usual places?  Can you be more explicit here?  How about, "the paths specified in `RC_PATHS`?

Preferably, you should assume that a developer/reader is only looking at this docstring and is unaware of the module's other functions and docstrings.

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+    if (name.startswith(ENV_PREFIX) and
+        len(name) > len(ENV_PREFIX)):
+      stripped_name = name[len(ENV_PREFIX):]
+
+      env_dict[stripped_name] = _colon_split(value)
+
+  return env_dict
+
+
+def get_rc():
```
vladimir-v-diaz, line 117: How come this function's docstring isn't formatted according to our code style guidelines?  Where is the section on exceptions, return values, purpose, etc.? 

The other functions in this module are also missing proper docstrings.

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+
+def set_settings():
+  """Read environment variables and RCfiles and write them to the
+  settings module.
+  """
+  user_settings = get_env()
+  user_settings.update(get_rc())
+
+  for name, value in user_settings.iteritems():
+    setattr(SETTINGS_MODULE, name, value)
```
vladimir-v-diaz: Is it possible to populate `in-toto.settings` with unexpected variables or values?  For example, `in-toto.settings.[]`  How do you expect the module to catch these exceptions or fail in these instances?

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+
+<Started>
+  Oct 26, 2017
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Test user settings module.
```
vladimir-v-diaz: I find the following comment easier to determine which module is being tested:

> Test in-toto.user_settings.py.

vladimir-v-diaz (commented):
My comments for the main modules, I still need to review `test_user_settings.py`

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+
+  for section in config.sections():
+    for name, value in config.items(section):
+      name = name.upper()
+      rc_dict[name] = _colon_split(value)
+
+  return rc_dict
+
+
+def set_settings():
```
vladimir-v-diaz, line 177: Is it wise to first backup the current settings before attempting to overwrite them?

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+    with open(".in_totorc", "w") as fobj:
+      fobj.write(
+          "[settings]\n"
+          "artifact_basepath=r/c/file\n"
+          "ARTIFACT_EXCLUDES=r:c:file\n"
+          "new_rc_setting=new rc setting")
+
+    os.environ["IN_TOTO_ARTIFACT_EXCLUDES"] = "e:n:v"
+    os.environ["IN_TOTO_artifact_basepath"] = "e/n/v"
+    os.environ["IN_TOTO_NEW_ENV_SETTING"] = "new env setting"
```
vladimir-v-diaz: Is this manually setting some environment variables?  Is it possible to temporarily set some environment variables and verify that environment variables are properly loaded?

Perhaps testing actual environment variables on the system is unnecessary, but something to consider.

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+
+import os
+import sys
+import unittest
+import logging
+import shutil
+import tempfile
+import in_toto.settings
+from in_toto.user_settings import set_settings, get_env, get_rc
```
vladimir-v-diaz: I guess it's okay to import individual items from a module (in this unit test), but you should get into the habit of using absolute import statements.

```
in-toto.user_settings.get_rc()
```

```
set_settings()
```

The former is more readable in my opinion, since it shows where the function comes from.  The latter can be confused for some function in this module.

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+    self.working_dir = os.getcwd()
+    self.test_dir = os.path.realpath(tempfile.mkdtemp())
+    os.chdir(self.test_dir)
+
+    # Backup settings
+    self.settings_backup = {}
+    for key in dir(in_toto.settings):
+      self.settings_backup[key] = getattr(in_toto.settings, key)
+
+    # Create an rc file and set some environment variables
```
vladimir-v-diaz: Another thing you can do is to have a pre-generated `.in_totorc` that you can load in the unit tests.  This approach might be faster, easier for others to inspect and review, etc.

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+
+    os.environ["IN_TOTO_ARTIFACT_EXCLUDES"] = "e:n:v"
+    os.environ["IN_TOTO_artifact_basepath"] = "e/n/v"
+    os.environ["IN_TOTO_NEW_ENV_SETTING"] = "new env setting"
+    os.environ["NOT_IN_TOTO_NOTHING"] = "nothing"
+
+
+  @classmethod
+  def tearDownClass(self):
+    os.chdir(self.working_dir)
```
vladimir-v-diaz, line 50: If you use pre-generated config files, another thing it free you from is the need to change directories and clean up temp files.  The unit tests would only need to load config files.

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+  def test_get_rc(self):
+    rc_dict = get_rc()
+    self.assertEquals(rc_dict["ARTIFACT_BASEPATH"], "r/c/file")
+    self.assertListEqual(rc_dict["ARTIFACT_EXCLUDES"], ["r", "c", "file"])
+    self.assertEquals(rc_dict["NEW_RC_SETTING"], "new rc setting")
+
+
+  def test_get_env(self):
```
vladimir-v-diaz: Should you test what happens if an empty environment is loaded?  That is, there are no in-toto environment variables set, and no in-toto config files can be loaded?

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Create an rc file and set some environment variables
+    with open(".in_totorc", "w") as fobj:
+      fobj.write(
+          "[settings]\n"
+          "artifact_basepath=r/c/file\n"
+          "ARTIFACT_EXCLUDES=r:c:file\n"
+          "new_rc_setting=new rc setting")
+
+    os.environ["IN_TOTO_ARTIFACT_EXCLUDES"] = "e:n:v"
```
vladimir-v-diaz: Can you test a value that contains special characters like `$`, `%`, `..`, etc.?  What other interesting things?  Python identifiers?  Function calls?

in_toto/user_settings.py
vladimir-v-diaz: Can you verify that someone can't cause unintended lines?  Is it possible to populate `in-toto.settings.py` with malicious input like: `"OS.REMOVE(my_files)".lower()`

Perhaps you don't have to worry about malicious input in config files and environment variables, but something to ponder.

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+  def test_get_env(self):
+    env_dict = get_env()
+    self.assertEquals(env_dict["ARTIFACT_BASEPATH"], "e/n/v")
+    self.assertListEqual(env_dict["ARTIFACT_EXCLUDES"], ["e", "n", "v"])
+
+    self.assertEquals(env_dict["NEW_ENV_SETTING"], "new env setting")
+    self.assertFalse("NOT_IN_TOTO_NOTHING" in env_dict)
+
+
+  def test_set_settings(self):
```
vladimir-v-diaz: Is it possible to test the import behavior that follows from the modified `__init__()`?  These test conditions appear to test environment and RCfiles indirectly.

test/test_user_settings.py
```diff
@@ -0,0 +1,96 @@
...
+    set_settings()
+    # RCfile settings have precedence over env settings
+    self.assertEquals(in_toto.settings.ARTIFACT_BASEPATH, "r/c/file")
+    self.assertListEqual(in_toto.settings.ARTIFACT_EXCLUDES, ["r", "c", "file"])
+    self.assertEquals(in_toto.settings.NEW_RC_SETTING, "new rc setting")
+
+    # Settings that are only in env are still used even if they are new
+    self.assertEquals(in_toto.settings.NEW_ENV_SETTING, "new env setting")
```
vladimir-v-diaz: Should you have a test condition that verifies overwritten settings?  Suppose `in-toto.settings.MY_SETTING` is already set, and you attempt to call `set_settings()`.  Is `MY_SETTING` overwritten, as expected?

in_toto/user_settings.py
```diff
@@ -0,0 +1,132 @@
...
+        ~/.config/in_toto/config
+        /etc/in_totorc
+        /etc/in_toto/config
+    2. Environment Variables
+    3. Settings defined in `settings.py`
+
+"""
+import os
+import ConfigParser
+import in_toto.settings as SETTINGS_MODULE
```
vladimir-v-diaz: Is there a reason why you use `import X as FOO`?  In this case,
`SETTINGS_MODULE` is only one character longer, and less readable than `in_toto.settings`.

vladimir-v-diaz (commented):

in_toto/__init__.py
lukpueh: Good point, I actually had a similar feeling.  It's probably better to only monkey patch the settings module when using a command line tool. What do you think?

in_toto/user_settings.py
lukpueh: Values form envvars or rcfiles. I will rephrase.

in_toto/user_settings.py
lukpueh: This list does include `.in_totorc`. See your [comment below](https://github.com/in-toto/in-toto/pull/143#discussion_r148293274).

in_toto/user_settings.py
lukpueh: I guess I rather reverse the list in the docstring. If I reversed here, I'd have to do a `RC_PATHS.reverse()` below where I use the list, or iterate backwards, which seems odd too.

in_toto/user_settings.py
lukpueh: Each line is an item in the list. `.in_totorc` just looks different because it is a relative path and therefore does not need to be `os.path.join`ed.

in_toto/__init__.py
vladimir-v-diaz: Yes, it makes more sense to me to have the command-line tool read the config file(s) and environment variables.  The command tool help message or its documentation should explain how the settings can be modified via the environment variables and config files.  It should also list the supported settings.

in_toto/user_settings.py
lukpueh: I thought that we only use this convention for API functions. But you are right, it says
*"for every function [..] that will be called from other modules"*.

I'll change this.

in_toto/user_settings.py
vladimir-v-diaz: Ahh, true.

Although it's overkill, I wonder if `os.path.join(os.getcwd(), ".in_totorc")` is more readable.

lukpueh: I think that's an overkill. :)

test/test_user_settings.py
lukpueh: I don't evaluate the strings. So it should be fine.

vladimir-v-diaz: Is is possible for someone to attempt a `import in-toto.settings` and get errors because of unexpected entries?

test/test_user_settings.py
lukpueh: I thought about it, but that would actually be a test for `os.environ`.

in_toto/user_settings.py
lukpueh: It does not seem possible, since I am assigning module variables with strings.

However, there isn't really a need to allow everything, so it might be better to make a whitelist, i.e. only allow settings for which there is a default in `settings.py`.

vladimir-v-diaz: Using a whitelist is the more secure approach in my opinion.

in_toto/user_settings.py
lukpueh: Not unless we use the backup for something (e.g. rollback on error). But considering that we use a [whitelist](https://github.com/in-toto/in-toto/pull/143#discussion_r148387751), which reduces the unexpected error likelihood; and only use this function in [cli tools](https://github.com/in-toto/in-toto/pull/143#discussion_r148377987), where I'd rather exit on error, I think we don't need to backup.

test/test_user_settings.py
lukpueh: Why not. Will add.

test/test_user_settings.py
lukpueh: `ARTIFACT_BASEPATH` and `ARTIFACT_EXCLUDES` have defaults in `settings.py`. So the overriding is already tested.

vladimir-v-diaz: I see.  Should the test and/or modules indicate that `ARTIFACT_BASEPATH` and `ARTIFACT_EXCLUDES` are actual in-toto settings?  If you happen to remove these settings in the future, do these tests conditions become invalid (in terms of failing to test for the overwriting behavior)?  It's good to note assumptions in the comments for these sort of situations

Also, assuming that `user_settings.py` overwrites settings in `in-toto.settings.py`, wouldn't a more accurate test (1) verify that the new setting is an actual supported setting in `in-toto.settings.py` (2) note the value of the setting and (3) verify that the value in (2) is overwritten with the new value?

lukpueh:
@vladimir-v-diaz, thanks for you comments. I tried to address them accordingly. Let me know what you think.

Update: It seems that  2ea0e5d broke it, because `ConfigParser` converts names to lower case. I'll fix this on Monday. You can review the rest if you want though.

test/test_user_settings.py
lukpueh: I added sanitization for the available settings where used in 4072cb3.

lukpueh:
I fixed the remaining issues regarding case sensitivity that broke the unittests last week, added some testcases and fixed missing settings sanitization where they are used. Mind giving it another pass, @vladimir-v-diaz?

README.md
```diff
@@ -105,6 +105,16 @@ in-toto-verify --layout <layout path>
...
 ```
 
+
+#### Settings
+in-toto provides a settings module as a place to store defaults that can
+be used throughout the package. When using in-toto from the command line,
+these settings can be configured using environment variables or RCfiles.
+Take a look at [`in_toto.settings`](https://github.com/in-toto/in-toto/blob/develop/in_toto/settings.py)
+to learn more about the available defaults and at
+[`in_toto.user_settings`](https://github.com/in-toto/in-toto/blob/develop/in_toto/user_settings.py)
```
vladimir-v-diaz: I get a 404 error with this link.

vladimir-v-diaz: I'm not so sure about the  documentation of this section.  Is the command-line tool intended to be run by a variety of users?  If so, wouldn't linking to a Python module might be inappropriate for these users?

lukpueh: Re 404: should have used relative path (`user_settings.py` isn't in the main branch yet).
Re inappropriate: Agreed, will change.

in_toto/runlib.py
```diff
@@ -119,7 +120,7 @@ def record_artifacts_as_dict(artifacts):
             the link command.
 
   <Exceptions>
-    None.
+    SettingsError if ARTIFACT_BASE_PATH or ARTIFACT_EXCLUDES can't be used
```
vladimir-v-diaz: It might be better if this listed the exception name: `in-toto.exceptions.SettingsError`, since the caller will have to specify the full name to correctly catch the exception.

in_toto/runlib.py
```diff
@@ -137,18 +138,29 @@ def record_artifacts_as_dict(artifacts):
     original_cwd = os.getcwd()
     try:
       os.chdir(in_toto.settings.ARTIFACT_BASE_PATH)
-    except OSError as e:
-      raise OSError("Review your ARTIFACT_BASE_PATH setting - {}".format(e))
+    except Exception as e:
```
vladimir-v-diaz, line 145: Is there a reason why you switched from catching `OSError` to catching `Exception`?  According to the [Python documentation](https://docs.python.org/3/library/os.html#module-os), all functions in the os modules raise `OSError` 

> in the case of invalid or inaccessible file names and paths, or other arguments that have the correct type, but are not accepted by the operating system.

in_toto/runlib.py
```diff
@@ -137,18 +138,29 @@ def record_artifacts_as_dict(artifacts):
...
+      raise in_toto.exceptions.SettingsError(
+          "Review your ARTIFACT_BASE_PATH setting '{}' - {}".format(
+          in_toto.settings.ARTIFACT_BASE_PATH, e))
 
   # Normalize passed paths
   norm_artifacts = []
   for path in artifacts:
     norm_artifacts.append(os.path.normpath(path))
 
+  # If ARTIFACT_EXCLUDES is set it must be a list of strings or an empty
```
vladimir-v-diaz: Or an empty what?

in_toto/runlib.py
```diff
@@ -137,18 +138,29 @@ def record_artifacts_as_dict(artifacts):
     original_cwd = os.getcwd()
     try:
       os.chdir(in_toto.settings.ARTIFACT_BASE_PATH)
-    except OSError as e:
-      raise OSError("Review your ARTIFACT_BASE_PATH setting - {}".format(e))
+    except Exception as e:
+      raise in_toto.exceptions.SettingsError(
+          "Review your ARTIFACT_BASE_PATH setting '{}' - {}".format(
```
vladimir-v-diaz, line 147: Consider providing more information on what could have gone wrong.  For example, an invalid path.  On the other hand, appending the exception, as you have done, would make it clear what is wrong with `ARTIFACT_BASE_PATH`.

in_toto/runlib.py
```diff
@@ -137,18 +138,29 @@ def record_artifacts_as_dict(artifacts):
...
 
   # Normalize passed paths
   norm_artifacts = []
   for path in artifacts:
     norm_artifacts.append(os.path.normpath(path))
 
+  # If ARTIFACT_EXCLUDES is set it must be a list of strings or an empty
+  # TODO: Change NAMES_SCHEMA to something more semantically accurate
+  if (in_toto.settings.ARTIFACT_EXCLUDES and not
+      securesystemslib.formats.NAMES_SCHEMA.matches(
```
vladimir-v-diaz: Maybe PATH_SCHEMA or PATHS_SCHEMA is a bit more accurate?  I am not certain, but ARTIFACT_EXCLUDES appears to be glob patterns?

in_toto/user_settings.py
```diff
@@ -12,59 +12,47 @@
...
+# i.e. the same setting in `/etc/in_toto/config` and `.in_totorc` (cwd) uses
+# the latter
 RC_PATHS = [
   os.path.join("/etc", "in_toto", "config"),
   os.path.join("/etc", "in_totorc"),
-  os.path.join(USER,  ".config", "in_toto", "config"),
-  os.path.join(USER, ".config", "in_toto"),
-  os.path.join(USER, ".in_toto", "config"),
-  os.path.join(USER, ".in_totorc"),
+  os.path.join(USER_PATH,  ".config", "in_toto", "config"),
```
vladimir-v-diaz: I guess home directory would be another good name for this.

in_toto/settings.py
```diff
@@ -12,36 +12,35 @@
...
-# FIXME: Add as command line argument or to config file, e.g .in-toto-ignore
+
+# See docstring of `in-toto.record_artifacts_as_dict` for how this is used
 ARTIFACT_EXCLUDES = ["*.link*", ".git", "*.pyc", "*~"]
 
 # Used as base path for --materials and --products arguments when running
 # in-toto-run/in-toto-record
-# FIXME: This is likely to become a command line argument
 # FIXME: Do we want different base paths for materials and products?
 ARTIFACT_BASE_PATH = None
```
vladimir-v-diaz: Should ARTIFACT_BASE_PATH be set to the user's home directory, by default?  Or current working directory?

in_toto/user_settings.py
```diff
@@ -76,18 +64,47 @@ def _colon_split(value):
...
+    Parse environment for variables with prefix `ENV_PREFIX` and return
+    a dict of key-value pairs.
+
+    The prefix `ENV_PREFIX` is stripped from the keys in the returned dict.
+
+    Values that contain colons (:) are split at the postion of the colons and
+    converted into a list.
+
+
+    Example:
```
vladimir-v-diaz, line 78: 👍

in_toto/user_settings.py
```diff
@@ -76,18 +64,47 @@ def _colon_split(value):
...
+      "ARTIFACT_BASE_PATH": "/home/user/project"
+      "ARTIFACT_EXCLUDES": ["*.link", ".gitignore"]
+    }
+    ```
+
+  <Exceptions>
+    None.
+
+  <Side Effects>
+    Calls function to read files from disk.
```
vladimir-v-diaz, line 99: For a while I used to list similar entries in Side Effects, but others have pointed out that this isn't really a side effect.

in_toto/user_settings.py
```diff
@@ -98,35 +115,93 @@ def get_env():
...
-  - RCfile format as expected by builtin ConfigParser
-  - The format dictates that there is at least one section,
-    however we are only interested in the values and ignore the sections
-  - Values that contain colons are converted into a list
-  - Variable names are converted to upper case
+  <Exceptions>
+    None.
+
+  <Side Effects>
+    Calls function to read files from disk.
```
vladimir-v-diaz, line 157: Another instance where someone might argue that this isn't really a side effect.

in_toto/user_settings.py
```diff
@@ -98,35 +115,93 @@ def get_env():
...
-    setattr(SETTINGS_MODULE, name, value)
\ No newline at end of file
+  # If the user has specified one of the settings whitelisted in
+  # IN_TOTO_SETTINGS per envvar or rcfile, override the item in `settings.py`
+  for setting in IN_TOTO_SETTINGS:
+    user_setting = user_settings.get(setting)
+    if user_setting:
+      log.info("Using setting {0}={1}".format(
+          setting, user_setting))
+      setattr(in_toto.settings, setting, user_setting)
```
vladimir-v-diaz: Consider adding an else-statement that logs a debug message for settings that are ignored/invalid.

test/test_user_settings.py
```diff
@@ -34,63 +32,72 @@ class TestUserSettings(unittest.TestCase):
...
-      self.settings_backup[key] = getattr(in_toto.settings, key)
-
-    # Create an rc file and set some environment variables
-    with open(".in_totorc", "w") as fobj:
-      fobj.write(
-          "[settings]\n"
-          "artifact_basepath=r/c/file\n"
-          "ARTIFACT_EXCLUDES=r:c:file\n"
-          "new_rc_setting=new rc setting")
+    # We use `demo_files` as test dir because it has an `.in_totorc`, which
```
vladimir-v-diaz: I don't understand this comment.  Is `demo_files` supposed to be `rc_test`?

vladimir-v-diaz (commented):
Here is the requested review of your follow-on commits.  I had minor comments.  Feel free to merge when you feel the pull request is ready.

in_toto/runlib.py
lukpueh: It could also be a TypeError. I figured I just catch all Exceptions and re-raise as `SettingsError`.

in_toto/runlib.py
lukpueh: I think adding the exception message is good enough. I guess we can always enhance this later, e.g. when taking care of #117.

in_toto/user_settings.py
lukpueh: I remember having this discussion and also that there were uncertainties about some things and whether they would qualify as side effect. But I am pretty sure that there was no doubt about I/O.

test/test_user_settings.py
lukpueh: Oh, I forgot to update this comment. I first added the rcfile to demo_files, as you suggested, but then discovered that this would break other tests that copied everything from demo_files and therefor also loaded and used the test rcfile.

in_toto/runlib.py
vladimir-v-diaz: You can also explicitly catch all the exceptions that you're aware of.  For the unexpected exceptions, it might be better to let the exception pass through. 
```
try:
    os.chdir(...)
except (OSError, TypeError) as e:
    ...
```

A problem I have encountered in the past is that the code would sometimes catch weird things like `NameError` and the exception message would be totally wrong.  You append the original exception, so I guess it would be easy to debug / catch these cases.

in_toto/user_settings.py
lukpueh: Adding that else-statement would actually log the default settings and not the ignored/invalid settings, which might be more interesting anyway. I will add this.

README.md
```diff
@@ -122,7 +122,7 @@ Also, setting values that contain colons are parsed as list.
 
 ##### Available Settings
 
-`ARTIFACT_EXCLUDES` Specifies a list of glob patterns that can be used to
+`ARTIFACT_EXCLUDE_PATTERNS` Specifies a list of glob patterns that can be used to
```
vladimir-v-diaz: :1st_place_medal:

vladimir-v-diaz:
LBTM (Looks Brilliant To Me)

:)

PR #153, opened 2017-11-27
Fix user settings leaking into other tests

lukpueh:
On some platforms the user settings created in test_user_settings leak into other tests and break them, as they expect the defaults from settings.py and a clean environment.

This change modifies the user_settings testcase to restore/reset the test settings in the testcase's teardown method.

### DIFF
--- a/test/test_user_settings.py
+++ b/test/test_user_settings.py
@@ -33,6 +33,11 @@ class TestUserSettings(unittest.TestCase):
   def setUpClass(self):
     self.working_dir = os.getcwd()
 
+    # Backup settings to restore them in `tearDownClass`
+    self.settings_backup = {}
+    for key in dir(in_toto.settings):
+      self.settings_backup[key] = getattr(in_toto.settings, key)
+
     # We use `rc_test` as test dir because it has an `.in_totorc`, which
     # is loaded (from CWD) in `user_settings.set_settings` related tests
     self.test_dir = os.path.join(os.path.dirname(__file__), "rc_test")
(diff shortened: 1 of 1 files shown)


test/test_user_settings.py
```diff
@@ -44,11 +49,21 @@ def setUpClass(self):
...
 
+    # Other unittests might depend on defaults:
+    # Restore monkey patched settings ...
+    for key, val in self.settings_backup.iteritems():
+      setattr(in_toto.settings, key, val)
+
+    # ... and delete test environment variables
+    del os.environ["IN_TOTO_ARTIFACT_EXCLUDE_PATTERNS"]
+    del os.environ["IN_TOTO_ARTIFACT_BASE_PATH"]
+    del os.environ["IN_TOTO_not_whitelisted"]
```
SantiagoTorres, line 64: The rest of this variable is lowercase. is this intentional?

test/test_user_settings.py
```diff
@@ -44,11 +49,21 @@ def setUpClass(self):
...
+    del os.environ["NOT_PARSED"]
+
```
SantiagoTorres, line 66: This is good by me, I wonder if we can move this to a decorator for the tests. Something like

```python
@set_environ(IN_TOTO_ARTIFACT_BASE_PATH="thisthing")
def test_this_and_that(self):
  ...
```

It's probably worth doing down the line, but I'd move it to a ticket.

SantiagoTorres (approved):

test/test_user_settings.py
lukpueh: Oh I wasn't aware of that decorator. I can do that right away.

test/test_user_settings.py
lukpueh: Yes, to acknowledge case sensitivity in `user_settings.get_env`.

test/test_user_settings.py
SantiagoTorres: no, we need to write it

lukpueh: oh ok :) then I guess the current approach is fine.

test/test_user_settings.py
SantiagoTorres: ah, ok ok

SantiagoTorres:
LGTM. Merging

Issue #18, opened 2016-10-18
Metadata pretty print


lukpueh:
Layout metadata and link metadata is stored in canonical json format. It would be nice to have a way of showing metadata in a logically structured and human friendly way as well.
- order fields logically instead of alphabetically
- strip keys and signatures

JustinCappos:
I'll comment more (after a shoulder surfing example), but I think there are
many subtle things that can be done in the display to make this very clear.

lukpueh:
See #155  for WIP implementation.

lukpueh:
Also see #161 for adding colorizing output.

ashutosh887:
@adityasaky I saw #546
Please let me know if I can work on this

lukpueh:
Thanks, @ashutosh887! @ShubhikaBhardwaj, seems to still be working on it. I suggest you pick a different issue.

ashutosh887:
Sure @lukpueh 
Thanks

iferminm:
Greetings,

Is anyone working on this? I'd like to contribute but there are not so many "good first issue"s. Please let me know

ashutosh887:
Hi @iferminm 

@lukpueh told me that @ShubhikaBharadwaj is still working on this :)
I was planning to

lukpueh:
> Hi @iferminm
> 
> @lukpueh told me that @ShubhikaBharadwaj is still working on this :) I was planning to

This no longer seems to be the case (see https://github.com/in-toto/in-toto/pull/546#issuecomment-1543456450). Feel free to take a stab at it!

hojoungjang:
Hello! I took a stab at this issue on this [PR#631](https://github.com/in-toto/in-toto/pull/631). I would love to see if anyone can review and work together to get this change merged.

trishankatdatadog:
Not that pretty printing isn't nice, but isn't this problem orthogonal to what python-in-toto should do? Aren't there better tools like `jq` that you could use in composition? I say this because I had asked for printing text in colour in securesystemslib which we later regretted because it added unnecessary third-party dependencies just to run python-tuf.

hojoungjang:
Got it, so should I explore using `jq` to achieve similar behavior implemented in my changes and close the PR for now?

trishankatdatadog:
> the

I'd say so. Does anyone else object?

adityasaky:
I think I broadly agree, this may not be something we want to do in the implementation itself.

[...]

PR #155, opened 2017-12-02
Metadata pretty printing

mike-k0:
This was a stab at issue #18.

There are two models that were analyzed for the purposes of understanding what it means to "print" them in a readable format that takes into accountability human readability.

The Layout model serves to display most of the aggregate information of the supply chain, whereas the Link model is much more isolated.

Using the in-toto demo as an example, this is what was decided were the primary concerns for a Link object being printed:
- First and foremost, to notify the end user what type of object we have (e.g Link)
- To identify the name of the "Link" (e.g "clone")
- To identify the "inputs" (e.g the directory, and the input materials)
- To identify the "outputs" (e.g the products - what was produced and captured by the Link)
- To provide a summary:
-- The summary includes up to four details (if applicable): files that were (1) created, (2) deleted, (3) modified, and (4) unchanged.

An example screenshot of the Link metadata is as follows:

Link metadata (example 1):
link_meta

In the above example, a link (name: clone), was run with the command "git clone https://github.com/in-toto/demo-project.git". Since there were no actual input files (materials), the summary indicated that there was now a new file created, "demo-project/foo.py".

Link metadata (example 2):
link_3

In the above example, a link (name: update-version) included updating the version of demo-project/foo.py. Although the input file and the output file names are the same, the summary still shows that there was a modification made to the actual input file, or material. Additionally, the command row indicates that there hadn't been any commands run.

Layout metadata:

The "Layout" metadata was responsible for capturing a broader scope of the entire supply chain, so it was important to be thorough in the information it printed:

- Information that let the user know the type of the object (Layout)
- The expiration of the Layout
- The readme information for the Layout
- The "keyids" of the Layout (the authorized parties, or functionary keys are included here)

- The step data (which can be broken down further)
-- The step name (and threshold, or minimum amount of functionaries required)
-- The public keys associated with that particular step in the supply chain
-- The expected command (cmd) to be ran in the supply chain
-- The expected inputs (materials)
-- The expected outputs (products)

- The inspection data (which can be broken down further)
-- The name of the inspection
-- The command that was run (for validation)
-- The expected inputs (materials) via inspection data
-- The expected outputs (products) via inspection data

Below is a sample printing of the entire Layout metadata, but of course it'd be possible to display step or inspection data separately as they currently have their own separate "display" method implementations.
lay2

- [x] The code follows the Code Style Guidelines


lukpueh:
@mikek2, thanks for working on this task! It would be great if we were able to parametrize what parts of a given metadata file are displayed and in what order, e.g. by passing format placeholders to your `display` function. What I had in mind initially was something like [git's pretty formats](https://git-scm.com/docs/pretty-formats).

Also feel free to comment on the original ticket or here in your PR so that we can discuss your thoughts.

mike-k0:
@lukpueh I like the idea of being able to exercise control via parameterization, because it'll allow for more concise printing of information, and I can see why that'd be a nice feature to have. Additionally, git's pretty formats seems like a useful reference point that I hadn't initially thought about when I was taking a stab at the issue.

I was also planning on attempting on trying to get a PR out for #117 (verbosity), but afterwards, your idea is something I would like to revisit.

in_toto/models/layout.py
```diff
@@ -120,6 +120,27 @@ def read(data):
 
     return Layout(**data)
 
+  def display(self):
+    """
+    <Purpose>
+      Represents the Layout class by returning its fields in a readable string
+    <Returns>
+      Layout fields: type, expiration, readme, keyids, steps, inspections
```
lukpueh: Please re-phrase. This contains a string and not individual fields.

in_toto/models/layout.py
```diff
@@ -120,6 +120,27 @@ def read(data):
 
     return Layout(**data)
 
+  def display(self):
+    """
+    <Purpose>
+      Represents the Layout class by returning its fields in a readable string
```
lukpueh: Could you add a very condensed form of the rationale for why you choose to represent the layout exactly this way?

in_toto/models/layout.py
```diff
@@ -120,6 +120,27 @@ def read(data):
...
+    if not self.readme:
+      readme = "readme: None"
+    else:
+      readme = self.readme
+    keyid = "keyids: {}".format("\n\t  ".join(self.keys))
+
+    steps = "step data:\n {}".format(self.steps)
+    inspections = "inspect data:\n {}".format(self.inspect)
+
+    return "  {}\n  {}\n  {}\n  {}\n  {}\n  {}".format(types, expiration, readme,
```
lukpueh: IMHO it would be more readable and easier to change if you defined a format string that already looks somewhat like what you will return, e.g.:

```python
return (
  "object: {}\n"
  "expiration: {}\n"
  "{}\n"
  "keyids: {}\n"
  "..."
  .format(self._type, self.exipration, self.readme, keyids .... 
)
```
What do you think?

in_toto/models/layout.py
```diff
@@ -248,6 +269,38 @@ def __init__(self, **kwargs):
...
+      expected_command = "  expected cmd: None"
+    else:
+      expected_command = "  expected cmd: {}".format(" ".join(self.expected_command))
+
+    expected_materials = "  expected input (materials):\n\t"
+    if not self.expected_materials:
+      expected_materials += "   None"
+    else:
+      for value in self.expected_materials:
+        expected_materials += "   {}\n\t".format(" ".join(value))
```
lukpueh: Might be worth writing a little `_display_artifact_rule()` helper function since you have copied this block 4 times.

in_toto/models/layout.py
```diff
@@ -248,6 +269,38 @@ def __init__(self, **kwargs):
 
     self.validate()
 
+  def __repr__(self):
```
lukpueh: Please don't overwrite `Step.__repr__` (or `Inspection.__repr__`). Just give them a display method like you did for Layout and Link.

in_toto/util.py
```diff
@@ -136,3 +135,29 @@ def prompt_generate_and_write_rsa_keypair(filepath, bits):
...
+    A tuple value which contains dict information: (created, deleted, modified, unchanged)
+  """
+  dict1_keys = set(dict1.keys())
+  dict2_keys = set(dict2.keys())
+
+  intersections = dict1_keys.intersection(dict2_keys)
+  created = dict1_keys - dict2_keys
+  deleted = dict2_keys - dict1_keys
+  modified = {o : (dict1[o], dict2[o]) for o in intersections if dict1[o] != dict2[o]}
+  unchanged = set(o for o in intersections if dict1[o] == dict2[o])
```
lukpueh: Please replace list comprehensions c.f. [code style guidelines](https://github.com/secure-systems-lab/code-style-guidelines#no-lambda-functions-or-lisp-esque-code) and use meaningful variable names.

in_toto/verifylib.py
```diff
@@ -1315,6 +1314,10 @@ def in_toto_verify(layout, layout_key_dict):
   # container (Metablock) that also carries the signatures
   layout = layout.signed
 
+  log.info("Begin printing layout metadata")
+  log.info(layout.display())
+  log.info("End printing layout metadata")
```
lukpueh: Please don't call you function here or in `runlib`.

lukpueh (changes requested):
Thanks for the contribution, I think it looks good overall. 

Please consider my comments for the entire PR (most of the comments apply in various places).

Please also write testcases that call your display functions and that cover all added lines.

in_toto/verifylib.py
mike-k0: I've removed these function calls from both `verifylib` and `runlib` in the latest commits.

in_toto/util.py
mike-k0: Thank you for this. I've cleaned up the syntax and comments for the compare_dictionaries function.

in_toto/models/layout.py
mike-k0: Thank you for this. The Step and Inspection classes both have a display method like Link's and Layout instead of repr methods.

in_toto/models/layout.py
mike-k0: Yup, this seems spot on. I've updated the code to look more like this.

in_toto/models/layout.py
mike-k0: Sorry about that. Comments have been edited.

in_toto/models/layout.py
mike-k0: Yup, done. Let me know if anything else is missing!

in_toto/models/layout.py
mike-k0: I created a `check_string()` function in util.py to address some issues of copy/pasting the same code over and over again.

in_toto/models/layout.py
```diff
@@ -248,6 +282,38 @@ def __init__(self, **kwargs):
...
+            expected_materials, 
+            expected_products
+    """
+    pubkeys = "{}".format("\n  ".join(self.pubkeys))
+    expected_command = "{}".format(" ".join(self.expected_command))
+    expected_materials = ""
+    expected_products = ""
+
+    for value in self.expected_materials:
+      expected_materials += " {}\n\t".format(" ".join(value))
```
lukpueh: In my last review suggested to write a helper function to display artifact rules.

in_toto/models/layout.py
```diff
@@ -248,6 +282,38 @@ def __init__(self, **kwargs):
...
+    for value in self.expected_products:
+      expected_products += " {}\n\t".format(" ".join(value))
+
+    return (
+      "\n    step name: {} (threshold: {})\n"
+      "    pubkeys: {}\n " 
+      "   expected cmd: {}\n " 
+      "   expected input (materials): \n\t{}\n " 
+      "   expected output (products): \n\t{} \n "
+      .format(self.name, self.threshold, pubkeys, in_toto.util.check_string(expected_command), \
```
lukpueh: You don't need to escape the newline inside parentheses.

in_toto/models/link.py
```diff
@@ -106,6 +107,40 @@ def read(data):
...
+    command = in_toto.util.check_string("{}".format(" ".join(self.command)))
+    materials = in_toto.util.check_string("{}".format(" ".join(self.materials)))
+    products = in_toto.util.check_string("{}".format(" ".join(self.products)))
+    summary = ""
+    new, deleted, modified, no_change = in_toto.util.compare_dictionaries(self.products, self.materials)
+
+    summary += "\n\tcreated: {}".format(in_toto.util.check_string(" ".join(new)))
+    summary += "\n\tdeleted: {}".format(in_toto.util.check_string(" ".join(deleted)))
+    summary += "\n\tmodified: {}".format(in_toto.util.check_string(" ".join(modified)))
+    summary += "\n\tunchanged: {}".format(in_toto.util.check_string(" ".join(no_change)))
```
lukpueh: You could use one formatted string here instead, but I guess that's a matter of taste.

in_toto/util.py
```diff
@@ -136,3 +136,49 @@ def prompt_generate_and_write_rsa_keypair(filepath, bits):
...
+  return created, deleted, modified, unchanged
+
+def check_string(string_arg):
+  """ Checks strings that are empty and defaults them to 'None' string"""
+  if not isinstance(string_arg, str):
+    return string_arg
+  else:
+    if not string_arg:
+      string_arg = "None"
+  return string_arg
```
lukpueh: I'm not sure if a `"None"` string is really more meaningful than just displaying an empty string. So I guess we don't really need the `check_string` function.

lukpueh (commented):
Looks good overall. Please consider some minor comments.

in_toto/util.py
mike-k0: All mentions of this function (`check_string`) have been removed in the latest commit!

in_toto/models/link.py
mike-k0: Yup, doable. This change is reflected in the latest commit!

in_toto/models/layout.py
mike-k0: Oops, I must have added this erroneously at some point in time. Removed in the latest changes!

in_toto/models/layout.py
mike-k0: Ooh, I missed this somehow... but a private method (`_display_list`) now takes care of it in the latest changes!

lukpueh:
Closing due to inactivity (referenced in related issue #18).

PR #156, opened 2017-12-04
Replace canonicaljson with securesystemslib.encode_canonical

### LINKED ISSUE #144: Uniform JSON string format
https://github.com/in-toto/in-toto/blob/e2f3f8b64c71e1cbe581051154c58cb949a6d579/in_toto/models/metadata.py#L48-L59

See related comment https://github.com/in-toto/in-toto/pull/142#discussion_r146687332:

> @SantiagoTorres, what do you think? I'd say, either we make it all (pretty printed) canonical JSON or none of it. The only really important thing is that the JSON string representation of the object held in the signed attribute (Link or Layout) is deterministic when we create or verify signatures.

lukpueh:
This PR removes the external dependency of `canonicaljson` and instead uses `securesystemslib.encode_canonical` to sign and verify the signable part of in_toto metadata.

It also adds a new property to the Signable class which returns that canonical json string (used to be returned by `repr`), to be more explicit about what is used for signing/verification.

Metablock's and Signable's `repr` methods now return a non-canonical indented JSON string. Former is used to write metadata to disk. Metadata does not have to be displayed (represented) or stored in canonical json, this is only required to generated and verify signatures.

The commit also updates the signatures of the metadata test files and adds a basic unittest for Signable.__repr__.

Closes #144


in_toto/models/metadata.py
```diff
@@ -46,17 +45,12 @@ def __init__(self, **kwargs):
...
-    # "canonical json" and the container is just "json".
-    # Is this really what we want?
-    return '{{"signed": {}, "signatures": {}}}'.format(self.signed,
-        canonicaljson.encode_canonical_json(self.signatures))
+    """Returns an indented JSON string of the metadata object. """
+    return json.dumps(
+        {
+          "signatures": self.signatures,
+          "signed": attr.asdict(self.signed)
+        }, indent=1, separators=(",", ": "), sort_keys=True)
```
SantiagoTorres, line 53: Awesome to see this `FIXME` go away

SantiagoTorres (approved):
LGTM

SantiagoTorres:
Just for the sake of completeness. We may want to use the display method from #155 as our repr now that this is taken care of?

Issue #11, opened 2016-10-17
Properly duplicate standard streams


lukpueh:
Currently we provide a command line flag to `toto-run` to either don't touch stdout and stderr, which is required if `toto-run` wraps interactive commands like e.g. `vi`, or to redirect and store them as byproducts to the link metadata.
Ideally we would like to duplicate the streams to allow both.
- Using a pipe won't work because processes like vi will complain
- Wrapping stdout/sterr in Python won't work because the suprocess will only take the filedescriptor and then uses it natively
- Reading from /dev/(stdout|stderr), /dev/tty is *NIX specific

baloyan:
I have implemented it in C.
Still I am trying to do it in Python, regarding to the Python documentation it is not straightforward (or almost impossible) problem as of today.

baloyan:
OK Lukas, I think I did it in Python. 
The solution is when we have error it write error message in ERROR.log file if we do not have error it means command runs correct so we can write STDOUT in OUT.log file.
Is it OK ?

lukpueh:
The current version of the specification does not mention `stderr` anymore, instead it lists [`stdin` as common byproduct](https://github.com/in-toto/docs/blame/69c29b41f4c1fe064b5d77d61d4b268fc171bcaa/in-toto-spec.md#L912-L917). 

When fixing this issue, let's also investigate how to redirect standard input.

PR #160, opened 2017-12-10
Duplicating standard output streams

ng1355:
Hi! This is a first attempt at issue #11. 

What I've done is adapted the current approach to log output to additionally echo it to the user's terminal. This approach uses subprocess.Popen instead of subprocess.call so toto-run can still do stuff (like echo output) while still redirecting output to files. These files are then read from during the subprocess' execution and outputted to the terminal, essentially redirecting the output while logging. 

As is, this program is completely functional with non-interactive programs on both UNIX and Windows systems, using the the "git clone" from the demo:
image
image

Certain interactive programs work well _enough_, such as Vim, where interactivity isn't hampered, and the output is logged properly. This still causes Vim to complain as it detects its output isn't connected to a tty:
image
image
Unfortunately, this doesn't seem to work on Windows, though - Vim just hangs. :( I'm not sure what causes this bug yet, might be how Vim interacts with the Windows console since I know Vim forces line buffering on unix. Curses applications like nano also behave weirdly and aren't very usable (control characters don't work). 

Subprocesses that use stdio's default buffering also have their own issues, such as in the following program I built that asks the user for a number and echoes it back. Without making an explicit call to fflush() in the program itself, toto-run blocks on input before the prompt is flushed to the terminal: 
image
image
An explicit flush in subprocess itself "fixes" this: 
image

As an aside, I'm not entirely sure if it's even possible to portably simulate a connected tty device. I wonder if it might be worth considering checking the user's platform and interacting with the underlying terminal accordingly. While this contribution is portable, it still seems to have platform-related bugs, like with Vim hanging on Windows. 

*I understand stderr is not required anymore, though I thought to implement it anyway to keep the byproducts unchanged. If desired I can strip out all references to stderr though.

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -273,17 +273,15 @@ def execute_link(link_cmd_args, record_streams):
   if record_streams:
 
     # Windows doesn't allow reopening TemporaryFile()s
-    stdout_file = tempfile.mkstemp()
-    stderr_file = tempfile.mkstemp()
-    stdout_name = stdout_file[1]
-    stderr_name = stderr_file[1]
+    stdout_fd, stdout_name = tempfile.mkstemp()
+    stderr_fd, stderr_name = tempfile.mkstemp()
 
     # Using the same fd for the subprocess and the readers doesn't work
     with \
-        io.open(stdout_name, 'wb') as writer, \
         io.open(stdout_name, 'rb', 1) as reader, \
-        io.open(stderr_name, 'wb') as err_write, \
-        io.open(stderr_name, 'rb', 1) as err_read:
+        io.open(stderr_name, 'rb', 1) as err_read, \
+        os.fdopen(stdout_fd, 'wb') as writer, \
+        os.fdopen(stderr_fd, 'wb') as err_write: 
 
       # Popen allows us to do work (read from the file) during the subproc.
       # Disabling buffering here (or during the io.open) does not to fix
@@ -310,11 +308,8 @@ def execute_link(link_cmd_args, record_streams):
       stderr_str = err_read.read()
       return_value = proc.poll()
 
-      os.close(stdout_file[0])
-      os.close(stderr_file[0])
-      
-      os.remove(stdout_name)
-      os.remove(stderr_name) 
+    os.remove(stdout_name)
+    os.remove(stderr_name) 
       
   else:
     return_value = subprocess.call(link_cmd_args)
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -278,62 +278,53 @@ def execute_link(link_cmd_args, record_streams):
     stdout_name = stdout_file[1]
     stderr_name = stderr_file[1]
 
-    # Open the files as r+b cause the subprocess to block indefinitely
+    # Using the same fd for the subprocess and the readers doesn't work
     with \
-    io.open(stdout_name, 'wb') as writer, \
-    io.open(stdout_name, 'rb', 1) as reader, \
-    io.open(stderr_name, 'wb') as err_write, \
-    io.open(stderr_name, 'rb', 1) as err_read:
+        io.open(stdout_name, 'wb') as writer, \
+        io.open(stdout_name, 'rb', 1) as reader, \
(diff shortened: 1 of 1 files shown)


in_toto/runlib.py
```diff
@@ -262,20 +263,68 @@ def execute_link(link_cmd_args, record_streams):
...
+    stderr_file = tempfile.mkstemp()
+    stdout_name = stdout_file[1]
+    stderr_name = stderr_file[1]
+
+    # Open the files as r+b cause the subprocess to block indefinitely
+    with \
+    io.open(stdout_name, 'wb') as writer, \
+    io.open(stdout_name, 'rb', 1) as reader, \
+    io.open(stderr_name, 'wb') as err_write, \
+    io.open(stderr_name, 'rb', 1) as err_read:
```
lukpueh: Note that you end up with 3 different filedescriptors per file here: `stdout_file[0] != writer.fileno() != reader.fileno()`.

I suppose you could use the filedescriptor that you get when calling `mkstemp()` for your reader or writer using `fdopen` instead of `open` then you wouldn't have to call `os.close` explicitly below (the `with` statement takes care of closing the files).

lukpueh: Please use double indentation (4 spaces) for line continuation and don't break the line after the `with`, i.e.:
```python
with io.open(stdout_name, 'wb') as writer, \
    io.open(stdout_name, 'rb', 1) as reader,
    ...
 # Popen allows ...
```

in_toto/runlib.py
```diff
@@ -262,20 +263,68 @@ def execute_link(link_cmd_args, record_streams):
...
+        # the buffering issue mentioned above. 
+        proc = subprocess.Popen((link_cmd_args), stdout=writer, \
+             stderr=err_write) 
+
+        # echo proc's logged stdout to our stdout
+        while proc.poll() is None:
+            sys.stdout.write(reader.read())
+            sys.stderr.write(err_read.read())
+            sys.stdout.flush()
+            sys.stderr.flush()
```
lukpueh: Could you add a unit test that evaluates `proc.poll() is None` to True?

in_toto/runlib.py
```diff
@@ -262,20 +263,68 @@ def execute_link(link_cmd_args, record_streams):
...
+
+    # Potentially for portability if the OS doesn't delete them on closing? 
+    # Raises error on OSX/Win
+    '''
+    try:
+        os.remove(stdout_name)
+        os.remove(stdout_name) 
+    except OSError:
+        pass
+    '''
```
lukpueh: Are you sure about this? For me removing the file works (on OSX). Maybe you get an exception because you are trying to remove the same file twice?

If think a good pattern to use `mkstemp` is:

```python
fd, path = tempfile.mkstemp()
try:
  with os.fdopen(fd, "w") as f:
    # do stuff with f
finally:
  os.remove(path)
```

lukpueh: Please use 2 spaces per indentation level. While you are at it, would you mind fixing the indentation in the else clause below?

lukpueh (changes requested):
Thanks for the great contribution!

in_toto/runlib.py
ng1355: Fixed the indentation. 

Duplicating the file descriptors is intentional. I tried giving the subprocess the same fd as the main process and, while the output is logged to the file, the .read() in the main proc returns empty strings on each call. 

I can inline the temp creation into the with statement, but it looks kinda ugly and requires me to save the filenames for os.remove:
image

I did change two of the io.opens() to fdopens however so that the original fds can be closed automatically. I'm wondering if its better to keep them as fdopens() so that its clear that file descriptors are being opened, or to use io.open() again to keep the file interface in this block consistent.

in_toto/runlib.py
ng1355: Yeah this turned out to just be a typo I overlooked, os.remove now raises no errors. Thanks for catching it.

lukpueh:
There's an updated edition of this PR in #252. Closing here. Thanks for preparing this, @ng1355!

PR #161, opened 2017-12-11
Adding verbosity options

mike-k0:
**Issue #**: #117 

The discussions outlined in issue 117 (revisiting verbosity) are widely under the consensus that the current verbosity of the in-toto system can sometimes be unclear, so I'd like to believe that these changes are in the step of the right direction.

The changes that have been made, for starters:

There are now "three" distinct verbose modes (from most verbose to least verbose):
* A "debug" mode
  * Specified by "-d" or "--debug"
  * Currently supported by in_toto_run, in_toto_verify, and in_toto_record
  * Log level is set to DEBUG (Will also print INFO, WARNING, ERROR, & CRITICAL)

* A "verbose" mode
  * Specified by "-v" or "--verbose"
  * Currently supported by in_toto_run, in_toto_verify, in_toto_record, and in_toto_sign
  * Log level is set to INFO (Will also print WARNING, ERROR, & CRITICAL)

* A "default" mode
  * If both debug mode and verbose mode are not detected, this is the mode that is set by default (no explicit parameters)
  * Currently supported by in_toto_run, in_toto_verify, in_toto_record, and in_toto_sign
  * Log level is set to WARNING (Will also print ERROR & CRITICAL)

* The subprocess32 warning has been slightly re-worded (for clarity).

* Currently, although the debug log messages are supported, there aren't many uses of debug messages in the actual in-toto source code. Now that there is a solid groundwork for a debug mode, it'll be much easier to add debug messages inside of in-toto for the future.

These are screenshots of running the in-toto demo with and without verbose output:

Default mode:
not-verbose

Verbose mode: (-v):
verbose-msg

* Additionally, as seen above, all log messages are now prefixed by their appropriate log level name. Previously, this was not the case for "INFO" messages printed to console, and now that it has been added, all output should remain consistent.

* An additional enhancement that has been made is colorized output. This makes it much easier to know what messages are being shown to the end user (green = INFO, yellow = WARNING, red = ERROR/CRITICAL).
  * This optional feature can be enabled by specifying "-c", or "--color"
  * Currently, only operating systems that support ANSI escape codes support the colorization of output. This means this feature is omitted when a Windows system tries to use the --color or -c parameter.

Examples of colorized versus non-colorized output:

Default mode (not colorized):
default-nocolor

Default mode: (colorized):
defaultcolor

Verbose mode (not colorized):
verbose-msg

Verbose mode (colorized):
verbose-color

These changes are only the beginning to a much greater scope of issues, but I believe this is a good starting point for establishing and enhancing clarity for the end user.

- [x] The code follows the Code Style Guidelines


in_toto/runlib.py
```diff
@@ -47,13 +48,15 @@
   try:
     import subprocess32 as subprocess
   except ImportError:
-    log.warn("POSIX users (Linux, BSD, etc.) are strongly encouraged to"
-        " install and use the much more recent subprocess32")
+    in_toto.settings.COLOR = in_toto.util.detect_colorization(sys.argv)
```
lukpueh: Please don't use `sys.arg` in the library module. If a user wants colorized output s/he should set it to True in the command line tool or by directly modifying the `in_toto.settings.COLOR`.

in_toto/runlib.py
```diff
@@ -47,13 +48,15 @@
...
+    log.warn("POSIX users (Linux, BSD, etc.) are encouraged to"
+        " install the new subprocess32 module"
+        " instead of the version provided in python "
+        + str(sys.version_info[0]) + "." + str(sys.version_info[1]))
```
lukpueh: I don't think this change is necessary. As a matter of fact, we'll probably remove this all too prominent warning once we have Python3 support.

in_toto/in_toto_record.py
```diff
@@ -156,9 +166,11 @@ def main():
...
   args = parser.parse_args()
 
-  # Turn on all the `log.info()` in the library
-  if args.verbose:
-    log.logging.getLogger().setLevel(log.logging.INFO)
+  # Distinguish log levels via parameterized input
+  log.logging.getLogger().setLevel(args.loglevel)
+
+  # Set color settings
+  in_toto.settings.COLOR = args.colorize
```
lukpueh: The problem with this approach is that it ignores the defaults in `in_toto.settings`. Can you change both to only override settings if the options are passed?

in_toto/in_toto_record.py
```diff
@@ -145,8 +147,16 @@ def main():
...
+  in_toto_args.add_argument("-v", "--verbose", dest="loglevel",
+      help="Verbose execution.", default=log.logging.WARNING, const=log.logging.INFO,
+      action="store_const")
+
+  in_toto_args.add_argument("-d", "--debug", dest="loglevel",
+      help="Debug statement execution.", default=log.logging.WARNING, const=log.logging.DEBUG,
+      action="store_const")
+
+  in_toto_args.add_argument("-c", "--color", dest="colorize",
+      help="Colorizes output.", default=False, action="store_true")
```
lukpueh: I'd rather not have too many new command line options. What do you think of providing multiple `-v` options?
i.e.: `-v` for `INFO` and  `-vv` for `DEBUG`

in_toto/util.py
```diff
@@ -136,3 +138,27 @@ def prompt_generate_and_write_rsa_keypair(filepath, bits):
   generate_and_write_rsa_keypair"""
   password = prompt_password()
   generate_and_write_rsa_keypair(filepath, bits, password)
+
+def color_code(msg, lvl, override=False):
```
lukpueh: Maybe `color_encode` would be a better name? Also `level` seems short enough for not needing to shorten it.

lukpueh: I think the module `in_toto.log` is a better place for this function. What do you think?

in_toto/util.py
```diff
@@ -136,3 +138,27 @@ def prompt_generate_and_write_rsa_keypair(filepath, bits):
...
+  if not override:
+    #FIXME: Add support for Windows (no support for ANSI escape codes)
+    if platform.system().lower() is "windows" or not in_toto.settings.COLOR:
+      return msg
+  level = log.logging.getLevelName(lvl)
+  levelGenerator = {  "CRITICAL" : "\x1b[31m", # red
+                      "ERROR" : "\x1b[31m", #  red
+                      "WARNING" : "\x1b[33m", # yellow
+                      "INFO" : "\x1b[32m",  # green
+                      "DEBUG" : "\x1b[35m" } # magenta
```
lukpueh: You could create constants for these colors.

in_toto/util.py
```diff
@@ -136,3 +138,27 @@ def prompt_generate_and_write_rsa_keypair(filepath, bits):
...
+    if level in key:
+      msg = "{}{}\x1b[0m".format(value, msg)
+  return msg
+
+def detect_colorization(arguments):
+  colorization = False
+  for value in arguments:
+    if value == "-c" or value == "--color":
+      colorization = True
+  return colorization
```
lukpueh: This function is not necessary.

in_toto/log.py
```diff
@@ -20,22 +20,26 @@
...
 
 def error(msg):
   """Prints unexpected errors """
-  logging.error("ERROR: {}".format(msg))
+  logging.error("ERROR: {}".format(in_toto.util.color_code(msg, 40)))
 
 def pass_verification(msg):
   """Prints passing verification routines. """
-  logging.critical("PASSING: {}".format(msg))
+  logging.critical("PASSING: {}".format(in_toto.util.color_code(msg, 50)))
```
lukpueh: I think `pass_verification` should have a more positively connotated color. 
Also, could you pass the constant names instead of the hardcoded values as `log.level` parameter? Or maybe it makes even more sense to just pass the color, defined in a constant, e.g. `log.GREEN`, ...

lukpueh (changes requested):
Thanks for your PR. Please consider my comments...

in_toto/util.py
mike-k0: This function has been removed entirely in the latest commit!

in_toto/util.py
mike-k0: Yup, done.

in_toto/util.py
mike-k0: Yup, renamed the function to `color_encode` and it's now sitting in `in_toto.log` as a function (and removed from util).

in_toto/runlib.py
mike-k0: Alright, the change has been reverted in the latest commit.

in_toto/runlib.py
mike-k0: Sorry about that. That's been removed now.

in_toto/log.py
mike-k0: Yup, changed the message for `pass_verification` to be GREEN instead of RED.

in_toto/in_toto_record.py
mike-k0: I've set the `default` parameter from the `argsparse` library to whatever is specified in `LOG_LEVEL` (from the settings module). Is that good enough? 

I also have an additional check afterwards, but I think it might be unnecessary because the default specified in `argsparse` might do the job. If you could look over that, i'd appreciate it.

in_toto/in_toto_record.py
mike-k0: Personally, I think it's more intuitive to leave --debug for debugging, and that's the implementation i've seen with some other software. 

I'm not entirely sure who's right here, but in my personal opinion, I think `-vv` isn't as intuitive.

in_toto/in_toto_record.py
lukpueh: If you don't mind just wrap it in an if.

mike-k0: I was asking because the coveralls API seems to drop coverage when certain statements are written (I don't know if this was intentional or not).

Additionally: are unit tests required for what's in the log module? The reason I ask is because it seems to be saying the debug function in the log module is missing in terms of "coverage".

lukpueh:
Closing due to inactivity. Verbosity options were extended in #183 (referenced in related issue #18 for colorizing output).

Issue #145, opened 2017-11-07
Revisit imports


lukpueh:
In  https://github.com/in-toto/in-toto/pull/142#discussion_r148342777 @awwad points out that we should be more careful in the way we import modules or parts of modules:

> Remember that when you from ... import a string, it is copied from the source module at import time. Thus, if FILENAME_FORMAT is changed in any module after this module is imported the first time, the value in this module and the value in in_toto.models.link (or any module that uses a normal import of in_toto.models.link) will not match. Testing is one context in which such a pattern might be used, as would be any kind of initial configuration setup at a higher level than this code. (This is true not just of strings, but of anything immutable or any variable you reassign later.)
It doesn't necessarily mean you shouldn't do this, but it's worth keeping in mind. (I've been burned by this.)

His suggestion also aligns with the [code guidelines](https://github.com/secure-systems-lab/code-style-guidelines#imports).

Although it does not look like there are any issue with this, we should probably change for the sake of conformity.

PR #163, opened 2018-01-03
Add pylint and bandit to tox and adapt code to pass linting

lukpueh:
Adds a lint test environment to tox.ini to run bandit and pylint using securesystemslib's generic (pylintrc)[https://github.com/secure-systems-lab/sample-documents/blob/master/pylintrc].
We separate static from dynamic testing, because former needs to run only once (not for each tested Python version).

The PR also fixes various issues pointed out by running the linters.

requirements**:

### DIFF
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -99,6 +99,11 @@ class Layout(Signable):
 
     self.validate()
 
+  @property
+  def type_(self):
+    """Getter for protected _type attribute. Trailing underscore used by
+    convention (pep8) to avoid conflict with Python's type keyword. """
+    return self._type
 
   @staticmethod
   def read(data):
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -126,9 +126,10 @@ class Layout(Signable):
           "Invalid _type value for layout (Should be 'layout')")
 
   def _validate_expires(self):
-    """Private method to verify the expiration field."""
+    """Private method to verify if the expiration field has the right format
+    and can be parsed."""
     try:
-      date = parse(self.expires)
+      parse(self.expires)
       securesystemslib.formats.ISO8601_DATETIME_SCHEMA.check_match(
           self.expires)
     except Exception as e:
@@ -150,7 +151,7 @@ class Layout(Signable):
 
     securesystemslib.formats.KEYDICT_SCHEMA.check_match(self.keys)
 
-    for keyid, key in six.iteritems(self.keys):
+    for junk, key in six.iteritems(self.keys):
       securesystemslib.formats.PUBLIC_KEY_SCHEMA.check_match(key)
 
   def _validate_steps_and_inspections(self):
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -29,7 +29,6 @@
       represents a hook that is run at verification
 """
 
-import json
 import attr
 import six
 import shlex
@@ -38,7 +37,6 @@ from datetime import datetime
 from dateutil.relativedelta import relativedelta
 from dateutil.parser import parse
 
-from in_toto.models.link import (UNFINISHED_FILENAME_FORMAT, FILENAME_FORMAT)
 from in_toto.models.common import Signable, ValidationMixin
 import in_toto.artifact_rules
 import in_toto.exceptions
--- a/in_toto/models/link.py
+++ b/in_toto/models/link.py
@@ -19,7 +19,6 @@
 """
 
 import attr
-import json
 import securesystemslib.formats
 from in_toto.models.common import Signable
 
--- a/in_toto/models/link.py
+++ b/in_toto/models/link.py
@@ -99,6 +99,11 @@ class Link(Signable):
 
     self.validate()
 
+  @property
+  def type_(self):
+    """Getter for protected _type attribute. Trailing underscore used by
+    convention (pep8) to avoid conflict with Python's type keyword. """
+    return self._type
 
   @staticmethod
   def read(data):
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -266,7 +266,8 @@ def execute_link(link_cmd_args, record_streams):
   # btw: we ignore them in the layout anyway
 
   if record_streams:
-    # Use SpooledTemporaryFile if we expect very large outputs
+    # We should consider using SpooledTemporaryFile if we expect very large
+    # outputs in the future
     stdout_file = tempfile.TemporaryFile()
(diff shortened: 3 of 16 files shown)


vladimir-v-diaz:
GitHub indicates that coverage/coveralls failed.  Do you know why?  I took a quick glance but the coveralls output wasn't too helpful.

vladimir-v-diaz:
The Travis output is inundated with test statements.  You should consider printing only test failures to make it easier to review the output and/or any test failures.

lukpueh:
I think it is because the number of covered lines decreased in some files (I removed obsolete imports), which makes not covered lines weigh more and therefor decrease the overall coverage.

vladimir-v-diaz:
What happens if there is a Pylint or Bandit failure?  Does the travis build fail?  Is the pull request blocked if there are any linting failures?

vladimir-v-diaz:
> I think it is because the number of covered lines decreased in some files (I removed obsolete imports), which makes not covered lines weigh more and therefor decrease the overall coverage.

I wonder why GitHub flags the pull request as a failure because of the coveralls check.

in_toto/models/layout.py
```diff
@@ -101,6 +99,11 @@ def __init__(self, **kwargs):
 
     self.validate()
 
+  @property
+  def type_(self):
+    """Getter for protected _type attribute. Trailing underscore used by
+    convention (pep8) to avoid conflict with Python's type keyword. """
```
vladimir-v-diaz, line 105: More importantly, the trailing underscore convention is recommended by our code style guidelines.
https://github.com/secure-systems-lab/code-style-guidelines#function-and-method-arguments

lukpueh:
*Re verbosity:* I added a comment in a related issue and will fix it there https://github.com/in-toto/in-toto/issues/117#issuecomment-355584168.
*Re fail:* If pylint or bandit fail, they return a non-zero value which aborts tox `commands` execution and returns a non-zero exit code, which in turn makes the travis build fail.

vladimir-v-diaz:
> Re fail: If pylint or bandit fail, they return a non-zero value which aborts tox commands execution and returns a non-zero exit code, which in turn makes the travis build fail.

Another thing you should verify is whether the output of a pylint/bandit failure is printed to the console.  If there is a pylint or bandit failure, is it clear from reading Travis's output?

in_toto/runlib.py
```diff
@@ -263,7 +266,7 @@ def execute_link(link_cmd_args, record_streams):
   # btw: we ignore them in the layout anyway
 
   if record_streams:
-    # XXX: Use SpooledTemporaryFile if we expect very large outputs
+    # Use SpooledTemporaryFile if we expect very large outputs
```
vladimir-v-diaz: Is this still a TODO?  It's not clear from the comment, now that `XXX:` is removed.

lukpueh:
*Re GitHub flag:* There are coveralls settings to configure thresholds, for minimum coverage and maximum coverage decrease.  I see for TUF you set 97% for the former. What to you think is a sensible default for the latter?

lukpueh:
> Another thing you should verify is whether the output of a pylint/bandit failure is printed to the console. If there is a pylint or bandit failure, is it clear from reading Travis's output?

Good catch, I just saw that travis actually didn't run the linters. This is fixed with 3a28ac9.

vladimir-v-diaz:
I'm not sure what's a good number for the maximum coverage decrease.  It is ideal that the majority of pull requests contain some unit testing, and that the minimum test coverage not fall below a certain threshold.

Is setting a minimum threshold below your current coverage sufficient?  Say, 99% minimum test coverage when your current coverage is 100%.  In this way, you can allow a large maximum coverage decrease to avoid any failed coveralls check, and still allow some pull requests that are not fully tested (perhaps it cannot be fully tested because it depends on another pending pull request).

in_toto/models/link.py
```diff
@@ -19,7 +19,6 @@
 """
 
 import attr
-import json
 import securesystemslib.formats
 from in_toto.models.common import Signable
```
vladimir-v-diaz, line 23: I see that the "from foo import bar" convention is being used.  Is there a reason why?  This convention is not advised by our guidelines

in_toto/runlib.py
lukpueh: Not really a todo, just a comment to consider this alternative if we run into problems.

in_toto/models/link.py
lukpueh: This is not related to this PR. But acknowledged in https://github.com/in-toto/in-toto/issues/145.

in_toto/runlib.py
vladimir-v-diaz: "... to consider this alternative if we run into problems."

Maybe, "We should consider using SpooledTemporaryFile if we expect very large outputs in the future."

lukpueh:
Good thoughts. I set the min coverage threshold to 99% and the max decrease to 0.5%. Latter should ignore fails due to line number decrease. 

In any case, we do allow and have performed merges that fail the coverage/coveralls test.

tox.ini
```diff
@@ -4,9 +4,26 @@
...
+envlist = lint,py27
 
-[testenv]
+[testenv:lint]
+deps =
+    pylint
+    bandit
+
+commands =
+    # Run pylint, using securesystemslib's pylintrc configuration file
```
vladimir-v-diaz: pylintrc is provided by "secure systems lab," not "securesystemslib."

tox.ini
```diff
@@ -4,9 +4,26 @@
...
+    # https://github.com/secure-systems-lab/sample-documents/blob/master/pylintrc
+    pylint in_toto
+
+    # Run bandit, a security linter from OpenStack Security
+    # Note: We exclude tests that fail on importing or using the subprocess or
+    # subprocess32 modules, which are considered to have low severity security
+    # implications, but are required by in-toto. We do not skip subprocess
+    # calls with shell=True (B602).
+    bandit -r in_toto --skip B404,B603
```
vladimir-v-diaz: Can you link to B404 and B603, to make it easier for the reader to understand the implications of not testing them?

https://docs.openstack.org/bandit/latest/plugins/subprocess_without_shell_equals_true.html

https://docs.openstack.org/bandit/latest/plugins/subprocess_popen_with_shell_equals_true.html

https://docs.openstack.org/bandit/latest/blacklists/blacklist_imports.html#b404-import-subprocess

vladimir-v-diaz:
Okay, that's my final comments for the review.

lukpueh:
Thanks for the review, @vladimir-v-diaz!

Issue #126, opened 2017-10-09
Revisit exception taxonomy


lukpueh:
Currently in-toto raises builtin Python errors as well as custom exceptions defined in [`securesystemslib.exceptions`](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/exceptions.py) and [`in-toto.exceptions`](https://github.com/in-toto/in-toto/blob/develop/in_toto/exceptions.py).

It might be worth taking some time to grep through those `raise` statements and re-assess the suitability of each raised error and its message. 

See also @SantiagoTorres' suggestion in https://github.com/in-toto/in-toto/pull/123#discussion_r143279743.

lukpueh:
While we are at it, we should also fix generic `catch Exception` statements, c.f. @SantiagoTorres's comment in https://github.com/in-toto/in-toto/pull/167#discussion_r160724178.

lukpueh:
Related: Redact GPG errors https://github.com/in-toto/in-toto/pull/168#discussion_r161009073

lukpueh:
@aaaaalbert pointed out this issue in a recent `in-toto.gpg`-related CR (see https://github.com/in-toto/in-toto/pull/257#discussion_r261180946).

Since we are in the process of moving our gpg implementation to `securesystemslib`, we might want to address the issue over there (see https://github.com/secure-systems-lab/securesystemslib/pull/163).

Tiku24:
Hey I'm a new beginner can you suggest me how I start with this

SantiagoTorres:
Hi @Tiku24 !  This is somewhat of a recurring issue, so it may not be best to get started with. However, if you're interested in working around this area, what this issue mostly means is we need to make sure that the exceptions we are throwing are meaningful. The bigger picture is: we want to ensure that we want to eventually determine a minimum set of error messages. Let me know if this makes sense and if you'd still want to give this a try

Cheers!

Tiku24:
> Hi @Tiku24 ! This is somewhat of a recurring issue, so it may not be best to get started with. However, if you're interested in working around this area, what this issue mostly means is we need to make sure that the exceptions we are throwing are meaningful. The bigger picture is: we want to ensure that we want to eventually determine a minimum set of error messages. Let me know if this makes sense and if you'd still want to give this a try
> 
> Cheers!

Ok so from where I should starts  ??

Tiku24:
Ok so from where I should starts

adityasaky:
[...]

PR #167, opened 2018-01-09
Add gpg support to in-toto-record cli tool and library functions

lukpueh:
This PR adds gpg signing capabilities to the `in-toto-record` command line tool and its corresponding library functions.
*Note: in-toto-record can be used to create link metadata for multipart steps*

Furthermore this PR makes the following enhancements in in-toto-record:
- Declutter the command line interface code (77cac9e5b9c6346ef49fc370d51840a97a2c6b18)
- Make in-toto-record's subcommands `start` and `stop` the first arguments. Shared and arguments specific to the subcommands come afterwards. Former is realized with a shared [`argparse` parent parser](https://docs.python.org/2.7/library/argparse.html#parents) (fa227d6e6b772f9209bb4ca5c6147893366c341d)
- Remove custom cli tool help message and improve auto help message with [`argparse` metavars](https://docs.python.org/2.7/library/argparse.html#metavar) (20ddbfdc7f8fc51521039a2abb40b048c86c9f28)
- Update README (7ad6d551261b1a1f822573e831e9f2d400e69c67)
- Refactor unit tests (5ced4b8bac49372ed07f64012a54c9c2e71f7abd)

requirements**:

### DIFF
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
 """
 <Program Name>
   in_toto_record.py
@@ -14,32 +13,41 @@
   See LICENSE for licensing information.
 
 <Purpose>
-  Provides a command line interface to start and stop in-toto link metadata
-  recording.
+  Provides a command line interface to create a link metadata file in multiple
+  steps. The commands are:
 
   start
-    Takes a step name, a functionary's signing key and optional
-    material paths.
     Creates a temporary link file containing the file hashes of the passed
-    materials and signs it with the functionary's key under
-    .<step name>.link-unfinished
+    materials and signs it with the passed functionary's key under
+    .<step name>.<keyid>.link-unfinished
 
   stop
-    Takes a step name, a functionary's signing key and optional
-    product paths.
-    Expects a .<step name>.link-unfinished in the current directory signed by
-    the functionary's signing key, adds the file hashes of the passed products,
-    updates the signature and renames the file  .<step name>.link-unfinished
-    to <step name>.link
-
+    Expects a .<step name>.<keyid>.link-unfinished in the current directory
+    signed by the passed functionary's key, adds the file hashes of the passed
+    products, updates the signature and renames the file
+    .<step name>.<keyid>.link-unfinished to <step name>.<keyid>.link
 
   The implementation of the tasks can be found in runlib.
 
-  Example Usage
+<Example Usage>
+  Create link file signed with specified 'key' stored on disk and record all
+  files in current working directory as materials and products.
+  Any files in the current working directory that you edit between running
+  the commands will have different hashes in their corresponding material and
+  product entries of the resulting link file 'edit-files.<keyid>.link'.
+
   ```
-  in-toto-record --step-name edit-files start --materials . --key bob
-  # Edit files manually ...
-  in-toto-record --step-name edit-files stop --products . --key bob
+  in-toto-record start --step-name edit-files -key /path/to/key --materials .
+  in-toto-record stop --step-name edit-files -key /path/to/key --products .
+  ```
+
+  # Create link file signed with the default gpg key and record a file named
+  # 'foo' as material and product.
+  # If you edit foo between running the commands the recorded hashes in the
+  # resulting link file 'edit-foo.<keyid>.link' will differ.
+  ```
+  in-toto-record start --step-name edit-foo --gpg --materials path/to/foo
+  in-toto-record stop --step-name edit-foo --gpg --products path/to/foo
   ```
 
 """
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -108,10 +108,10 @@ def main():
 
   try:
     if args.command == "start":
-      in_toto.runlib.in_toto_record_start(args.step_name, key, args.materials)
+      in_toto.runlib.in_toto_record_start(args.step_name, args.materials, key)
 
     elif args.command == "stop": # pragma: no branch
-      in_toto.runlib.in_toto_record_stop(args.step_name, key, args.products)
+      in_toto.runlib.in_toto_record_stop(args.step_name, args.products, key)
 
     # Else is caught by argparser
 
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -56,10 +56,9 @@ def main():
   parser = argparse.ArgumentParser(
       description="Starts or stops link metadata recording")
 
+  # The subparsers inherit the arguments from the parent parser
+  parent_parser = argparse.ArgumentParser(add_help=False)
   subparsers = parser.add_subparsers(dest="command")
-
-  subparser_start = subparsers.add_parser('start')
-  subparser_stop = subparsers.add_parser('stop')
 
   # Whitespace padding to align with program name
   lpad = (len(parser.prog) + 1) * " "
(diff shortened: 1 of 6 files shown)


in_toto/in_toto_record.py
```diff
@@ -14,166 +13,141 @@
...
   # Override defaults in settings.py with environment variables and RCfiles
   in_toto.user_settings.set_settings()
 
+  # If `--gpg` was set without argument it has the value `True` and
+  # we will try to sign with the default key
+  gpg_use_default = (args.gpg == True)
+
+  # Otherwise we interpret it as actual keyid
+  gpg_keyid = None
+  if args.gpg != True:
```
SantiagoTorres: should this be
```python
if not gpg_use_default and args.gpg:
  ...
```
or so?

in_toto/in_toto_record.py
```diff
@@ -14,166 +13,141 @@
   See LICENSE for licensing information.
 
 <Purpose>
-  Provides a command line interface to start and stop in-toto link metadata
-  recording.
+  Provides a command line interface to create a link metadata file in multiple
```
SantiagoTorres: What do you think of changing "multiple" to "two"?

in_toto/in_toto_record.py
```diff
@@ -14,166 +13,141 @@
...
+  gpg_use_default = (args.gpg == True)
+
+  # Otherwise we interpret it as actual keyid
+  gpg_keyid = None
+  if args.gpg != True:
+    gpg_keyid = args.gpg
+
   # We load the key here because it might prompt the user for a password in
   # case the key is encrypted. Something that should not happen in the library.
+  key = None
```
SantiagoTorres, line 127: AFAIR, we need a ward that key is not used along gpg right? Is this being done by argparse or do we need to add it.

in_toto/in_toto_record.py
```diff
@@ -14,166 +13,141 @@
...
+  if args.key:
+    try:
+      key = in_toto.util.prompt_import_rsa_key_from_file(args.key)
+
+    except Exception as e:
```
SantiagoTorres, line 132: 👍 for maximizing code reuse. I think we can add specific exception handling code here though. What do you think?

in_toto/in_toto_record.py
```diff
@@ -14,166 +13,141 @@
...
+      sys.exit(1)
+
   try:
-    key = in_toto.util.prompt_import_rsa_key_from_file(args.key)
+    if args.command == "start":
+      in_toto.runlib.in_toto_record_start(args.step_name, args.materials,
+          signing_key=key, gpg_keyid=gpg_keyid,
+          gpg_use_default=gpg_use_default, gpg_home=args.gpg_home)
+
+    elif args.command == "stop": # pragma: no branch
```
SantiagoTorres: I could go both ways, but if argparse already catches args.command == "start" || "stop" we could just use an else here no?

in_toto/runlib.py
```diff
@@ -487,34 +522,70 @@ def in_toto_record_start(step_name, key, material_list):
 
   link_metadata = Metablock(signed=link)
 
-  log.info("Signing link metadata with key '{:.8}...'...".format(key["keyid"]))
-  link_metadata.sign(key)
+  if signing_key:
+    log.info("Signing link metadata using passed key...")
+    signature = link_metadata.sign(signing_key)
+
+  elif gpg_keyid:
```
SantiagoTorres, line 529: I think we can take the `else` away by having `elif gpg_keyid or gpg_use_default:`

test/test_in_toto_record.py
```diff
@@ -43,101 +41,135 @@ class TestInTotoRecordTool(unittest.TestCase):
...
+    args = ["--step-name", "test6", "--gpg", "--gpg-home", self.gnupg_home]
+    if in_toto.gpg.util.is_version_fully_supported():
+      self._test_cli_sys_exit(["start"] + args, 0)
+      self._test_cli_sys_exit(["stop"] + args, 0)
+
+    else:
+      # Fail (exit 1) start recording on not fully supported gpg version
+      self._test_cli_sys_exit(["start"] + args, 1)
+
+      # NOTE: Stop recording with default gpg on not fully supported gpg
```
SantiagoTorres: I don't think this comment is clear enough...

in_toto/in_toto_record.py
lukpueh: Hm. I was actually referring to the (potentially) multiple steps in between `start` and `record,`. But you are right, this is not clear. What about something like:
*"...create a link metadata file in two steps, in order to provide evidence for arbitrary activities. Note: If you want to provide evidence for a single command take a look at in-toto-record."*?

in_toto/in_toto_record.py
lukpueh: Works either way, but yours is definitely more readable. I'll update and write a better comment.

in_toto/in_toto_record.py
lukpueh: Right, I'm using an argparse mutually exclusive group for `key` and `gpg` above. So this is taken care of.

in_toto/in_toto_record.py
lukpueh: Yes. This applies to all command line tools. I made a note to fix this in a separate PR https://github.com/in-toto/in-toto/issues/126#issuecomment-356707753

in_toto/in_toto_record.py
lukpueh: I had the same thoughts (in various places throughout the repo). I am not sure about *the right way* to do this. For the sake of readability something like `else: # args.command == "stop"` might be a good compromise. :)

test/test_in_toto_record.py
lukpueh: Ok, I'll reword.

in_toto/runlib.py
lukpueh: Or even `else: # gpg_keyid or gpg_use_default`, c.f. https://github.com/in-toto/in-toto/pull/167#discussion_r160774560 :)

lukpueh: Oh, the reason why I did it that way was to have different log statements. "*using passed GPG keyid"* vs. "*using default GPG key"*.
I could put the function call which is the same for both in an `else`, and have an additional `if/else` inside for the log statement, but I don't think that that would increase readability.

in_toto/in_toto_record.py
lukpueh: FYI https://github.com/secure-systems-lab/code-style-guidelines/issues/18

in_toto/runlib.py
vladimir-v-diaz: Another option is to relocate the log statement to `sign_gpg()`.  It might be easier to handle the "GPG keyid" vs "default GPG key" case there.

Note: The guidelines recommend never to use short circuit evaluation to produce output.
https://github.com/secure-systems-lab/code-style-guidelines#never-use-short-circuit-evaluation-to-produce-output

SantiagoTorres: Hmm, I think the line is blurry on what is "short circuit evaluation" then. This is just an `or` clause. Am I missing something?

SantiagoTorres: re "custom message for default keyid". I don't know how useful is it really. We can leave it though...

SantiagoTorres:
LGTM, other than that open discussion regarding the short-circuit evaluation.

in_toto/runlib.py
lukpueh: Alright, I added a pointer https://github.com/in-toto/in-toto/issues/117#issuecomment-356756728. Thanks for the review!

PR #168, opened 2018-01-10
Add support for Python 3.4, 3.5 and 3.6

### LINKED ISSUE #23: Python 3 compatibility
The current code base is mostly Python 3 compatible. Some minor changes are required though to use `in-toto` in both Python 2 and 3. 
- Replace `print` with `print()` (cf. @SantiagoTorres's and @vladimir-v-diaz's [comments](https://github.com/in-toto/in-toto/pull/19#discussion_r84122308)
- ...

lukpueh:
**Fixes issue #**:
Closes #23 

Add support for Python 3.4, 3.5 and 3.6 (in addition to Python 2.7 support).

Commits whose titles start with `Py2/3` contain all the changes for cross compatibility. The remaining commits update Travis and tox configs to run linters and unittests on all versions and fix coverage issues.

requirements**:

### DIFF
--- a/in_toto/gpg/util.py
+++ b/in_toto/gpg/util.py
@@ -207,7 +207,8 @@ def get_version():
 
   """
   command = shlex.split(in_toto.gpg.constants.GPG_VERSION_COMMAND)
-  process = subprocess.Popen(command, stdout=subprocess.PIPE)
+  process = subprocess.Popen(command, stdout=subprocess.PIPE,
+      universal_newlines=True)
   full_version_info, junk = process.communicate()
 
   version_string = re.search(r'(\d\.\d\.\d+)', full_version_info).group(1)
--- a/in_toto/gpg/util.py
+++ b/in_toto/gpg/util.py
@@ -146,7 +146,7 @@ def compute_keyid(pubkey_packet_data):
   hasher.update(b'\x99')
   hasher.update(struct.pack(">H", len(pubkey_packet_data)))
   hasher.update(bytes(pubkey_packet_data))
-  return binascii.hexlify(hasher.finalize())
+  return binascii.hexlify(hasher.finalize()).decode("ascii")
 
 
 def parse_subpackets(subpacket_octets):
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -43,7 +43,7 @@ from in_toto.models.metadata import Metablock
 
 # POSIX users (Linux, BSD, etc.) are strongly encouraged to
 # install and use the much more recent subprocess32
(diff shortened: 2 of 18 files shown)


in_toto/gpg/util.py
```diff
@@ -146,7 +146,7 @@ def compute_keyid(pubkey_packet_data):
   hasher.update(b'\x99')
   hasher.update(struct.pack(">H", len(pubkey_packet_data)))
   hasher.update(bytes(pubkey_packet_data))
-  return binascii.hexlify(hasher.finalize())
+  return binascii.hexlify(hasher.finalize()).decode("ascii")
```
SantiagoTorres, line 149: I wonder if for all of these instances we want an `errors='ignore'` or some sort of error handling...

in_toto/runlib.py
```diff
@@ -250,36 +249,18 @@ def execute_link(link_cmd_args, record_streams):
...
-        stdout=stdout_file, stderr=stderr_file)
-
-    stdout_file.seek(0)
-    stderr_file.seek(0)
+    process = subprocess.Popen(link_cmd_args, stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE, universal_newlines=True)
 
-    stdout_str = stdout_file.read()
-    stderr_str = stderr_file.read()
+    stdout_str, stderr_str = process.communicate()
```
SantiagoTorres, line 262: I thought we had opted to not do this due to performance concerns/buffering. Is this still the case?

SantiagoTorres (approved):
I'm actually surprised how many little changes we had to do.

in_toto/gpg/util.py
lukpueh: I think just blowing up is better. What we could do is to catch exceptions and re-raise with a redacted error message. Maybe worth a note in #126.

in_toto/runlib.py
lukpueh: I timed the function call for `ls -Rl /` and it executed without error in about 4 minutes, returning a stdout string of about 170MB.

```python
>>> import timeit
>>> timeit.Timer("in_toto.runlib.execute_link(['ls', '-Rl', '/'], True)", setup="import in_toto.runlib").timeit(1)
247.98906087875366
```

The timed `ls -Rl` in bash takes longer probably due to writing the ouput to the terminal.
```shell
$ time ls -Rl /
real  4m57.299s
user  0m51.453s
sys 2m20.026s
```

I think it's fine for now. Also it is only a temporary solution, if we decide to adapt https://github.com/in-toto/in-toto/pull/160

lukpueh:
Let me know if you think it is good to merge.

PR #150, opened 2017-11-14
WIP: gpg: add pgp interface.

SantiagoTorres:
This adds the first stab at the signature generation and verification
routines. This is work in progress, as the codebase must be cleaned and
probably refactored into something more concise.

This is still a WIP branch, but we can start discussing and udpating the code
here.

PR #169, opened 2018-01-16
WIP:debian: add debian packaging support

SantiagoTorres:
**Fixes issue #**:
No issue has been assigned to this

This pull request adds all the control files required for debian packaging of the in-toto CLI and python libraries

requirements**:

This is a WIP PR, we may need to change a thing or two before merging. Namely:

- There is no gpg2 package/binary on debian stable. It's just plain gpg, we may want to be smarter about it (probably add a patch on the debian side)
- Right now it's shipped with the default python version for the distribution. This may or may not be wrong, as we support both python versions...


debian/changelog
```diff
@@ -0,0 +1,5 @@
+in-toto (0.9.1) unstable; urgency=low
```
lukpueh: in-toto is at 0.1.1 currently but the initial Debian release will be for version 0.2.0.

debian/control
```diff
@@ -0,0 +1,40 @@
+Source: in-toto
```
lukpueh, line 1: should this be `python-in-toto`?

debian/control
```diff
@@ -0,0 +1,40 @@
...
+Section: devel
+Priority: optional
+Maintainer: in-toto developers <in-toto-dev@googlegroups.com>
+Uploaders:
+ Santiago Torres-Arias <santiago@nyu.edu>,
+ Lukas Puehringer <lukpueh@nyu.edu>,
```
lukpueh: Please fix my email address

debian/control
```diff
@@ -0,0 +1,40 @@
...
+ python-cryptography,
+ python-attr,
+ python-dateutil,
+ python-six,
+ python-iso8601,
+ python-mock,
+ dh-python (>= 2.20160818~),
+ dpkg-dev (>= 1.17.14),
+ python-securesystemslib (= 0.10.8),
+Standards-Version: 4.1.1
```
lukpueh: IIUC this refers to the [Debian policy version](https://www.debian.org/doc/debian-policy/). How did you choose this version? The most recent version is 4.1.3, released on 2017-12-27

debian/control
```diff
@@ -0,0 +1,40 @@
...
+Homepage: https://in-toto.io
+Vcs-Git: https://github.com/in-toto/in-toto.git
+Vcs-Browser: https://github.com/in-toto/in-toto
+
+Package: in-toto
```
lukpueh, line 26: Should this also be `python-in-toto`?

debian/README.source
```diff
@@ -0,0 +1,15 @@
+in-toto: hacking tips
+========================
+
+Running the test suite
+----------------------
+
+In order to test you need unittest installed. An easy way to run the tests is
+to run
+
+    $ PYTHONPATH=. ./tests/runtests.py
```
lukpueh, line 10: The script is in `test/runtests.py`

debian/copyright
```diff
@@ -0,0 +1,31 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: in-toto
+Source: https://github.com/in-toto/in-toto
+
+Files: *
+Copyright: 2015-2017 New York University
```
lukpueh: Why 2015-2017, when the LICENSE file says 2016?

debian/rules
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/make -f
+
+export PATH := $(CURDIR)/bin:$(PATH)
+export PYBUILD_DESTDIR = debian/in-toto
```
lukpueh, line 4: `debian/python-in-toto`?

debian/rules
```diff
@@ -0,0 +1,10 @@
...
+export PYBUILD_SYSTEM = distutils
+export PYBUILD_TEST_ARGS = test 
+export PYBUILD_NAME = in_toto
+
+%:
+	dh $@ --with python2  --buildsystem=pybuild
```
lukpueh, line 10: We can add `python3` now

debian/upstream/signing-key.asc
```diff
@@ -0,0 +1,97 @@
...
+PBJJ0rwN6ST+M8Rl+7KsGpu1fm73D9v9WzFAnIIwEBoo8ON0VfJQ5NXAcSwf/qGN
+tQp+9rEuDENUtYhN9nMv6/ibO7lw8PM5TCzHhGQ2sp0Eob07hvXbgN72WibkcVj1
+z0rwixZnDCRP8qUGqVGlbaND2U84Usf2U7yLJpETbduwku/yoiDfyL/kdz8kEA4y
+U/dBBwjo34NtCHhTEiyrEJyPYzf+Hho1KSROZzjtCQleoYXiL89Dv96+4WWJ//4W
+/lXYCl/ljW0Py7XQ0IdZbhxeJht08VMebbrxiXBBWLWuS+/xrhBX3ajk002Xr1Ms
+SoIFHjORXUIGUpyHAT6sY6MlpIZP/76tY5yzAk8DqJ60MNbS+Jyrqk9eaZ+lc9uv
+t6V0KQe7Uhsd6pllGw6q5cBfyzyK3l8goADYT9ox3b+cRXgrYgHH9cTyG6Uf+rnz
+pvNV0uPx4mvSq/w6D1tWqWHLgBa4AgDxv3f8JxOqDrc/Ug==
+=p3C2
+-----END PGP PUBLIC KEY BLOCK-----
```
lukpueh, line 97: I saw that this file is used together with a [`debian/watch` file and the `pgpsigurlmangle` option](https://www.debian.org/doc/debian-policy/#optional-upstream-source-location-debian-watch). Do we want to add such a file?

setup.py
```diff
@@ -64,7 +64,8 @@
     'Topic :: Security',
     'Topic :: Software Development'
   ],
-  packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
+  packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests",
+      "debian"]),
```
lukpueh: Unrelated: 
Funny thing, we are not excluding the only `test` directory of the project. :)
For our setup the following line would be what we want:
```python
packages=find_packages(exclude=["test", "test.*", "debian"])
```
Or maybe explicitly including packages would be even easier?,
```python
packages=["in_toto", "in_toto.gpg", "in_toto.models"]
```

Worth an extra PR?

setup.py
```diff
@@ -64,7 +64,8 @@
...
   install_requires=["six", "securesystemslib==0.10.8", "attrs",
                     "python-dateutil", "iso8601"],
```
lukpueh: Unrelated:
Looking at `debian/control`, where you correctly added cryptography as a dependency (re-introduced in #150), I realized that we have to re-add it here and in `requirements.txt` as well. Worth an extra PR?

debian/control
vladimir-v-diaz: @lukpueh 

Always causing problems...
Fix your GitHub username so that it matches your email address. :)

setup.py
lukpueh: Fixed with https://github.com/in-toto/in-toto/pull/173. Please merge there first and resolve conflict here.

setup.py
lukpueh: Fixed with #174

debian/README.source
lukpueh: `test` will be renamed to `tests` with https://github.com/in-toto/in-toto/pull/173.

SantiagoTorres: hmm, so would we change this back and forth? what do you suggest?

debian/changelog
SantiagoTorres: ok, will change to 0.2.0

debian/control
SantiagoTorres: No, since it's a CLI tool, not just a python library.

debian/control
SantiagoTorres: Vlad speaking the truth here 👍

debian/control
SantiagoTorres: I chose this version based on what other packages that we depend have. We can update the policy down the line, but it's not a blocker right now.

debian/copyright
SantiagoTorres: hmm, this is a good question. Didn't we start this project in 2015? I'm curious as to why we have 2016 over there

debian/README.source
lukpueh: #173 is already merged. So you can disregard my comment. :)

debian/rules
SantiagoTorres: I'm hesitant on this, due to this being a CLI tool. I think we can settle these with h0lger.

debian/upstream/signing-key.asc
SantiagoTorres: I don't think it's mandatory. Do you think it's worth adding given our time constraints?

debian/copyright
lukpueh: The initial commit was in May 2016.

setup.py
SantiagoTorres: kk, now I get what you mean. I'll rebase in a sec.

debian/control
lukpueh: Got it.

debian/control
lukpueh: Let's leave it like this for the time being.

debian/upstream/signing-key.asc
lukpueh: We can add this later.

debian/copyright
SantiagoTorres: Kk. I'll change for consistency

SantiagoTorres:
@lukpueh @vladimir-v-diaz ping. I think all comments are addressed.

vladimir-v-diaz:
https://pypi.org/project/securesystemslib/
https://github.com/secure-systems-lab/securesystemslib/releases/tag/v0.10.9

Cheers.

PR #170, opened 2018-01-16
Revise signature and step threshold verification

### LINKED ISSUE #135: Revisit signature verification

lukpueh:
**Fixes issue #**:
Closes #135 #136 

Revises in-toto layout and link signature verification.

The PR aims at better separating cryptographic verification of signatures from the authorization and signature requirement model as specified by in-toto.

For more information read the issue descriptions closed by this PR and the commit messages

The PR also makes adds some minor unrelated unittest fixes to obtain 100% code coverage :)

requirements**:

### DIFF
--- a/in_toto/gpg/common.py
+++ b/in_toto/gpg/common.py
@@ -184,7 +184,7 @@ def parse_signature_packet(data):
   # partial one is available.
   keyid = []
   for subpacket_tuple in hashed_subpacket_info:
-    if subpacket_tuple[0] == FULL_KEYID_SUBPACKET:
+    if subpacket_tuple[0] == FULL_KEYID_SUBPACKET: # pragma: no cover
       keyid.append(subpacket_tuple)
 
   # Excluded so that coverage does not vary in different test environments
--- a/in_toto/models/metadata.py
+++ b/in_toto/models/metadata.py
@@ -180,76 +180,71 @@ class Metablock(object):
     return signature
 
 
-  def verify_signatures(self, keys_dict):
+  def verify_signature(self, verify_key):
     """
     <Purpose>
       Verifies all signatures found in `self.signatures` using the public keys
       from the passed dictionary of keys and the utf-8 encoded canonical JSON
       bytes of the Link or Layout object contained in `self.signed`.
 
-      If a signature matches `in_toto.gpg.formats.SIGNATURE_SCHEMA`,
+      If the signature matches `in_toto.gpg.formats.SIGNATURE_SCHEMA`,
       `in_toto.gpg.functions.gpg_verify_signature` is used for verification,
-      `securesystemslib.keys.verify_signature` is used otherwise.
+      if the signature matches `securesystemslib.formats.SIGNATURE_SCHEMA`
+      `securesystemslib.keys.verify_signature` is used.
 
       Note: In case of securesystemslib we actually pass the dictionary
       representation of the data to be verified and
       `securesystemslib.keys.verify_signature` converts it to
       canonical JSON utf-8 encoded bytes before verifying the signature.
 
-
-      Verification fails if,
-        - the passed keys don't have the right format,
-        - the object is not signed,
-        - there is a signature for which no key was passed,
-        - if any of the signatures is actually broken.
-
-      Note:
-      This will be revised with in-toto/in-toto#135
-
     <Arguments>
-      keys_dict:
-              Verifying keys in the format:
-              in_toto.formats.ANY_VERIFY_KEY_DICT_SCHEMA
+      verify_key:
+              Verifying key in the format:
+              in_toto.formats.ANY_VERIFY_KEY_SCHEMA
 
     <Exceptions>
       FormatError
-            if the passed key dictionary is not conformant with
-            in_toto.formats.ANY_VERIFY_KEY_DICT_SCHEMA
+            If the passed key is not conformant with
+            `in_toto.formats.ANY_VERIFY_KEY_SCHEMA`
 
       SignatureVerificationError
-            if the Metablock is not signed
+            If the Metablock does not carry a signature signed with the
+            private key corresponding to the passed verification key
 
-            if the Metablock carries a signature for which no key is found in
-            the passed key dictionary, which means that multiple signatures
-            have to be verified at once
+            If the signature corresponding to the passed verification key
+            does not match securesystemslib's or in_toto.gpg's
+            signature schema.
 
-            if any of the verified signatures is actually broken
+            If the verified signature is invalid.
 
     <Returns>
       None.
 
     """
-    in_toto.formats.ANY_VERIFY_KEY_DICT_SCHEMA.check_match(keys_dict)
+    in_toto.formats.ANY_VERIFY_KEY_SCHEMA.check_match(verify_key)
+    verify_keyid = verify_key["keyid"]
 
-    if not self.signatures or len(self.signatures) <= 0:
-      raise SignatureVerificationError("No signatures found")
-
+    # Find signature by verification key id and raise exception if not found
+    signature = None
     for signature in self.signatures:
-      keyid = signature["keyid"]
-      try:
-        key = keys_dict[keyid]
+      if signature["keyid"] == verify_keyid:
+        break
+    else:
+      raise SignatureVerificationError("No signature found for key '{}'"
+          .format(verify_keyid))
 
-      except KeyError:
-        raise SignatureVerificationError(
-            "Signature key not found, key id is '{0}'".format(keyid))
+    if in_toto.gpg.formats.SIGNATURE_SCHEMA.matches(signature):
+      valid = in_toto.gpg.functions.gpg_verify_signature(signature, verify_key,
+          self.signed.signable_bytes)
 
-      if in_toto.gpg.formats.SIGNATURE_SCHEMA.matches(signature):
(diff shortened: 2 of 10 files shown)


in_toto/models/metadata.py
```diff
@@ -180,76 +180,71 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
...
 
-            if the Metablock carries a signature for which no key is found in
-            the passed key dictionary, which means that multiple signatures
-            have to be verified at once
+            If the signature corresponding to the passed verification key
+            does not match securesystemslib's or in_toto.gpg's
+            signature schema.
 
-            if any of the verified signatures is actually broken
+            If the verified signature is invalid.
```
awwad: "Verified signature" implies validity.

Try one of these or similar:
- "If the signature to be verified is invalid"
- "If the signature is invalid"

in_toto/models/metadata.py
```diff
@@ -180,76 +180,71 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
     return signature
 
 
-  def verify_signatures(self, keys_dict):
+  def verify_signature(self, verify_key):
```
SantiagoTorres: I'd rename `verify_key` to `verification_key`.

in_toto/models/metadata.py
```diff
@@ -180,76 +180,71 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
...
 
     """
-    in_toto.formats.ANY_VERIFY_KEY_DICT_SCHEMA.check_match(keys_dict)
-
-    if not self.signatures or len(self.signatures) <= 0:
-      raise SignatureVerificationError("No signatures found")
+    in_toto.formats.ANY_VERIFY_KEY_SCHEMA.check_match(verify_key)
+    verify_keyid = verify_key["keyid"]
 
+    # Find signature by verification key id and raise exception if not found
```
SantiagoTorres: I'd also rephrase this comment. Probably to something like

```python
# Find a signature that corresponds to this verification keyid and raise an exception if it doesn't exist.
```

SantiagoTorres:
Two incredibly minor things imho 👍

in_toto/gpg/common.py
```diff
@@ -184,7 +184,7 @@ def parse_signature_packet(data):
   # partial one is available.
   keyid = []
   for subpacket_tuple in hashed_subpacket_info:
-    if subpacket_tuple[0] == FULL_KEYID_SUBPACKET:
+    if subpacket_tuple[0] == FULL_KEYID_SUBPACKET: # pragma: no cover
       keyid.append(subpacket_tuple)
```
awwad, line 188: I don't understand this enough to know why it merits `no cover`. (not reviewed)

tests/test_verifylib.py
```diff
@@ -1016,6 +1013,260 @@ def test_verify_failing_inspection_rules(self):
...
+        self.alice_keyid: link_alice,
+      }
+    }
+
+    with self.assertRaises(ThresholdVerificationError):
+      verify_threshold_constraints(layout, chain_link_dict)
+
+
+
+  def test_treshold_constraints_pas_with_equal_links(self):
```
awwad: typos in this line

(if anything calls this, that presumably has to be changed, too)

tests/test_verifylib.py
```diff
@@ -1016,6 +1013,260 @@ def test_verify_failing_inspection_rules(self):
...
+      self.name: {
+        self.bob_keyid: link_bob,
+      }
+    }
+
+    with self.assertRaises(ThresholdVerificationError):
+      verify_threshold_constraints(layout, chain_link_dict)
+
+
+  def test_treshold_constraints_fail_with_unequal_links(self):
```
awwad: typo - threshold

(if anything calls this, that presumably has to be changed)

in_toto/models/metadata.py
```diff
@@ -180,76 +180,72 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
     return signature
 
 
-  def verify_signatures(self, keys_dict):
+  def verify_signature(self, verification_key):
     """
     <Purpose>
       Verifies all signatures found in `self.signatures` using the public keys
       from the passed dictionary of keys and the utf-8 encoded canonical JSON
       bytes of the Link or Layout object contained in `self.signed`.
```
awwad: This docstring is wrong now, right? It verifies just the one signature?

in_toto/models/metadata.py
```diff
@@ -180,76 +180,72 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
...
+    elif securesystemslib.formats.SIGNATURE_SCHEMA.matches(signature):
+      valid = securesystemslib.keys.verify_signature(
+          verification_key, signature, self.signed.signable_dict)
 
-      else:
-        valid = securesystemslib.keys.verify_signature(
-            key, signature, self.signed.signable_dict)
+    else:
+      # TODO: Add Metablock format validator that prevent malformed signatures
+      valid = False
```
awwad, line 248: I think that an error should be raised here indicating that the key is not in a known or supported format.

in_toto/verifylib.py
```diff
@@ -300,34 +289,68 @@ def verify_all_steps_signatures(layout, chain_link_dict):
...
-    Raises an exception if a needed key can not be found in the passed
-    keys_dict or if a verification fails.
-    TBA (see https://github.com/in-toto/in-toto/issues/6)
+    ThresholdVerificationError
+            If any of the steps of the passed layout does not have enough
+            (`step.threshold`) links signed by different authorized
+            functionaries.
+
+  <Returns>
+    A chain_link_dict containing only links with valid signatures created by
```
awwad, line 298: Very minor point: link_chain suggests a chain (made of links) and chain_link suggests a link (for a chain). Since this is a chain, I'd go with the former. (If it seems to make sense to people, don't worry about it. /: )

awwad (commented):
Makes sense (:

in_toto/gpg/common.py
lukpueh: The unit tests only cover this on systems with fully supported versions of gpg, which unfortunately travis does not have. I hinted at it in the commit message. :)

in_toto/models/metadata.py
lukpueh: This is odd. I was convinced that I had updated the docstring. Probably lost in rebase. Will fix. Good eyes, anyways! 👍

in_toto/models/metadata.py
lukpueh: Will change

in_toto/models/metadata.py
lukpueh: I thought about that too, but I think it's okay (for now) to treat a malformed signature and an invalid signature alike. In any case, there is a TODO comment, to move the format validation elsewhere.

in_toto/verifylib.py
lukpueh: I remember the discussion about the names for these dictionaries (c.f. https://github.com/in-toto/in-toto/pull/69#discussion_r100401016). I am definitely not happy with the current names, but I'm not convinced that `link_chain` is a huge improvement either. Given that the dicts are fairly well described in the docstring args sections (with examples), I guess we can leave them for the time being.

tests/test_verifylib.py
lukpueh: Good eyes! Will fix (here and below).

lukpueh:
Thanks for the reviews, @awwad and @SantiagoTorres!

PR #173, opened 2018-01-17
Rename 'test' directory to 'tests'

lukpueh:
**Fixes issue #**:
Fixes comment in unrelated review https://github.com/in-toto/in-toto/pull/169#discussion_r162169144

Despite all possible wildcard and "tests" dir combinations in `setup'py`'s  `packages` section,  we didn't exclude the only test directory that exists in the project, i.e. `test`, as intended.

This PR switches to explicitly naming the few packages we want to include, which also obsoletes the `find_package` module. 

If we want to go back to excluding unwanted packages we'd have to use (already taking into account the soon to be [merged, but to be excluded "debian" dir](https://github.com/in-toto/in-toto/pull/169#discussion_r162169144)):
```python
packages=find_packages(exclude=["test", "test.*", "debian"])
```
or
```python
packages=find_packages(exclude=["test", "test.models", "debian"])
```

**UPDATE**
As pointed out by @vladimir-v-diaz, it is better to [exclude the unwanted packages](https://github.com/in-toto/in-toto/pull/173#discussion_r162384472). 

I reverted the initial patch of this PR (and the PR title) and instead renamed the `test` directory to `tests` in order to align with the original exclusions in `setup.py` and common naming practices in other projects.

requirements**:


setup.py
```diff
@@ -64,7 +64,7 @@
     'Topic :: Security',
     'Topic :: Software Development'
   ],
-  packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
+  packages=["in_toto", "in_toto.models", "in_toto.gpg"],
```
vladimir-v-diaz: One potential downside to this approach is that if you later add a subdirectory or rename a subdirectory, you will have to remember to update `setup.py`.  You might only notice the mistake when users install from release and complain about import errors.

Assuming that Travis doesn't build from a release, continuous integration might not catch the mistake.

lukpueh: You have a point there. Maybe the best thing would be to leave the line as it was (recommended by [`setuptools` docs](http://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages)) and instead rename the the `test` directory to `tests`? What do you think?

lukpueh: ... or revert the commit and remove the trailing `s` from `tests` in the excludes.

vladimir-v-diaz:
I recommend that you follow the approach used in the setuptools docs.

Renaming `test` to `tests` makes sense.  Since there is more than one unit test, naming it `tests` seems more precise.  I think most projects use the directory name `tests`.

PR #177, opened 2018-01-24
Add metablock validators

lukpueh:
Add `_validate*` methods (using `ValidationMixin`) to `Metablock` class, to validate that the `signed` property contains a valid `Link` or `Layout` object and the `signatures` property is a list and only contains valid signatures.
Validation is performed at the end of `Metablock`s constructor.

*unrelated*: This PR also adds an comment to an existing layout validator.

requirements**:

### DIFF
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -136,6 +136,8 @@ class Layout(Signable):
     """Private method to verify if the expiration field has the right format
     and can be parsed."""
     try:
+      # We do both 'parse' and 'check_match' because the format check does not
+      # detect bogus dates (e.g. Jan 35th) and parse can do more formats.
       parse(self.expires)
       securesystemslib.formats.ISO8601_DATETIME_SCHEMA.check_match(
           self.expires)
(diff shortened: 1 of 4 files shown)


in_toto/models/layout.py
```diff
@@ -136,6 +136,8 @@ def _validate_expires(self):
     """Private method to verify if the expiration field has the right format
     and can be parsed."""
     try:
+      # We do both 'parse' and 'check_match' because the format check does not
+      # detect bogus dates (e.g. Jan 35th) and parse can do more formats.
```
awwad, line 140: Regarding the second part -- "and parse can do more formats": is that to say that parse accepts multiple formats and so won't verify that the format is the correct one (and that is why check_match is still needed)?

awwad (approved):
Looks fine to me.

in_toto/models/layout.py
lukpueh: yep, that's what it means. should I elaborate in the comment?

PR #178, opened 2018-01-24
Add pack artifact rules

lukpueh:

Steps and Inspections carry artifact rules in list forms. To parse a rule in list form into a dictionary (`rule_data`), which is easier to work with, we use `unpack_rule`.

This PR adds a `pack_rule` function to reverse the unpacking, with arguments based on the fields in `rule_data`, plus a shortcut to directly use the `rule_data` dictionary returned from, e.g:
`rule == pack_rule_data(unpack_rule(rule))`.

Additionally this commit adds shortcuts for create, delete, modify, allow and disallow rules.

This functions can be used for layout creation.

This PR further 
- adds string and list of string schemas,
- fixes artifact rule docstrings, and
- renames `artifact_rules` to `rulelib` 
  *Note: IMHO the lib part makes it more clear that the functions can be used autonomously, e.g. when creating a layout (c.f. an upcoming PR)*

requirements**:

### DIFF
--- a/in_toto/rulelib.py
+++ b/in_toto/rulelib.py
@@ -1,6 +1,6 @@
 """
 <Module Name>
-  in_toto/artifact_rules.py
+  in_toto/rulelib.py
 
 <Author>
   Santiago Torres-Arias <santiago@nyu.edu>
--- a/in_toto/artifact_rules.py
+++ b/in_toto/artifact_rules.py
@@ -21,6 +21,10 @@ import in_toto.formats
 import securesystemslib.exceptions
 import securesystemslib.formats
 
+GENERIC_RULES = {"create", "modify", "delete", "allow", "disallow",}
+COMPLEX_RULES = {"match",}
+ALL_RULES = GENERIC_RULES | COMPLEX_RULES
+
 def unpack_rule(rule):
   """
   Parses the rule and extracts and returns the necessary data to apply the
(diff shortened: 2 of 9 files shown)


in_toto/rulelib.py
```diff
@@ -167,10 +155,133 @@ def unpack_rule(rule):
...
+      MATCH <pattern> [IN <source-path-prefix>] WITH (MATERIALS|PRODUCTS)
+          [IN <destination-path-prefix>] FROM <step>,
+      CREATE <pattern>,
+      DELETE <pattern>,
+      MODIFY <pattern>,
+      ALLOW <pattern>,
+      DISALLOW <pattern>
+
+  """
+  in_toto.formats.ANY_STRING_SCHEMA.check_match(rule_type)
```
SantiagoTorres, line 217: I don't know if it's overkill, but we could probably have an `MATCH_RULE_SCHEMA` using an `anyof` of `ALL_RULES` and simplify the if below. Specially considering we are going heavy on the scheme checkers.

in_toto/rulelib.py
```diff
@@ -167,10 +155,133 @@ def unpack_rule(rule):
...
+        (dest_type.lower() == "materials" or dest_type.lower() == "products")):
+      raise securesystemslib.exceptions.FormatError("'{}' is not a valid"
+          " 'dest_type'. Rules of type 'MATCH' require a destination type of"
+          " either 'MATERIALS' or 'PRODUCTS' (case insensitive)."
+          .format(dest_type))
+
+    if not (in_toto.formats.ANY_STRING_SCHEMA.matches(dest_name) and dest_name):
+      raise securesystemslib.exceptions.FormatError("'{}' is not a valid"
+          " 'dest_name'. Rules of type 'MATCH' require a step name as a"
+          " destination name.".format(dest_name))
```
SantiagoTorres, line 236: likewise, I wonder if we could turn these checks into schema checks...

SantiagoTorres (approved):
Minor style suggestions. LGTM

in_toto/rulelib.py
lukpueh: Definitely a good idea to use schema. But maybe worth an extra refactor ticket?

lukpueh:
Thanks for the review, @SantiagoTorres. I ticketized your request in #185.

PR #179, opened 2018-01-25
Remove command parse from Step/Inspection constructors

lukpueh:
We used to accept the command for `Step`'s `expected_command` and `Inspection`'s `run` properties as list or string and in the latter case transform it to a list using `shlex`. 

However, it is better to not magically transform the metadata in the constructors, as it is likely to break existing signatures. Instead we provide the convenient parsing functionality in separate `Step`/`Inspection` methods.

requirements**:

### DIFF
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -44,6 +44,7 @@ import in_toto.formats
 
 import securesystemslib.exceptions
 import securesystemslib.formats
+import securesystemslib.schema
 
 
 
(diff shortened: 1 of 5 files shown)


awwad (approved):
This looks good. It's a good thing to expect to sign the format of expected_command provided: lists are expected to be provided now, and lists are what is signed. Previously, strings could be provided and lists would be signed. The new function to do that conversion isn't called automatically but is kept as a convenience for the user, which is good.

It's worth noting, though, separate from this PR, that taking the Step definition dictionary as **kwargs in the Step constructor means that if there are extra elements in the dictionary, those will be ignored rather than raising any sort of error. (This is *probably* not what you want, though that's debatable.) It's further true that since there are default arguments, if there is, say, a typo in an element name in the original dictionary, that element will be ignored and the default value used silently. This is probably bad and you can fix that by changing the constructor from using kwargs to using specific arguments. While that would occupy more lines, that would have the additional advantage of being more readable, I think.

Once that's done, it would be wise to add some more test cases, testing the behaviors I just described.

lukpueh:
Thanks for the review, @awwad. Your comment will be addressed with #186.

PR #180, opened 2018-01-25
Generalize step and inspection

lukpueh:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #**:
Closes #16 

**Description of the changes being introduced by the pull request**:
The `Step` and `Inspection` class both carry `expected_materials` and `expected_products` and already are treated similarly during final product verification, in terms of rule verification. Additionally both items have a name that is used to associate link metadata with the corresponding item.

This PR adds a superclass `SupplyChainItem` (suggestions for a better name are welcome) in order to reduce code redundancy.

This PR also overrides `SupplyChainItem`'s `__repr__` method to return a JSON formatted String like for `Metablock`, `Layout` and `Link`.

This PR also adds convenience methods to parse artifact rules from a string into a list and add it to a supply chain item's `expected_materials` or `expected_products` property.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #181, opened 2018-01-26
Add layout creation convenience methods

lukpueh:
Adds methods to the layout class to ease layout creation.

Here is a list of the newly added methods:
- set_relative_expiration
- get_step_name_list
- get_step_by_name
- remove_step_by_name
- get_inspection_name_list
- get_inspection_by_name
- remove_inspection_by_name
- get_functionary_key_id_list
- add_functionary_key
- add_functionary_key_from_path
- add_functionary_key_from_gpg_keyid
- add_functionary_keys_from_paths
- add_functionary_keys_from_gpg_keyids

Additionally refines and fixes various docstrings, comments, exceptions messages and spacings in the the layout class.

Please merge #180 and #179 first. I'd like to give the docstrings, comments in the parts of the layout module affected by those PRs another pass here.

requirements**:

### DIFF
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -362,11 +362,7 @@ class Layout(Signable):
       A list of functionary keyids.
 
     """
-    keyid_list = []
-    for keyid in list(self.keys.keys()):
-      keyid_list.append(keyid)
-
-    return keyid_list
+    return list(self.keys.keys())
 
 
   def add_functionary_key(self, key):
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -779,17 +779,12 @@ class Inspection(SupplyChainItem):
 
   """
   _type = attr.ib()
-  name = attr.ib()
   run = attr.ib()
 
 
   def __init__(self, **kwargs):
     super(Inspection, self).__init__(**kwargs)
     self._type = "inspection"
-    self.name = kwargs.get("name")
-    self.expected_materials = kwargs.get("expected_materials", [])
-    self.expected_products = kwargs.get("expected_products", [])
-
     self.run = kwargs.get("run", [])
 
     self.validate()
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -111,7 +111,7 @@ class Layout(Signable):
     # Assign a default expiration (one month) if not expiration date is passed
     # TODO: Is one month a sensible default? In any case, we need a valid
     # expiration date in order for the layout object to validate.
-    # (c.f. self._validate_expires)
+    # (see self._validate_expires)
     self.expires = kwargs.get("expires")
     if not self.expires:
       self.set_relative_expiration(months=1)
@@ -142,7 +142,7 @@ class Layout(Signable):
     """
     <Purpose>
       Static method to instantiate a layout object from a Python dictionary,
-      e.g.: by parsing its JSON representation. The method expects any
+      e.g. by parsing its JSON representation. The method expects any
       contained steps and inspections to be Python dictionaries as well, and
       tries to instantiate the corresponding objects using the step's and
       inspection's read methods respectively.
@@ -493,7 +493,8 @@ class Layout(Signable):
     return key_dict
 
 
-  def add_functionary_keys_from_gpg_keyids(self, gpg_keyid_list, gpg_home=None):
+  def add_functionary_keys_from_gpg_keyids(self, gpg_keyid_list,
+      gpg_home=None):
     """
     <Purpose>
       Load functionary public keys from the GPG keychain, located at the
@@ -573,7 +574,7 @@ class Layout(Signable):
     names_seen = set()
     if type(self.steps) != list:
       raise securesystemslib.exceptions.FormatError(
-          "The steps section should be a list!")
+          "The steps field should be a list!")
 
     for step in self.steps:
       if not isinstance(step, Step):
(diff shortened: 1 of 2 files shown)


in_toto/models/layout.py
```diff
@@ -87,31 +66,99 @@ class Layout(Signable):
...
+    <Purpose>
+      Static method to instantiate a layout object from a Python dictionary,
+      e.g.: by parsing its JSON representation. The method expects any
+      contained steps and inspections to be Python dictionaries as well, and
+      tries to instantiate the corresponding objects using the step's and
+      inspection's read methods respectively.
+
+    <Arguments>
+      data:
+              A dictionary containing layout metadata.
```
SantiagoTorres, line 152: Can we add "conforming to the in-toto specification" (and probably the version too?)

in_toto/models/layout.py
```diff
@@ -126,12 +173,376 @@ def read(data):
...
+              If any of days, months or years is passed and is not an
+              integer.
+
+    """
+    securesystemslib.schema.Integer().check_match(days)
+    securesystemslib.schema.Integer().check_match(months)
+    securesystemslib.schema.Integer().check_match(years)
+
+    self.expires = (datetime.today() + relativedelta(
+        days=days, months=months, years=years)).strftime("%Y-%m-%dT%H:%M:%SZ")
```
SantiagoTorres, line 204: Do we have a schema for date? can we add a guard?

in_toto/models/layout.py
```diff
@@ -126,12 +173,376 @@ def read(data):
...
+      Return a list of the functionary keyids from the layout's keys
+      dictionary.
+
+    <Returns>
+      A list of functionary keyids.
+
+    """
+    keyid_list = []
+    for keyid in list(self.keys.keys()):
+      keyid_list.append(keyid)
```
SantiagoTorres: I know this goes against the our guidelines, but a list comprehension here would be way more readable imho.

tests/models/test_layout.py
```diff
@@ -19,25 +19,198 @@
...
+      layout.set_relative_expiration(days=0, months=0, years=[])
+
+    layout.set_relative_expiration(days=1)
+    layout._validate_expires()
+    layout.set_relative_expiration(months=2)
+    layout._validate_expires()
+    layout.set_relative_expiration(years=3)
+    layout._validate_expires()
+    layout.set_relative_expiration(days=3, months=2, years=1)
+    layout._validate_expires()
```
SantiagoTorres, line 96: can you throw some negative numbers in here to see what happens? :)

tests/models/test_layout.py
```diff
@@ -19,25 +19,198 @@
...
+      layout.get_step_by_name(None)
+
+
+  def test_remove_step_by_name(self):
+    """Test removing step by name. """
+    names = ["a", "b", "c"]
+    layout = Layout(steps=[Step(name=name) for name in names])
+    layout.remove_step_by_name("b")
+    self.assertEqual(len(layout.steps), 2)
+    self.assertTrue("b" not in layout.get_step_name_list())
```
SantiagoTorres: Can you try a double-remove?

tests/models/test_layout.py
```diff
@@ -19,25 +19,198 @@
...
+  def test_remove_inspection_by_name(self):
+    """Test removing inspection by name. """
+    names = ["a", "b", "c"]
+    layout = Layout(inspect=[Inspection(name=name) for name in names])
+    layout.remove_inspection_by_name("b")
+    self.assertEqual(len(layout.inspect), 2)
+    self.assertTrue("b" not in layout.get_inspection_name_list())
+
+    with self.assertRaises(securesystemslib.exceptions.FormatError):
+      layout.remove_inspection_by_name(False)
```
SantiagoTorres: I wonder if testing the following sequence would be a good idea:

1. Add step `a`
1. Remove step `a`
1. Add *inspection* `a`
1. Remove inspection `a`

SantiagoTorres (changes requested):
Looking good, but I think it's missing a little bit more tests.

in_toto/models/layout.py
lukpueh: Will fix with #186.

in_toto/models/layout.py
lukpueh: There's a `_validate_expires` method on the Layout, which we could use (it checks schema and date validity). But I'm unit testing `set_relative_expiration` together with `_validate_expires`, so I don't think it is necessary to add a guard here.

in_toto/models/layout.py
lukpueh: I guess that's a matter of taste. And while I'm a big fan of list comprehensions, I think we should stick to our guidelines.

tests/models/test_layout.py
lukpueh: Sure, will do.

tests/models/test_layout.py
lukpueh: Sure. What's expected behavior? Currently it would do nothing.

in_toto/models/layout.py
vladimir-v-diaz: Why is `keyid_list` needed?  Doesn't the following work?
```Python
return list(self.keys.keys())
```

Note: the list() is needed for Python 2+3 compatibility, otherwise `return self.keys.keys()` should be fine.

tests/models/test_layout.py
lukpueh: Why exactly? Note, I didn't add `add_{step, inspection}` methods in this PR, which need testing. Currently the remove methods are tested separately like so:
```
for item in [step, inspection]:
    create layout with items a, b, c
    remove item a
    assert that item is not on on layout
```
However, I can add an assert that the item is on the layout before being removed.

in_toto/models/layout.py
lukpueh: Good thinking Vlad! :)

lukpueh:
Thanks, @SantiagoTorres and @vladimir-v-diaz. I replied to/ addressed your comments. And did some more docstring cleaning. Mind giving it another pass?

SantiagoTorres (approved):

PR #182, opened 2018-01-26
Add a layout creation example document

lukpueh:
Adds a markdown document that provides a Python snippet to create a demo layout like the one used in the in-toto/demo.

Also updates README to link to that document instead of the in-toto demo layout creation script and to link to the layout creation web tool.

Note that the methods used in the here added code snipped are added in the PRs #179, #180 and #181. So it would be good to merge those PRs first.

requirements**:

### DIFF
--- a/layout-creation.md
+++ b/layout-creation.md
@@ -1,9 +1,14 @@
 # Layout Creation Example
 
-The following Python snippet shows how to use in-toto to create a software
-supply chain layout like the one that is used for the in-toto demo.
-*Take a look at the [demo repo](https://github.com/in-toto/demo) for more
-information about that supply chain*.
+While it is possible to write an in-toto layout from scratch using a text
+editor and the JSON format, the preferred way is to use the in-toto model
+classes provided by the reference implementation.
+
+This document shows how to use in-toto's model classes and their convenience
+methods to create a supply chain layout like the one that is being used in the
+in-toto demo. Take a look at the [demo repo](https://github.com/in-toto/demo)
+for more details about the supply chain represented by this layout.
+
 
 
 ```python
@@ -12,16 +17,21 @@ from securesystemslib.interface import (generate_and_write_rsa_keypair,
 from in_toto.models.layout import Layout, Step, Inspection
 from in_toto.models.metadata import Metablock
 
-# Create RSA private and public keys
 
-# Alice is the project owner and her private key is used to sign the layout
+# in-toto provides functions to create RSA key pairs if you don't have them yet
+
+# In this example Alice is the project owner, whose private key is used to sign
+# the layout. The corresponding public key will be used during final product
+# verification.
 alice_path = generate_and_write_rsa_keypair("alice", password="123")
 alice_key = import_rsa_privatekey_from_file(alice_path, password="123")
 
 # Bob and Carl are both functionaries, i.e. they are authorized to carry out
-# different steps of the supply chain. Their public keys are added to the
-# supply chain layout, in order to verify the signatures of the link metadata
-# that Bob and Carl will generate when carrying out their respective tasks.
+# different steps of the supply chain. Their public keys will be added to the
+# layout, in order to verify the signatures of the link metadata that Bob and
+# Carl will generate when carrying out their respective tasks.
+# Bob and Carl will each require their private key when creating link metadata
+# for a step.
 bob_path = generate_and_write_rsa_keypair("bob", password="123")
 carl_path = generate_and_write_rsa_keypair("carl", password="123")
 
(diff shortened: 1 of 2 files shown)


layout-creation.md
```diff
@@ -0,0 +1,99 @@
+# Layout Creation Example
+
+The following Python snippet shows how to use in-toto to create a software
+supply chain layout like the one that is used for the in-toto demo.
```
vladimir-v-diaz: Can you also use a CLI to create a software supply layout?  I'm not sure from this opening paragraph if the example is the only way, the preferred, or an optional way to create a software supply layout.

layout-creation.md
```diff
@@ -0,0 +1,99 @@
...
+from securesystemslib.interface import (generate_and_write_rsa_keypair,
+    import_rsa_privatekey_from_file)
+from in_toto.models.layout import Layout, Step, Inspection
+from in_toto.models.metadata import Metablock
+
+# Create RSA private and public keys
+
+# Alice is the project owner and her private key is used to sign the layout
+alice_path = generate_and_write_rsa_keypair("alice", password="123")
+alice_key = import_rsa_privatekey_from_file(alice_path, password="123")
```
vladimir-v-diaz: Do you need to save the keys to disk?  This can potentially be reduced to one line by just generating the keys in memory.
For example:
```Python
securesystemslib.keys.generate_rsa_key(bits=3072, scheme='rsassa-pss-sha256')
```

layout-creation.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Create layout steps
+
+# Each step describes a task that is required to be carried out in this supply
+# chain. A step must have a unique name to associate the related link metadata
+# (i.e. the evidence that the step was carried out). Additionally, each step
+# should list rules, about the present related files before and after the step
+# was carried out. A steps pubkeys field lists the keyids of functionaries
+# authorized to perform the step.
+step_clone = Step(name="clone")
```
vladimir-v-diaz: The comment does a good job of explaining the meaning of an in-toto step.  However, I'm not sure what the the example step is supposed to represent.  I suspect its related to a git clone, but I don't know for certain from just reading the comment.

Consider reiterating, in the opening paragraph, the software supply chain of this example.  What are the expected steps that the example is going to create?

layout-creation.md
```diff
@@ -0,0 +1,99 @@
...
+step_clone.set_expected_command_from_string(
+    "git clone https://github.com/in-toto/demo-project.git")
+step_clone.add_product_rule_from_string("CREATE demo-project/foo.py")
+step_clone.add_product_rule_from_string("DISALLOW *")
```
vladimir-v-diaz: non-blocker:
I'm curious, are there multiple ways of adding a product rule?  From String?  From dict?

If a user can only add a product rule from a string, it might be cleaner to remove the `from_string` from the method name.

```Python
step_clone.add_product_rule("CREATE demo-project/foo.py")
```

layout-creation.md
```diff
@@ -0,0 +1,99 @@
...
+step_clone.pubkeys = [bob_pubkey["keyid"]]
+
+step_update = Step(name="update-version")
+step_update.add_material_rule_from_string(
+    "MATCH demo-project/* WITH PRODUCTS FROM clone")
+step_update.add_material_rule_from_string("DISALLOW *")
+step_update.add_product_rule_from_string("ALLOW demo-project/foo.py")
+step_update.add_product_rule_from_string("DISALLOW *")
+step_update.pubkeys = [bob_pubkey["keyid"]]
```
vladimir-v-diaz: Maybe state in a comment which steps Bob and Carl are trusted to sign.  Bob = "update-version" and "clone", Carl = "package".

Otherwise, the reader has to concentrate on lines 53, 61, and 71 to determine which functionary is responsible for which step.

layout-creation.md
```diff
@@ -0,0 +1,99 @@
...
+step_clone.add_product_rule_from_string("CREATE demo-project/foo.py")
+step_clone.add_product_rule_from_string("DISALLOW *")
+step_clone.pubkeys = [bob_pubkey["keyid"]]
+
+step_update = Step(name="update-version")
+step_update.add_material_rule_from_string(
+    "MATCH demo-project/* WITH PRODUCTS FROM clone")
+step_update.add_material_rule_from_string("DISALLOW *")
+step_update.add_product_rule_from_string("ALLOW demo-project/foo.py")
+step_update.add_product_rule_from_string("DISALLOW *")
```
vladimir-v-diaz, line 109: > step_update.add_product_rule_from_string("ALLOW demo-project/foo.py")
> step_update.add_product_rule_from_string("DISALLOW *")

These two rules seem to contradict each other.  I personally find it easier to grok `DISALLOW * EXCEPT "demo-project/foo.py"` or `ALLOW * EXCEPT "foo"`.  However, I'm not familiar with in-toto's rules and why things are designed this way.

Is there a reason why `DISALLOW *` isn't the default case/behavior?

layout-creation.md
```diff
@@ -0,0 +1,99 @@
...
+step_package.add_product_rule_from_string("DISALLOW *")
+step_package.pubkeys = [carl_pubkey["keyid"]]
+
+
+# Create inspection
+
+# Inspections are commands that are executed upon in-toto final product
+# verification. In this case, we define an inspection that expands the final
+# product and verifies that the contents of the archive match with what was
+# put into the archive.
```
vladimir-v-diaz: This explanation of the inspection is good.  A similar explanation would be nice for the others.

layout-creation.md
```diff
@@ -0,0 +1,99 @@
...
+# Add steps and inspections to layout
+layout.steps = [step_clone, step_update, step_package]
+layout.inspect = [inspection]
+
+
+# Eventually the layout gets wrapped in a generic in-toto metablock, which
+# provides functions to sign the metadata contents and write them to a file.
+metablock = Metablock(signed=layout)
+metablock.sign(alice_key)
+metablock.dump("root.layout")
```
vladimir-v-diaz: Do the other functionaries sign their own Metablock?  Is there a document that covers that?

vladimir-v-diaz (commented):
My review of pr #182.

layout-creation.md
lukpueh: This document explains the usage of some of the new layout creation convenience methods introduced with #181.

Alternatively you can create a layout by directly assigning the required values to the metadata objects. Here's the difference

```python
# Manipulate objects directly
layout.expires = "2018-03-08T13:55:21Z"
step.expected_products.append(["CREATE", "foo"])

# Use convenience methods
layout.set_relative_expiration(month=1)
step.add_product_rule_from_string("CREATE foo")
```
Independently, you can also create a layout manually using a text editor.

I'll try to rephrase the opening paragraph.

layout-creation.md
lukpueh: You will need to keep them around.
Functionaries need their private keys to sign link metadata and the owner will want to keep the private key to resign the layout and the public key for verification.

I'll rephrase

layout-creation.md
lukpueh: I'd like to keep redundancy to minimum (the demo's README already explains the supply chain in detail). But I will add a little bit more information/ pointers.

layout-creation.md
lukpueh: Per specification each artifact rule is stored as a list, e.g. `["CREATE", "foo"]`.  The `*from_string` method is for convenience.

layout-creation.md
lukpueh: See my reply in https://github.com/in-toto/in-toto/pull/182#discussion_r167036829.
But yes, I'll add more information about the supply chain.

layout-creation.md
lukpueh: (1) We can consider that. Care for opening an issue, maybe on the [specification repo](https://github.com/in-toto/docs)?
(2) We used to have implicit `DISALLOW *` but received a request to change it, see #131 and https://github.com/in-toto/docs/issues/4

layout-creation.md
lukpueh: Functionaries don't need to sign their metadata (links) explicitly. That happens automatically when using `in-toto-run`.

I'll rephrase the comments here, and maybe in the beginning where the keys are created.

lukpueh:
Thanks for the review, @vladimir-v-diaz. I addressed your comments and will merge.

PR #183, opened 2018-02-01
Refactor logging

### LINKED ISSUE #117: Revisit verbosity (stdout/stderr)

lukpueh:
**Fixes issue #**:
Closes #117 

The goal of this PR is to make better use of Python's `logging` module for debugging as well as user feedback.

Before, in-toto used wrappers for logging functions directly on the `in_toto.log` module. Now these wrappers are removed and instead we register a base `in_toto` logger, using a custom `logging.Logger` subclass with an enhanced error method and a convenience method to set the logging level, e.g. in command line interfaces that accept a `verbose` and/or `quiet` option, which are also added in this PR.

See  in_toto.log's doc header for detailed usage instructions.

The PR also removes the `LOG_LEVEL` setting and adds a `DEBUG` setting (boolean) which affects the format and log level of the in-toto base logger.

The PR also replaces some `warning` with `info` log statements. For user feedback we should always use log type info, because it is disabled by default when using the library directly, but enabled when using command line interfaces with the `verbose` option.

Furthermore the PR fixes test log inundation by
- replacing the base logger's  default `StreamHandler` with `NullHandler` when running tests to suppress all `logging` based output.
- adding global flag buffer flag to runtests' `TextTestRunner` to suppress non `logging` based output.
- adding a stderr Pipe to subprocess calls in `in_toto.gpg`, to suppress gpg warnings (we are not
interested in stderr (but stdout) to parse keys and stderr only floods the test log due to permission warnings related to the test gpg keychain.

requirements**:


in_toto/in_toto_mock.py
```diff
@@ -91,8 +96,8 @@ def main():
 
   args = parser.parse_args()
 
-  # Turn on all the `log.info()` in the library
-  log.logging.getLogger().setLevel(log.logging.INFO)
+  # Default to verbose
```
SantiagoTorres: I don't know how descriptive this comment is. I personally find it a little confusing.

in_toto/log.py
```diff
@@ -12,30 +12,118 @@
...
+  def setLevelVerboseOrQuiet(self, verbose, quiet):
+    """Convenience method to set the logger's verbosity level based on the
+    passed booleans verbose and quiet (useful for cli tools). """
+    if verbose:
+      self.setLevel(logging.INFO)
+
+    elif quiet:
+      # TODO: Is it enough to use logging.CRITICAL + 1 to suppress all output?
+      # A saver way would be to use a NullHandler or Filters.
+      self.setLevel(self.QUIET)
```
SantiagoTorres, line 105: I think this is "ok" imho.

tests/test_in_toto_record.py
```diff
@@ -105,11 +101,6 @@ def test_start_stop(self):
     self._test_cli_sys_exit(["stop"] + args + ["--products",
         self.test_artifact2, self.test_artifact2], 0)
 
-    # Start/stop recording verbosely
-    args = ["--step-name", "test4", "--key", self.key_path, "--verbose"]
-    self._test_cli_sys_exit(["start"] + args, 0)
-    self._test_cli_sys_exit(["stop"] + args, 0)
-
```
SantiagoTorres, line 112: Why did this test disappear?

tests/test_in_toto_run.py
```diff
@@ -38,8 +37,7 @@
 
 import in_toto.gpg.util
 
-# Suppress all the user feedback that we print using a base logger
-logging.getLogger().setLevel(logging.CRITICAL)
+
```
SantiagoTorres, line 40: I think this whitespace is one too many.

tests/test_settings.py
```diff
@@ -0,0 +1,28 @@
...
+import in_toto.settings
+
+
+class TestSettings(unittest.TestCase):
+  def test_debug_not_true(self):
+    """in_toto.settings.DEBUG should not be commited with True. """
+    self.assertFalse(in_toto.settings.DEBUG)
+
+if __name__ == "__main__":
+  unittest.main()
```
SantiagoTorres, line 28: What's the difference between `test_settings` and `test_user_settings`?

SantiagoTorres (approved):
LGTM, although I have some questions.

in_toto/in_toto_mock.py
lukpueh: I can rephrase, e.g.:
*in-toto-mock should not be used to secure the supply chain but only to try out in-toto-run, hence it gives verbose user feedback*

in_toto/log.py
lukpueh: :)

tests/test_in_toto_record.py
lukpueh: The test used to trigger an `if` statement in in_toto_record. That `if` statement was replaced by a logger method call `setLevelVerboseOrQuiet`, which I test separately in `test_log.py`.

Therefore, we are good in terms of coverage. In terms of interface testing, you are right, I could call the the cli tool with the verbose flag and test if it passes the right thing to `setLevelVerboseOrQuiet`. Not sure if necessary. What do you think?

tests/test_in_toto_run.py
lukpueh: This is where our code style guideline actually suggests [5 blank lines](https://github.com/secure-systems-lab/code-style-guidelines#blank-lines), which I find too much. I think 3 are a fair compromise...

tests/test_settings.py
lukpueh: One tests the `settings` module and the other one tests the `user_settings` module. The latter can be used to configure the former by loading environment variables and rcfiles.

tests/test_in_toto_run.py
SantiagoTorres: wow, TIL. Ok I guess...

tests/test_settings.py
SantiagoTorres: > One tests the settings module and the other one tests the user_settings module

I mean that much I could tell. I was imagining both of them are the same functionality, with `user_settings` being just a module/function to override the former and thus needed no conceptual separation. 

I don't think it's a deal at all though.

tests/test_in_toto_record.py
SantiagoTorres: I don't think it's necessary. I just wanted to make sure you didn't delete that test for the sake of testing or something akin to that.

lukpueh:
Thanks for the reviewing and the button clicking, @SantiagoTorres

PR #184, opened 2018-02-06
Refactor cli tools

lukpueh:
This PR cleans up the command line interface code, by
-  adding consistent and sensible return codes,
- enhancing custom and autogenerated argparse help messages,
- removing obsolete library wrappers,
- fixing a small argparse bug in in_toto_record,
- unifying cli unittests

Read commit messages for more infos about the changes.

requirements**:

### DIFF
--- a/in_toto/in_toto_keygen.py
+++ b/in_toto/in_toto_keygen.py
@@ -30,6 +30,12 @@
 
   python in_toto_keygen.py -p bob_keys 2048
 
+
+<Return Codes>
+  2 if an exception occurred during argument parsing
+  1 if an exception occurred
+  0 if no exception occurred
+
 """
 import sys
 import argparse
--- a/in_toto/in_toto_keygen.py
+++ b/in_toto/in_toto_keygen.py
@@ -107,7 +107,7 @@ def main():
       sys.exit(0)
 
   except Exception as e:
-    log.error('The following error occurred - {}'.format(e))
+    log.error("(in-toto-keygen) {0}: {1}".format(type(e).__name__, e))
     sys.exit(1)
 
 if __name__ == "__main__":
--- a/in_toto/in_toto_keygen.py
+++ b/in_toto/in_toto_keygen.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 """
 <Program Name>
   in_toto_keygen.py
--- a/in_toto/in_toto_keygen.py
+++ b/in_toto/in_toto_keygen.py
@@ -20,7 +20,7 @@
   generated. By default it is set as 3072.
 
   General Usage:
-  python in_toto_keygen.py [-p] <filename> [bits]
+  in-toto-keygen [-p] <filename> [bits]
 
   Example Usage:
   Suppose Bob wants to create the keys of size 2048 bits and dump them with
@@ -29,7 +29,7 @@
   would be encrypted and dumped as "bob_keys" and public key would be dumped
   as "bob_keys.pub". Bob will use the following command:
 
-  python in_toto_keygen.py -p bob_keys 2048
+  in-toto-keygen -p bob_keys 2048
 
 
 <Return Codes>
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -67,7 +67,7 @@ def main():
 
   # FIXME: Do we limit the allowed characters for the name?
   named_args.add_argument("-n", "--name", type=str, required=True,
-      help="Unique name for link metadata", metavar="<unique step name>")
+      help="Unique name for link metadata", metavar="<name>")
 
   # FIXME: This is not yet ideal.
   # What should we do with tokens like > or ; ?
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -31,8 +31,15 @@
   in-toto-mock --name write-code -- touch foo.py
   ```
 
-"""
+<Arguments>
+  name:
+          A unique name to relate mock link metadata with a step defined
+          in the layout.
+  command:
+          A list where the first element is a command and the remaining
+          elements are arguments passed to that command.
 
+"""
 import sys
 import argparse
 import logging
@@ -42,34 +49,6 @@ import in_toto.runlib
 log = logging.getLogger("in_toto")
 
 
-def in_toto_mock(name, link_cmd_args):
-  """
-  <Purpose>
-    Calls runlib.in_toto_mock with name and link_cmd_args.
-
-  <Arguments>
-    name:
-            A unique name to relate mock link metadata with a step defined
-            in the layout.
-    link_cmd_args:
-            A list where the first element is a command and the remaining
-            elements are arguments passed to that command.
-
-  <Exceptions>
-    SystemExit if any exception occurs
-
-  <Side Effects>
-    Calls sys.exit(1) if an exception is raised
-
-  <Returns>
-    None.
-  """
-
-  try:
-    in_toto.runlib.in_toto_mock(name, link_cmd_args)
-  except Exception as e:
-    log.error("in toto mock - {}".format(e))
-    sys.exit(1)
 
 def main():
   """Parse arguments and call in_toto_mock. """
@@ -100,7 +79,15 @@ def main():
   # TRY out in-toto-run, with fewer command lines args and max. user feedback.
   log.setLevel(logging.INFO)
 
-  in_toto_mock(args.name, args.link_cmd)
+  try:
+    in_toto.runlib.in_toto_mock(args.name, args.link_cmd)
+
+  except Exception as e:
+    log.error("in toto mock - {}".format(e))
+    sys.exit(1)
+
+  sys.exit(0)
+
 
 if __name__ == "__main__":
   main()
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -54,23 +54,18 @@ def main():
   """Parse arguments and call in_toto_mock. """
   parser = argparse.ArgumentParser(
       description="Executes link command and records metadata")
-  # Whitespace padding to align with program name
-  lpad = (len(parser.prog) + 1) * " "
 
-  parser.usage = ("\n"
-      "%(prog)s  --name <unique step name>\n{0}"
-               " -- <cmd> [args]\n\n"
-               .format(lpad))
+  parser.usage = "%(prog)s [-h] --name <unique step name> -- <command> [args]"
 
-  in_toto_args = parser.add_argument_group("in-toto options")
+  named_args = parser.add_argument_group("required named arguments")
 
   # FIXME: Do we limit the allowed characters for the name?
-  in_toto_args.add_argument("-n", "--name", type=str, required=True,
-      help="Unique name for link metadata")
+  named_args.add_argument("-n", "--name", type=str, required=True,
+      help="Unique name for link metadata", metavar="<unique step name>")
 
   # FIXME: This is not yet ideal.
   # What should we do with tokens like > or ; ?
-  in_toto_args.add_argument("link_cmd", nargs="+",
+  parser.add_argument("link_cmd", nargs="+", metavar="<command>",
     help="Link command to be executed with options and arguments")
 
   args = parser.parse_args()
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -13,37 +13,40 @@
   See LICENSE for licensing information.
 
 <Purpose>
-  Stripped down variant of in-toto-run command that can be used to mock metadata
-  generation of in-toto-run, without the need to specify a key and knowing all
-  the command line arguments. Generated MockLink is unsigned and includes working
-  directory, byproducts and used current directory as material and products.
-
-  Provides a command line interface which takes any link command of the software
-  supply chain as input and generates mock in_toto metadata.
-
-  in_toto_mock options are separated from the command to be executed by
-  a double dash.
-
-  The implementation of the tasks can be found in runlib.
-
-  Example Usage
-  ```
-  in-toto-mock --name write-code -- touch foo.py
-  ```
-
-<Arguments>
-  name:
-          A unique name to relate mock link metadata with a step defined
-          in the layout.
-  command:
-          A list where the first element is a command and the remaining
-          elements are arguments passed to that command.
-
+  Provides a command line interface for runlib.in_toto_mock.
 
 <Return Codes>
   2 if an exception occurred during argument parsing
   1 if an exception occurred
   0 if no exception occurred
+
+<Help>
+usage: in-toto-mock [-h] --name <name> -- <command> [args]
+
+A stripped down variant of 'in-toto-run' that can be used to create unsigned
+link metadata for the passed command, recording all files in the current
+working directory as materials and products.
+
+This command should not be used to secure the supply chain but only to try
+out the 'in-toto-run' command.
+
+positional arguments:
+  <command>             Command to be executed with options and arguments,
+                        separated from 'in-toto-mock' options by double dash
+                        '--'.
+
+optional arguments:
+  -h, --help            show this help message and exit
+
+required named arguments:
+  -n <name>, --name <name>
+                        Name used to associate the resulting link metadata
+                        with the corresponding step defined in an in-toto
+                        layout.
+
+examples:
+  Generate link metadata 'foo' for the activity of creating file 'bar'.
+    in-toto-mock --name foo -- touch bar
 
 """
 import sys
@@ -59,20 +62,38 @@ log = logging.getLogger("in_toto")
 def main():
   """Parse arguments and call in_toto_mock. """
   parser = argparse.ArgumentParser(
-      description="Executes link command and records metadata")
+     formatter_class=argparse.RawDescriptionHelpFormatter,
+      description="""
+A stripped down variant of 'in-toto-run' that can be used to create unsigned
+link metadata for the passed command, recording all files in the current
+working directory as materials and products.
 
-  parser.usage = "%(prog)s [-h] --name <unique step name> -- <command> [args]"
+This command should not be used to secure the supply chain but only to try
+out the 'in-toto-run' command.""")
+
+  parser.usage = "%(prog)s [-h] --name <name> -- <command> [args]"
+
+  parser.epilog = """
+examples:
+  Generate link metadata 'foo' for the activity of creating file 'bar'.
+    {prog} --name foo -- touch bar
+
+""".format(prog=parser.prog)
 
   named_args = parser.add_argument_group("required named arguments")
 
   # FIXME: Do we limit the allowed characters for the name?
   named_args.add_argument("-n", "--name", type=str, required=True,
-      help="Unique name for link metadata", metavar="<name>")
+      metavar="<name>", help=(
+      "Name used to associate the resulting link metadata with the"
+      " corresponding step defined in an in-toto layout."))
 
   # FIXME: This is not yet ideal.
   # What should we do with tokens like > or ; ?
   parser.add_argument("link_cmd", nargs="+", metavar="<command>",
-    help="Link command to be executed with options and arguments")
+      help=(
+      "Command to be executed with options and arguments, separated from"
+      " 'in-toto-mock' options by double dash '--'."))
 
   args = parser.parse_args()
 
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -76,8 +76,8 @@ def main():
 
   args = parser.parse_args()
 
-  # in-toto-mock should NOT be used to secure the supply chain but only to
-  # TRY out in-toto-run, with fewer command lines args and max. user feedback.
+  # in-toto-mock should not be used to secure the supply chain but only to try
+  # out in-toto-run with max. user feedback, hence we set a verbose log level
   log.setLevel(logging.INFO)
 
   try:
--- a/in_toto/in_toto_mock.py
+++ b/in_toto/in_toto_mock.py
@@ -84,7 +84,7 @@ def main():
     in_toto.runlib.in_toto_mock(args.name, args.link_cmd)
 
   except Exception as e:
-    log.error("in toto mock - {}".format(e))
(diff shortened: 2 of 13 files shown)


in_toto/in_toto_keygen.py
```diff
@@ -30,6 +30,12 @@
 
   python in_toto_keygen.py -p bob_keys 2048
 
+
+<Return Codes>
+  2 if an exception occurred during argument parsing
+  1 if an exception occurred
+  0 if no exception occurred
+
```
vladimir-v-diaz: Apart from this docstring, are these return codes documented anywhere?  How do you inform users, who will not necessarily read the docstring, of these return codes?

in_toto/in_toto_keygen.py
```diff
@@ -30,6 +30,12 @@
 
   python in_toto_keygen.py -p bob_keys 2048
```
vladimir-v-diaz: Do you plan to support additional algorithms with this CLI?  It might be good to state this somewhere.

vladimir-v-diaz: If `in_toto_keygen.py` is added as a global script, you can simplify this example as:
```Bash
$ in_toto_keygen.py -p bob_keys 2048
```

vladimir-v-diaz (commented):
My review of `in_toto_keygen.py`.

Some of the wrapped argument strings should use 4 spaces of indentation.

in_toto/in_toto_mock.py
```diff
@@ -42,56 +55,23 @@
 log = logging.getLogger("in_toto")
```
vladimir-v-diaz, line 59: Is there a reason why you don't use the "a.b.c.." convention when specifying the logger name?
https://docs.python.org/2/library/logging.html#logging.getLogger

in_toto/in_toto_mock.py
```diff
@@ -31,8 +31,21 @@
   in-toto-mock --name write-code -- touch foo.py
```
vladimir-v-diaz: I think you need to specify the `.py` extension with `in-toto-mock`. Actually, should that be `in_toto_mock.py`!?

Also, is the `--` by itself needed/correct in `-- touch foo.py`?

I was expecting it to be: `$ in-toto-mock.py --name write-code touch foo.py`

in_toto/in_toto_mock.py
```diff
@@ -31,8 +31,21 @@
...
   ```
 
-"""
+<Arguments>
+  name:
+          A unique name to relate mock link metadata with a step defined
+          in the layout.
+  command:
```
vladimir-v-diaz: The `command` argument in this docstring might be confusing to some users.  On the one hand, `--name` is an actual command-line argument, but `--command` is not.  That is, `$ in_toto_mock.py --name foo touch bar.log` is the expected format, where `touch bar.log` are positional arguments.

in_toto/in_toto_mock.py
```diff
@@ -100,7 +80,15 @@ def main():
   # TRY out in-toto-run, with fewer command lines args and max. user feedback.
```
vladimir-v-diaz: I don't understand why this comment is here.  Should it go in the module docstring?

in_toto/in_toto_mock.py
```diff
@@ -100,7 +80,15 @@ def main():
...
   log.setLevel(logging.INFO)
 
-  in_toto_mock(args.name, args.link_cmd)
+  try:
+    in_toto.runlib.in_toto_mock(args.name, args.link_cmd)
+
+  except Exception as e:
+    log.error("in toto mock - {}".format(e))
```
vladimir-v-diaz: Why is it written as `in toto mock` and not `in_toto_mock.py` or `in-toto mock`?

in_toto/in_toto_mock.py
```diff
@@ -31,8 +31,21 @@
...
+  name:
+          A unique name to relate mock link metadata with a step defined
+          in the layout.
+  command:
+          A list where the first element is a command and the remaining
+          elements are arguments passed to that command.
+
+
+<Return Codes>
+  2 if an exception occurred during argument parsing
```
vladimir-v-diaz: Where is `2` returned in the code?

vladimir-v-diaz (commented):
Here is my review of `in_toto_mock.py`.

in_toto/in_toto_record.py
```diff
@@ -76,6 +82,7 @@ def main():
   # The subparsers inherit the arguments from the parent parser
   parent_parser = argparse.ArgumentParser(add_help=False)
   subparsers = parser.add_subparsers(dest="command")
+  subparsers.required = True
```
vladimir-v-diaz: Can you comment on why this `required = True` is... required?

in_toto/in_toto_run.py
```diff
@@ -27,6 +27,40 @@
       -- vi foo.py
   ```
 
+<Arguments>
+  step_name:
+          A unique name to relate link metadata with a step defined in the
+          layout.
+  material_list:
```
vladimir-v-diaz: Should `material_list` be `materials`?  Please verify the names of the other command-line options, like `products` instead of `product_list`.

in_toto/in_toto_run.py
```diff
@@ -40,106 +74,47 @@
...
-               "[--materials <filepath>[ <filepath> ...]]\n{0}"
-               "[--products <filepath>[ <filepath> ...]]\n{0}"
-               "[--record-streams]\n{0}"
-               "[--no-command]\n{0}"
-               "[--verbose | --quiet] -- <cmd> [args]\n\n"
-               .format(lpad))
-
-  in_toto_args = parser.add_argument_group("in-toto options")
+
+  parser.usage = ("%(prog)s <named arguments> [optional arguments]"
```
vladimir-v-diaz: Is is better for the usage string to specify the required named arguments?

`(%s(prog)s --step-name foo --products ...)`

in_toto/in_toto_run.py
```diff
@@ -40,106 +74,47 @@
...
-  in_toto_args.add_argument("-n", "--step-name", type=str, required=True,
-      help="Unique name for link metadata")
+  named_args.add_argument("-n", "--step-name", type=str, required=True,
+      help="Unique name for link metadata", metavar="<unique step name>")
 
-  in_toto_args.add_argument("-m", "--materials", type=str, required=False,
-      nargs='+', help="Files to record before link command execution")
+  parser.add_argument("-m", "--materials", type=str, required=False,
+      nargs='+', help="Files to record before link command execution",
+      metavar="<material path>")
```
vladimir-v-diaz: Should you instead use `<material paths>` (note: plural) to indicate that more than one path can be given?

in_toto/in_toto_run.py
```diff
@@ -40,106 +74,47 @@
...
+      help="Path to GPG keyring (if not set the default keyring is used)",
+      metavar="<gpg keyring path>")
 
-  in_toto_args.add_argument("-b", "--record-streams",
+  parser.add_argument("-b", "--record-streams",
       help="If set redirects stdout/stderr and stores to link metadata",
       dest="record_streams", default=False, action="store_true")
 
-  in_toto_args.add_argument("-x", "--no-command",
+  parser.add_argument("-x", "--no-command",
```
vladimir-v-diaz: I think this command-line argument is not documented in the module's docstring.

vladimir-v-diaz (commented):
Here is my review of `in_toto_run.py`

in_toto/in_toto_sign.py
```diff
@@ -302,12 +302,12 @@ def main():
...
         (args.gpg != None and len(args.gpg) > 1)):
       parser.print_help()
-      parser.exit(2, "too many arguments: {} Hence signing Link metadata"
+      parser.error("too many arguments: {} Hence signing Link metadata"
           " with multiple keys is not allowed.".format(link_error_message))
 
     if args.append:
       parser.print_help()
-      parser.exit(2, "wrong arguments: {}. Hence adding signatures to"
+      parser.error("wrong arguments: {}. Hence adding signatures to"
```
vladimir-v-diaz, line 395: So the entire error message is:

```
("Link metadata is associated with a single functionary and is usually namespaced accordingly: '<name>.<keyid>.link'.")
Hence signing Link metadata with multiple keys is not allowed.
```

What if a functionary wants to use multiple keys?

vladimir-v-diaz (commented):
Comment for `in_toto_sign.py`

in_toto/in_toto_sign.py
vladimir-v-diaz: For instance, one functionary wants the ability to sign with one of two keys under their control.  One key is stored on their laptop, and a different, and also trusted, key is on their Yubikey.

in_toto/in_toto_verify.py
```diff
@@ -16,12 +16,16 @@
   Provides a command line interface that wraps the verification of
   in_toto final product.
```
vladimir-v-diaz: I've seen this a few times now.  Should `in_toto` be `in-toto`?

in_toto/in_toto_verify.py
```diff
@@ -36,6 +40,27 @@
       --gpg 8465A1E2E0FB2B40ADB2478E18FB3F537E0C8A17 --gpg-home ~/.gnupg
   ```
 
+<Arguments>
+  layout:
+          Path to layout metadata file that is being verified.
+
+  layout keys:
+          List of path(s) to project owner public key(s), used to verify the
+          root layout's signature(s).
```
vladimir-v-diaz: Can `in-toto-verify` verify any layout file?  I suppose it's referring to the root example above, but this comment makes it seem like only the root layout can be verified.

in_toto/in_toto_verify.py
```diff
@@ -45,104 +70,34 @@
...
+  parser.usage = "%(prog)s <named arguments> [optional arguments]"
 
-  parser.usage = ("\n"
-      "%(prog)s --layout <layout path>\n{0}"
-               "{{--layout-keys <filepath>[ <filepath> ...], "
-               " --gpg <keyid> [ <keyid> ...]}} \n{0}"
-               "[--gpg-home <path to gpg keyring>]\n{0}"
-               "[--verbose | --quiet]\n\n"
-               .format(lpad))
+  named_args = parser.add_argument_group("required named arguments")
```
vladimir-v-diaz: `<required arguments>` and `required_args` might be clearer.

in_toto/in_toto_verify.py
```diff
@@ -160,10 +114,34 @@ def main():
   # Note: Passing both at the same time is possible.
   if (args.layout_keys == None) and (args.gpg == None):
     parser.print_help()
-    parser.exit(2, "wrong arguments: specify at least one of "
-      " `--layout-keys path [path ...]` or `--gpg id [id ...]`")
+    parser.error("wrong arguments: specify at least one of"
```
vladimir-v-diaz, line 179: The Note: above states that passing both at the same time is possible, however, this parser.error implies otherwise.

in_toto/in_toto_verify.py
```diff
@@ -45,104 +70,34 @@
...
-               .format(lpad))
+  named_args = parser.add_argument_group("required named arguments")
 
-  in_toto_args = parser.add_argument_group("in-toto options")
+  named_args.add_argument("-l", "--layout", type=str, required=True,
+      help="Root layout to use for verification", metavar="<layout path>")
 
-  in_toto_args.add_argument("-l", "--layout", type=str, required=True,
-      help="Root layout to use for verification")
+  named_args.add_argument("-k", "--layout-keys", type=str,
```
vladimir-v-diaz: This `add_argument()` call does not set required = True.  So is --layout-keys optional?  The `named_args` group indicates that they are required arguments.

in_toto/in_toto_verify.py
```diff
@@ -45,104 +70,34 @@
...
+      nargs="+", help="Key(s) to verify root layout signature",
```
vladimir-v-diaz: Is there only a root layout?  In other words, are there no general layouts?  Are signatures stored in something called a root layout?

vladimir-v-diaz (commented):
My review of `in-toto-verify.py`.

tests/common.py
```diff
@@ -0,0 +1,79 @@
...
+
+<Started>
+  Feb 6, 2018
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Common code for in-toto unittests, import like so:
+  `import tests.common`
```
vladimir-v-diaz, line 17: It might help if you explained how this module should be used.

Namely, that test cases inherit from CliTestCase:
```
class MyTestCase(tests.common.CliTestCase):
```

Do you also want to mention the provided convenience/mocking methods?

vladimir-v-diaz (commented):
My review of `tests/common.py`.

vladimir-v-diaz:
@lukpueh 
Okay, finished reviewing all of #184.

in_toto/in_toto_keygen.py
lukpueh: I don't think it's necessary, since this follows convention.

in_toto/in_toto_keygen.py
lukpueh: Not at the moment. This tool was initially developed to allow quick `securesystemslib` key pair creation from the command line with out having to open a Python interpreter and import modules. The intended use case was in-toto layout creation web tool, where it's reputability was challenged (https://github.com/in-toto/layout-web-tool/issues/17).

We can discuss if we want to extend this tool and maybe move it over to `securesystemslib` what do you think?

lukpueh: True, will change.

in_toto/in_toto_mock.py
lukpueh: Yes, there is and it is documented in [`in_toto.log`](https://github.com/in-toto/in-toto/blob/develop/in_toto/log.py#L35-L66) as suggested by the comment.

in_toto/in_toto_mock.py
lukpueh: Nope, [`in-toto-mock` works fine](https://github.com/in-toto/in-toto/blob/develop/setup.py#L80-L85) when in-toto is installed.

lukpueh: Per *nix and [argparse](https://docs.python.org/dev/library/argparse.html#arguments-containing) convention is used to separate optional arguments from positional arguments.

in_toto/in_toto_keygen.py
vladimir-v-diaz: Which convention?

I think documenting return/status codes is standard practice.

Examples:
[HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)
[FTP return codes](https://en.wikipedia.org/wiki/List_of_FTP_server_return_codes)
[Pylint exit codes](https://docs.pylint.org/en/1.6.0/run.html#exit-codes)

Since you only really return 1-2 for exceptions, perhaps you can say that the script returns a non-zero value for any exception.

in_toto/in_toto_mock.py
lukpueh: Well technically, both are arguments, one *optional* (which I find confusing because it's actually required, see b19b17fdd4865896f70fb4f4ee8f8ebcb081d11d) and one *positional*.

Also, I think it's fine, because as you mentioned above, the user will probably not look at the docstring but rather at the command's help message, which should make the usage clear:
```
$ in-toto-mock -h
usage: in-toto-mock [-h] --name <unique step name> -- <command> [args]

Executes link command and records metadata

positional arguments:
  <command>             Link command to be executed with options and arguments

optional arguments:
  -h, --help            show this help message and exit

required named arguments:
  -n <unique step name>, --name <unique step name>
                        Unique name for link metadata
```

in_toto/in_toto_keygen.py
lukpueh: Shell command exit codes follow the convention of returning non-zero in case of failure and zero in case of success.
However, documentation is provided in the form of above docstring.

in_toto/in_toto_mock.py
lukpueh: Santiago requested to clarify the comment I had before an the PR that modified this line:
https://github.com/in-toto/in-toto/pull/183#discussion_r166432991.

The *"max. user feedback"* is the crucial part, that should explain why I set the log level to `INFO`.

in_toto/in_toto_mock.py
lukpueh: Good question. I'll make uniform with `log.error`s in other command line tools.

in_toto/in_toto_mock.py
lukpueh: By `ArgumentParser.error`, which can called by `argparse` internally when parsing errors, and I call it explicitly too (see 4ae540139e0831350ea49782ecd956a2ad3480d8).

in_toto/in_toto_record.py
lukpueh: I hint at it in the related commit message (d74b274177c1ea0dc2ff71f8fceb9138e7d93af2), but I can add a comment too.

in_toto/in_toto_run.py
lukpueh: Yes, I can change this. We actually don't have a guideline for documenting arpgarse arguments. 

There are several ways:
 - an *abstract* name for the argument (that's what i did),
 - the argument name(s) the command line user has to use (e.g. `-n/--step-name`), which only works for positional arguments,
 - the variable where argparse stores argument (e.g. `step_name`),
 - a combination of above

What do you suggest?

in_toto/in_toto_run.py
lukpueh: That's what I had before, but it was messier to maintain. And the user can always use `--help` for a detailed description.

in_toto/in_toto_run.py
lukpueh: No, given that metavar, argparse will create an accurate help message: 
```
-m <material path> [<material path> ...], --materials <material path> [<material path> ...]
```

in_toto/in_toto_run.py
lukpueh: Good eyes, I will add it.

in_toto/in_toto_sign.py
lukpueh: This behavior  was already topic for discussion in https://github.com/in-toto/in-toto/pull/125 :)

The specification currently does not explicitly prohibit signing one link with multiple keys, however due to the naming convention `<related step name>.<SIGNING KEY PREFIX>.link` and the way final product verification will load links for steps, it does make sense and that's why this utility prohibits it.

Note that this doesn't mean that a functionary can't have multiple authorized keys to sign a link. All the public keys that a link can be signed with need to be embedded in the layout's `keys` section, and in to authorize them for a specific step, the keyid's must be in the step's `pubkeys` list.

in_toto/in_toto_run.py
lukpueh: Another way would be to just copy-paste the output of the command's help message.
I think I'll do that, and I will probably extend the help messages a little bit.

in_toto/in_toto_verify.py
lukpueh: Well, the project is called *in-toto*`, but due to Python naming restriction, the package is called `in_toto`. In above case it should be *in-toto*. I'll grep through the project and will change where necessary.

in_toto/in_toto_verify.py
lukpueh: Will change.

in_toto/in_toto_verify.py
lukpueh: Ok, will change.

in_toto/in_toto_verify.py
lukpueh: It says, at *"least one of keys [...] or gpg [...]"*, which IMHO doesn't imply that they are mutually exclusive.

in_toto/in_toto_verify.py
lukpueh: This is due to a shortcoming of `argparse`. There's no way of specifying a required (non mutual exclusive) group. If I'd make each argument required, the user couldn't specify only one of them, which is what I want.
I'll update the help message so that it's clear that only one of layout keys or gpg ids is okay.

in_toto/in_toto_verify.py
lukpueh: We have not exact definition of what a *root layout* is as opposed to a *layout*.

In my understanding, the root layout of a project specifies the entire supply chain for that project, however steps of the supply chain can be delegated to sublayouts (with arbitrary nesting). As a consequence, a root layout of a project can itself be a sublayout of another project.

In order to verify the supply chain of the final product of a project, you need to pass that project's root layout to `in-toto-verify`.

tests/common.py
lukpueh: I think that the class and method docstrings of `CLITestCase` describes its usage well enough.

in_toto/in_toto_run.py
vladimir-v-diaz: I suggest something like the following:

```
<Arguments>
    -f/--foo (required argument):
        blah blah.
    -b/--bar (optional argument):
        bam bam.
    step-name (positional argument):
        Not a nickname, but a stepname.
```

You can also copy-paste the output of the command's help message.

in_toto/in_toto_mock.py
vladimir-v-diaz: Oh, it wasn't obvious to me how the comment relates to `log.setLevel(...)`.  Consider explicitly stating the connection, as you proposed in #183.

"in-toto-mock should not be used to secure the supply chain but only to try out in-toto-run, **hence it gives verbose user feedback**"

in_toto/in_toto_verify.py
lukpueh: Actually, I think I'll leave it as is. Because the `<required named arguments>` section doesn't list required positional arguments, which I would expect if it were called `<required arguments>`.
As for the variable name, `required_named_args` would probably be the best, but that seems a little long.

lukpueh:
Thanks for the detailed review, @vladimir-v-diaz. 
I addressed all your comments and also improved the help messages for all tools (except for keygen) massively. Please let me know what you think.

lukpueh:
I gave the help texts another pass and fixed some spacing and typos. I'd say I addressed all of @vladimir-v-diaz's comments. Thanks for reviewing! Merging...

PR #187, opened 2018-02-15
FIX: gpg: s/pgp+SHA1/pgp+SHA2/

SantiagoTorres:
The code used to identify the pgp-compliant algorithms as PGP+sha1. In reality,
we are using sha256 as the hash algorithm. Change the default string to
pgp+SHA2 to properly document what are the methods used for hashing.

In addition, the string has been moved to a constant within gpg.formats.py so
it can be easily replaced in the future.

### DIFF
--- a/in_toto/gpg/formats.py
+++ b/in_toto/gpg/formats.py
@@ -58,6 +58,9 @@ import securesystemslib.schema as ssl_schema
 import securesystemslib.formats as ssl_formats
 
 
+GPG_HASH_ALGORITHM_STRING = "pgp+SHA2"
+GPG_PUBKEY_METHOD_STRING = "pgp+rsa-pkcsv1.5"
+
 RSA_PUBKEYVAL_SCHEMA = ssl_schema.Object(
   object_name = "RSA_PUBKEYVAL_SCHEMA",
   e = ssl_schema.AnyString(),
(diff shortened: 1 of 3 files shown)


in_toto/gpg/formats.py
```diff
@@ -58,6 +58,9 @@
 import securesystemslib.formats as ssl_formats
 
 
+GPG_HASH_ALGORITHM_STRING = "pgp+SHA2"
+GPG_PUBKEY_METHOD_STRING = "pgp+rsa-pkcsv1.5"
```
lukpueh: Mind also using this constant [here in `gpg/constants.py`](https://github.com/SantiagoTorres/in-toto/blob/f3fcabb4c96f1fceaa41e93b5690e86c12746978/in_toto/gpg/constants.py#L38)?

lukpueh: Since you're at it, maybe you can also create a constant for `pgp+dsa-fips-180-2` and use it below in `DSA_PUBKEY_SCHEMA` and as well [here in `gpg/constants.py`](https://github.com/SantiagoTorres/in-toto/blob/f3fcabb4c96f1fceaa41e93b5690e86c12746978/in_toto/gpg/constants.py#L43).

You'd probably have to use a different name for both then, e.g. `RSA_SIGNATURE_METHOD_STRING` and `DSA_SIGNATURE_METHOD_STRING`, what do you think?

in_toto/gpg/formats.py
```diff
@@ -58,6 +58,9 @@
 import securesystemslib.formats as ssl_formats
 
 
+GPG_HASH_ALGORITHM_STRING = "pgp+SHA2"
```
lukpueh, line 61: Please also fix in [docstring](https://github.com/SantiagoTorres/in-toto/blob/f3fcabb4c96f1fceaa41e93b5690e86c12746978/in_toto/gpg/formats.py#L33).

lukpueh (changes requested):
Thanks for the PR, @SantiagoTorres! I have a few minor comments, please fix and it's good to go.

in_toto/gpg/formats.py
SantiagoTorres: That may probably be a good idea. As for putting them on `constants.py` it is not trivial, as far as I understand that'd create a circular dependency...

in_toto/gpg/formats.py
SantiagoTorres: fixed.

in_toto/gpg/formats.py
lukpueh: Importing and assigning the *constants* defined here in `gpg/constants.py` should not create a circular dependency.

It might look a little weird, but I think that's better than hardcoding the same string twice.

lukpueh (approved):

Issue #130, opened 2017-10-11
Adopt sublayout link metadata namespacing


lukpueh:
According to recent specification changes in-toto sublayout verification expects link metadata files pertaining to a sublayout to reside in a fileystem directory with the following naming scheme: 

`<SUBLAYOUT NAME>.<SUBLAYOUT OWNER KEYID PREFIX>`

**SUBLAYOUT NAME:** Name of step in superlayout that is carried out as sublayout
**SUBLAYOUT OWNER:** Functionary authorized to perform the step of superlayout that is carried out as sublayout
**KEYID PREFIX:** Left-most 8 characters of keyid

See [4.5.2 Namespacing link metadata](https://github.com/in-toto/docs/blob/69c29b41f4c1fe064b5d77d61d4b268fc171bcaa/in-toto-spec.md#452-namespacing-link-metadata) for more details

PR #189, opened 2018-02-22
Adopt sublayout link namespacing

lukpueh:
**Fixes issue #**:
#130 

This PR adopts the part in the specification that suggests that links corresponding to the steps of a sublayout are to be loaded from a subdirectory namespaced using the name of the step in the superlayout for which the sublayout serves as link and the sublayout's signing key keyid prefix.

In that course `verifylib.in_toto_verify` and the corresponding cli tool `in-toto-verify` get a new argument that can also be used to tell the verification routine to load links from a different place than the current working directory (default). For sublayouts this path is generated automatically as described above.

requirements**:

### DIFF
--- a/tests/test_verifylib.py
+++ b/tests/test_verifylib.py
@@ -1328,7 +1328,7 @@ class TestVerifySublayouts(unittest.TestCase):
     self.super_layout.steps.append(sub_layout_step)
 
     # Load the super layout links (i.e. the sublayout)
-    self.super_layout_links = load_links_for_layout(self.super_layout)
+    self.super_layout_links = load_links_for_layout(self.super_layout, ".")
 
   @classmethod
   def tearDownClass(self):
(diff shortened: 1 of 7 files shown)


tests/test_verifylib.py
```diff
@@ -1326,7 +1354,119 @@ def tearDownClass(self):
   def test_verify_demo_as_sublayout(self):
     """Test super layout's passing sublayout verification. """
     verify_sublayouts(
-        self.super_layout, self.super_layout_links)
+        self.super_layout, self.super_layout_links, ".")
+
+
+
+
+
```
SantiagoTorres, line 1362: whitespace?

SantiagoTorres (approved):
Other than whitespace I think this is :+1: . I'm wondering if we want to change the `link_dir` variable to `namespace` or so. I don't know whether it makes it clearer or not though...

tests/test_verifylib.py
lukpueh: That's what the [style guideline suggests](https://github.com/secure-systems-lab/code-style-guidelines#blank-lines). I know, in a recent comment I said that I prefer 3 over 5 blank lines. But I guess I felt particularly guideline obedient here. :)

Should we open a ticket to clean up all blank lines everywhere?

SantiagoTorres: nah, I don't think it's necessary

PR #192, opened 2018-03-02
Follow symbolic links when recording artifacts

lukpueh:
This PR adds an optional parameter to the artifact record function to toggle following symbolic links to directories.

Furthermore the PR modifies the calling `runlib` functions, used with `in-toto-run` and `in-toto-record`,  to call with `follow_links=True` .

*UPDATE:*
This change also affects in-toto's final product verification, where, when running inspections, we record artifacts (now following symlinks) in the directory, where the verification is executed.

requirements**:

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -130,10 +130,13 @@ def record_artifacts_as_dict(artifacts, follow_symlink_dirs=False):
             A list of file or directory paths used as materials or products for
             the link command.
 
-            Follow symbolic links if the linked dir exists (default is False).
     follow_symlink_dirs: (optional)
+            Follow symlinked dirs if the linked dir exists (default is False).
+            The recorded path contains the symlink name, not the resolved name.
             NOTE: This parameter toggles following linked directories only,
             linked files are always recorded, independently of this parameter.
+            NOTE: Beware of infinite recursions that can occur if a symlink
+            points to a parent directory or itself.
 
   <Exceptions>
     in_toto.exceptions.SettingsError
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -87,7 +87,7 @@ def _apply_exclude_patterns(names, exclude_patterns):
   return names
 
 
-def record_artifacts_as_dict(artifacts, follow_links=False):
+def record_artifacts_as_dict(artifacts, follow_symlink_dirs=False):
   """
   <Purpose>
     Hashes each file in the passed path list. If the path list contains
@@ -130,8 +130,8 @@ def record_artifacts_as_dict(artifacts, follow_links=False):
             A list of file or directory paths used as materials or products for
             the link command.
 
-    follow_links: (optional)
             Follow symbolic links if the linked dir exists (default is False).
+    follow_symlink_dirs: (optional)
             NOTE: This parameter toggles following linked directories only,
             linked files are always recorded, independently of this parameter.
 
@@ -184,7 +184,8 @@ def record_artifacts_as_dict(artifacts, follow_links=False):
       artifacts_dict[artifact] = _hash_artifact(artifact)
 
     elif os.path.isdir(artifact):
-      for root, dirs, files in os.walk(artifact, followlinks=follow_links):
+      for root, dirs, files in os.walk(artifact,
+          followlinks=follow_symlink_dirs):
         # Create a list of normalized dirpaths
         dirpaths = []
         for dirname in dirs:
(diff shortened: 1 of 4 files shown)


JustinCappos:
Are there security implications for this?  Symlinks are very useful in causing a lot of race conditions.

lukpueh:
Good question. Do you have anything specific in mind? I am not sure if race conditions are a problem, as we only read from files. The only potential issue that I can think of is getting trapped in an infinite recursion (see [note in `os.walk` docs](https://docs.python.org/3.6/library/os.html#os.walk)).

Btw. I forgot to mention that the change also affects in-toto's final product verification, where, when running inspections, we also record artifacts in the directory, where the verification is executed.

in_toto/runlib.py
```diff
@@ -87,7 +87,7 @@ def _apply_exclude_patterns(names, exclude_patterns):
   return names
 
 
-def record_artifacts_as_dict(artifacts):
+def record_artifacts_as_dict(artifacts, follow_links=False):
```
vladimir-v-diaz: Unrelated to this pull request:

> Paths are normalized for matching and storing by left stripping "./"

Is this behavior similar to https://github.com/theupdateframework/tuf/issues/639?  @JustinCappos recommended that we raise an exception if a path includes a leading "/"

in_toto/runlib.py
```diff
@@ -130,6 +130,11 @@ def record_artifacts_as_dict(artifacts):
             A list of file or directory paths used as materials or products for
             the link command.
 
+    follow_links: (optional)
+            Follow symbolic links if the linked dir exists (default is False).
+            NOTE: This parameter toggles following linked directories only,
```
vladimir-v-diaz: > Follow symbolic links if the **linked** dir exists (default is False).

I recommend you use "symlink" when referring to symbolic links, otherwise it clashes with in-toto's use of "link cmd," "link files," "Link Object," etc.

vladimir-v-diaz: Would it help to rename `follow_links` to something like `follow_symlink_dir`?  It might clear up the possible confusion with the default/different behavior with symlinked files.

vladimir-v-diaz: You should also consider adding a note about an infinite recursion possibility if a symlink points to a parent directory of itself.

in_toto/runlib.py
```diff
@@ -398,7 +403,7 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
 
   if material_list:
     log.info("Recording materials '{}'...".format(", ".join(material_list)))
-  materials_dict = record_artifacts_as_dict(material_list)
+  materials_dict = record_artifacts_as_dict(material_list, follow_links=True)
```
vladimir-v-diaz: So `follow_links` is always set to true for someone using/calling `in_toto_run`?  The docstring for `in_toto_run` should probably mention this default behavior.

vladimir-v-diaz (commented):
Here is my review of `runlib.py.`

tests/test_runlib.py
```diff
@@ -144,7 +142,7 @@ def test_bad_base_path_setting(self):
     for setting in ["path/does/not/exist", 12345, True]:
       in_toto.settings.ARTIFACT_BASE_PATH = setting
       with self.assertRaises(in_toto.exceptions.SettingsError):
-        record_artifacts_as_dict(["."])
+        record_artifacts_as_dict(["."] )
```
vladimir-v-diaz: Why was a space added?

tests/test_runlib.py
```diff
@@ -181,8 +179,98 @@ def test_record_dot_check_files_hash_dict_schema(self):
...
+      # Test only the files were recorded ...
+      self.assertListEqual(sorted(list(artifacts_dict.keys())),
+          sorted(self.full_file_path_list))
+
+    for link in links:
+      os.unlink(link)
+
+
+  def test_record_follow_symlinked_directories(self):
+    """Record files in symlinked dirs if follow_links is True. """
```
vladimir-v-diaz: I don't recall whether the docstring for `follow_links` explains it, but which file path is recorded when a symlinked directory is visited?  The original path, or the resolved one?

vladimir-v-diaz (commented):
Here is my review of `test_runlib.py.`

in_toto/runlib.py
lukpueh: No, this is different. It is about normalizing/unifying relative path that might start with or without `./`
```python
>>> os.path.normpath("./relative/to/cwd") == "relative/to/cwd"
True
```
IIRC, your case is about normalizing/unifying absolute and relative paths, when the absolute root is also the relative root of a path, right?

in_toto/runlib.py
lukpueh: All good points, I will change as you suggest

in_toto/runlib.py
lukpueh: Will do

lukpueh:
Thanks for the review @vladimir-v-diaz. Merging...

Issue #194, opened 2018-03-02
Support path left strip when recording artifacts


lukpueh:
**Description of issue or feature request**:
Add support for left stripping artifact paths before storing them to link metadata. This could give more flexibility to functionaries and simplify writing artifact rules.

**Current behavior**:
Artifacts are recorded as full paths or relative to the current working directory from where `in-toto-run` or `in-toto-record` is executed (see [`--materials`/`--products` cli arguments](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py#L39-L46)), e.g:

```
$ in-toto-run --materials /home/foo ../bar baz  <...>
```
```
{
  _type: "link"
  materials : {
    "/home/foo": HASHDICT,
    "../bar": HASHDICT,
    "baz": HASHDICT
  }
  ...
}
```

**Expected behavior**:
```
$ in-toto-run --lstrip-paths /home/  ../ --materials /home/foo ../bar baz  <...>
``` 
```
{
  _type: "link"
  materials : {
    "foo": HASHDICT,
    "bar": HASHDICT,
    "baz": HASHDICT
  }
  ...
}
```

This change could make the (less obvious) `ARTIFACT_BASE_PATH` setting obsolete, which when used makes the artifact recording function change to that path and then look for the artifacts, passed using the `--materials/--products` arguments, relative from that path and also record them as such relative paths. (see [`runlib.record_artifacts_as_dict`](https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L148-L156))

adityasaky:
I just wanted to put it out there that I'm working on this issue right now.

adityasaky:
Continuing from the related discussion at #249:

We should discuss about adding a `/` at the end of the prefix if it's not already there. I imagine the prefix will always refer to a directory (some ancestor typically). If `/` isn't provided as part of the prefix, the key in the dictionary may then end up beginning as `/foo` instead of just `foo`.

Also, if we choose to add a `/`, that can cause a separate issue if we're not normalizing the prefix and we have a mix of unix and windows-style paths, and a windows-style path ends up with a `/` at the end. Thoughts?

lukpueh:
@adityasaky, I am not so sure about this. Maybe there are cases where the caller wants to end up with `/foo`. I think not adding a trailing prefix would make this option more flexible. You can add a comment  in the code that a trailing separator might be something to consider.

adityasaky:
Okay, sounds good.

[...]

PR #199, opened 2018-03-12
Add base path cli argument to in-toto-run/in-toto-record

lukpueh:
in-toto provides an `ARTIFACT_BASE_PATH` setting to be used as base path for any materials and products, specified via `--materials` and `--products` arguments, when recording link metadata.

This PR adds a `--base-path`, argument to `in-toto-run` and `in-toto-record` to specify the base path on the cli tool. A base path passed via argument overrides a base path specified via setting (or envvar or RCfile).

*Note:*
The base path setting was originally added to record relative artifact paths, not relative to the current working directory, e.g. to record an artifact at `/path/to/bar` as `bar`, which before was only possible when the directory, in which `in-toto-run`/`in-toto-record` was executed, was `/path/to/`.

An alternative approach that might be more intuitive is proposed in #194.

requirements**:

### DIFF
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -37,3 +37,12 @@ EXCLUDE_KWARGS = {
           " e.g.: environment variables or RCfiles. See"
           " ARTIFACT_EXCLUDE_PATTERNS documentation for additional info.")
   }
+
+BASE_PATH_ARGS = ["--base-path"]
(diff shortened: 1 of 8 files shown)


SantiagoTorres:
The discussion of this PR is missing for some reason, but it LGTM now.

PR #200, opened 2018-03-12
FIX: requirements: Drop pinned version of securesystemslib

SantiagoTorres:
Remove the `==` on securesystemslib. As we no longer require a pinned version, we can change it to `>=`. The `>=` is however required to ensure that no older, api-breaking, version are installed for whatever reason.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-securesystemslib==0.10.10
+securesystemslib>=0.10.10
 cryptography
 attrs
 python-dateutil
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
-securesystemslib==0.10.10
+securesystemslib>=0.10.10
```
vladimir-v-diaz, line 1: Is the `>=` needed to catch breaking changes during development?  It might be good to explain in the pull request summary.

Also, is it okay to still pin securesystemslib to 0.10.10 in `setup.py?`

SantiagoTorres: I'll elaborate on this. Also thanks for the reminder on the 0.10.10 in setup.py. Sorry for the late response!

vladimir-v-diaz:
LGTM.

PR #201, opened 2018-03-13
FIX: verifylib: fix subkey keyid resolution

SantiagoTorres:
The load_links_for_layout method in verifylib didn't account for subkey_keyid's
when resolving signatures. Add code to iterate over the subkeys when verifying
signatures made with gpg keys.

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -129,16 +129,20 @@ def load_links_for_layout(layout, link_dir_path):
     # We try to load a link for every authorized functionary, but don't fail
     # if the file does not exist (authorized != required)
     # FIXME: Should we really pass on IOError, or just skip inexistent links?
-    for keyid in step.pubkeys:
-      filename = FILENAME_FORMAT.format(step_name=step.name, keyid=keyid)
-      filepath = os.path.join(link_dir_path, filename)
+    for authorized_keyid in step.pubkeys:
+      # Iterate over the authorized key and if present over subkeys
+      for keyid in [authorized_keyid] + list(layout.keys.get(authorized_keyid,
+          {}).get("subkeys", {}).keys()):
 
-      try:
-        metadata = Metablock.load(filepath)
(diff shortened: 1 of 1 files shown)


in_toto/verifylib.py
```diff
@@ -129,16 +129,20 @@ def load_links_for_layout(layout, link_dir_path):
...
+          metadata = Metablock.load(filepath)
+          links_per_step[keyid] = metadata
 
-      except IOError:
-        pass
+        except IOError:
+          pass
 
     # Check if the step has been performed by enough number of functionaries
     if len(links_per_step) < step.threshold:
```
vladimir-v-diaz, line 148: Is the master key, along with its subkeys, considered one unique functionary?  Correct me if I'm wrong, but it seems like the code is counting each subkey towards the threshold requirement.

I was expecting (MasterKey, SubKey1, SubKey2) to be treated as one trusted key.  If you want three different people to sign for a step, isn't it wrong to accept three different subkeys from just one person?

lukpueh: Good thinking, @vladimir-v-diaz! This is something that we have to consider, but not in above function, which only performs a preliminary threshold verification, i.e. do we even have enough link files, but rather in the actual signature threshold verification function [`veriflylib. verify_link_signature_thresholds`](https://github.com/SantiagoTorres/in-toto/blob/1a1a1d82e275fa25b609765ef1514b99239497a4/in_toto/verifylib.py#L282).

Currently,  a functionary could indeed bypass the required threshold by providing `threshold` link files signed with different subkeys of an authorized master key.

SantiagoTorres: Hmm, I guess a check is in order in the right place. I'll give this a go this week...

SantiagoTorres: I've moved this discussion to ticket #202. I'll be merging this as-is in order to prepare a new minor release.

PR #204, opened 2018-04-17
Fix firewall rules

SantiagoTorres:
that is provided, the better.

**Fixes issue #**: 

None, this issue was brought up from conversations with potential integrators

The current behavior of the artifact rules throw exceptions when an element in the *filtered* queue doesn't match the target. For example, if a file `foo` is being checked for a `MODIFY` rule then a `VerificationError` will be raised if `foo` is indeed not modified. This is not how firewalls behave.

Instead, `foo` should just remain in the queue and expected to be matched by other rules later down the line. The only rule that should throw an error is a `DISALLOW` rule when artifacts are matched.

requirements**:

This is still somewhat of a WIP branch, as I'd like to have some feedback from @vladimir-v-diaz and @lukpueh (if possible)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## in-toto v0.2.2
 
 * Add support for gpg signing subkeys.
-* Drop strict requirement on securesystemslib 0.9.
+* Drop strict requirement on securesystemslib 0.10.
 * Command line tool changes:
     - Add a --base-path parameter to in-toto record and in-toto run
     - in-toto-record now follows symbolic links
@@ -11,6 +11,7 @@
 * Adds support for sublayout namespacing (for in-toto spec 0.9 compliance)
 * Path prefix is normalized during in-toto verification:
     - Paths such as foo//bar match with foo/bar.
+* Dropped obsolete SettingsError
 
 
 ## in-toto v0.2.1
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -642,18 +642,15 @@ def verify_match_rule(rule, source_artifacts_queue, source_artifacts, links):
     try:
       dest_artifact = dest_artifacts[full_dest_path]
 
-    except Exception:
-      raise RuleVerificationError("Rule '{rule}' failed, destination artifact"
-          " '{path}' not found in {type} of '{name}'"
-              .format(rule=" ".join(rule), path=full_dest_path, name=dest_name,
-                  type=dest_type))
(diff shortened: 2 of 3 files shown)


SantiagoTorres:
FYI: these failures appear to be due to the set operator using a different `PYTHONHASHSEED`

vladimir-v-diaz:
@JustinCappos requested that @awwad review more in-toto pull requests, so I'll add Sebastien to the list of reviewers.  For now, I need to focus on fixing Windows-related issues for the TUF project.

SantiagoTorres:
> @JustinCappos requested that @awwad review more in-toto pull requests, so I'll add Sebastien to the list of reviewers. For now, I need to focus on fixing Windows-related issues for the TUF project.

Noted, and thanks!

@awwad welcome aboard!

awwad:
So if I understand correctly, with this PR merged, if you have some rule set wherein rule 1 reads: `MATCH arty WITH PRODUCTS FROM step1`,  rules 2 through 5 have nothing to do with arty, and rule 6 is `DISALLOW *`, then the error you get when arty exists and has a different hash than that listed for arty in the step1.link products is an error about arty being disallowed by rule 6, right?

awwad:
Also, with this construction (rule verifier functions adjust the input to remove what they've matched), none of the hash mismatch info gets reported anywhere -- neither trickles up in an exception that's caught somewhere nor logged. Is that okay in this context? Is that going to be regrettable eventually?

In any case, the PR seems sane.

awwad (approved):

SantiagoTorres:
wow, sorry for some reason I missed this PR review. 

> So if I understand correctly, with this PR merged, if you have some rule set wherein rule 1 reads: MATCH arty WITH PRODUCTS FROM step1, rules 2 through 5 have nothing to do with arty, and rule 6 is DISALLOW *, then the error you get when arty exists and has a different hash than that listed for arty in the step1.link products is an error about arty being disallowed by rule 6, right?

Yes, this is pretty much it. We're trying to have DISALLOW be the only rule that fails, where the others match and process. That's the behavior we had described in the specification.

> Also, with this construction (rule verifier functions adjust the input to remove what they've matched), none of the hash mismatch info gets reported anywhere -- neither trickles up in an exception that's caught somewhere nor logged. Is that okay in this context? Is that going to be regrettable eventually?

We were worried about this too, but it's a shared issue on firewall-like semantics. I don't think we can make much about it (yet).

> In any case, the PR seems sane.

Thanks a lot! Merging :)

PR #205, opened 2018-05-04
Add partial Ed25519 support

trishankatdatadog:
**Fixes issue #**:

Presently, in-toto supports using only RSA keys.

This PR supports using both RSA and Ed25519 keys.

requirements**:


in_toto/util.py
```diff
@@ -134,8 +149,39 @@ def prompt_password(prompt="Enter password: "):
   return getpass.getpass(prompt, sys.stderr)
 
 
+def prompt_import_private_key_from_file(filepath):
+  """Trys to load an Ed25519 or RSA key without password. If a CryptoError
+  occurs, prompts the user for a password and trys to load the the key again.
+  """
+  # FIXME: The more secure method is to load the private key using the key type
```
trishankatdatadog, line 156: This function is dangerous in the sense that it makes the assumption that a private key is always either an Ed25519, or an RSA key, in that order. I haven't fully thought this through, but an attacker might be able to exploit this in order to bind private keys to [mismatching cryptographic schemes](https://www.nccgroup.trust/us/our-research/docker-notary/).

A more robust function would either: (1) ask the user to explicitly identify the keytype, and / or (2) correctly guess the keytype from information in the private key itself.

trishankatdatadog:
Related to https://github.com/secure-systems-lab/securesystemslib/pull/148

trishankatdatadog:
It looks like the tests are failing because of some fixes for GPG that is on [add-substitution-layer](https://github.com/in-toto/in-toto/tree/add-substitution-layer) but not [develop](https://github.com/in-toto/in-toto/tree/develop).

Update: huh, I spoke too soon. For whatever reason, tox fails on my machine, but works on coveralls. Who the hells knows what's happening.

trishankatdatadog:
This PR is certainly not complete, because the other CLI tools need to be patched and tested, too.

tox.ini
```diff
@@ -15,6 +15,7 @@ deps =
     bandit
 
 commands =
+    pip install -U https://github.com/secure-systems-lab/securesystemslib/archive/master.zip
```
trishankatdatadog, line 18: Also, please be sure to remove this line --- it's only there for testing.

vladimir-v-diaz:
Note: I reviewed pull request #209, which is based on and expands this pull request.

SantiagoTorres:
I merged #209, which supersedes this PR. I'll close this one now.

PR #206, opened 2018-05-16
ENH:WIP: Add substitution layer

SantiagoTorres:
**Fixes issue #**:

None

This pull request adds the substitution layer code for our current integrators.
Note that this solution is just a stepping stone into a more robust,
full-fledged solution for a version 0.3.0 of in-toto

    Integrators and package managers that are performing in-toto verification 
need a way to signal some contextual information about the artifacts that 
are being verified (e.g., what's the name of the actual package that was 
received!). The simplest way to provide this functionality is by adding a 
simple replacement layer. Thus, a package manager that wants to verify can call: 
 
   `in_toto_verify(metadata_path, root_layout_keys, param_substitutions=...) `
     
This PR enables this functionality on the root layout, so that artifact 
names or commands to run can be provided to in-toto before execution. 
The parameter dictionary is a simple non-nested {key:value} set of strings 
that can be passed to perform substitution on python-like formatted strings 
within the layout. The tokens to replace are to be defined by the project 
owner when creating the layout. 

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -312,7 +312,8 @@ def substitute_parameters(layout, parameter_dictionary):
     step.expected_products = new_product_rules
 
   for inspection in layout.inspect:
-    inspection.run = inspection.run.format(**parameter_dictionary)
+    new_run = [x.format(**parameter_dictionary) for x in inspection.run]
+    inspection.run = new_run
 
 
 def verify_layout_signatures(layout_metablock, keys_dict):
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -312,8 +312,20 @@ def substitute_parameters(layout, parameter_dictionary):
     step.expected_products = new_product_rules
 
   for inspection in layout.inspect:
+
+    inspection.expected_materials = [
+        [ s.format(**parameter_dictionary) for s in m ] \
+        for m in inspection.expected_materials
+    ]
+
+    inspection.expected_products = [
+        [ s.format(**parameter_dictionary) for s in p ] \
+        for p in inspection.expected_products
+    ]
+
     new_run = [x.format(**parameter_dictionary) for x in inspection.run]
     inspection.run = new_run
+
 
 
 def verify_layout_signatures(layout_metablock, keys_dict):
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -257,8 +257,12 @@ def substitute_parameters(layout, parameter_dictionary):
             A dictionary containing key-value pairs for substitution.
 
   <Exceptions>
-    securesystemslib.FormatException:
+    securesystemslib.exceptions.FormatError:
       if the parameter dictionary is malformed.
+
+    KeyError:
+      if one of the keys in the parameter dictionary are not present for
+      substitution
 
   <Side Effects>
     The layout object will have any tags replaced with the corresponding
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -4,6 +4,7 @@
 
 <Author>
   Lukas Puehringer <lukas.puehringer@nyu.edu>
+  Santiago Torres-Arias <santiago@nyu.edu>
 
 <Started>
   June 28, 2016
@@ -47,6 +48,35 @@ import in_toto.rulelib
 
 # Inherits from in_toto base logger (c.f. in_toto.log)
 log = logging.getLogger(__name__)
+
+
+def _sanitize_parameter_dictionary(parameter_dictionary):
+  """
+  <Purpose>
+    Internal dictionary that ensures that the parameter dictionary:
+      - only consists of string-based keys
+      - All values are strings
+      - None of the keys have any characters outside of the allowed character
+        set ([a-zA-Z0-9-_]).
+
+  <Arguments>
+    parameter_dictionary:
+      The dictionary to verify.
+
+  <Exceptions>
+    SchemaMismatchError if the dictionary doesn't comply with the requirements
+    described above.
+
+  <Side Effects>
+    None.
+
+  <Returns>
+    None.
+  """
+  for key in parameter_dictionary:
+    in_toto.formats.PARAMETER_DICTIONARY_KEY.check_match(key)
+    in_toto.formats.PARAMETER_DICTIONARY_VALUE.check_match(parameter_dictionary[key])
+
 
 def _raise_on_bad_retval(return_value, command=None):
   """
@@ -238,6 +268,51 @@ def verify_layout_expiration(layout):
   expire_datetime = iso8601.parse_date(layout.expires)
   if expire_datetime < datetime.datetime.now(tz.tzutc()):
     raise LayoutExpiredError("Layout expired")
+
+
+def substitute_parameters(layout, parameter_dictionary):
+  """
+  <Purpose>
+    This function is a transitionary measure for parameter substitution (or
+    any other solution defined by the in-toto team). As of now, it acts as 
+    a very simple replacement layer for python-like parameters
+
+  <Arguments>
+    layout:
+            The Layout object to process.
+
+      parameter_dictionary:
+            A dictionary containing key-value pairs for substitution.
+
+  <Exceptions>
+    securesystemslib.FormatException:
+      if the parameter dictionary is malformed.
+
+  <Side Effects>
+    The layout object will have any tags replaced with the corresponding
+    values defined in the parameter dictionary.
+  """
+  _sanitize_parameter_dictionary(parameter_dictionary)
+
+  for step in layout.steps:
+
+    step.expected_command = [x.format(**parameter_dictionary) for x in step.expected_command]
+
+    new_material_rules = []
+    for rule in step.expected_materials:
+      new_rule = [x.format(**parameter_dictionary) for x in rule]
+      new_material_rules.append(new_rule)
+
+    new_product_rules = []
+    for rule in step.expected_products:
+      new_rule = [x.format(**parameter_dictionary) for x in rule]
+      new_product_rules.append(new_rule)
+
+    step.expected_materials = new_material_rules
+    step.expected_products = new_product_rules
+
+  for inspection in layout.inspect:
+    inspection.run = inspection.run.format(**parameter_dictionary)
 
 
 def verify_layout_signatures(layout_metablock, keys_dict):
@@ -1338,7 +1413,8 @@ def get_summary_link(layout, reduced_chain_link_dict):
   return Metablock(signed=summary_link)
 
 
-def in_toto_verify(layout, layout_key_dict, link_dir_path="."):
+def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
+    substitution_parameters=None):
   """
   <Purpose>
     Does entire in-toto supply chain verification of a final product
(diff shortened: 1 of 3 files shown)


vladimir-v-diaz:
Is this pull request still a work in progress (as indicated in the title)?

Can you also please add an example in the summary of how you expect users to use parameter substitution with in-toto?

in_toto/verifylib.py
```diff
@@ -48,6 +49,35 @@
 # Inherits from in_toto base logger (c.f. in_toto.log)
 log = logging.getLogger(__name__)
 
+
+def _sanitize_parameter_dictionary(parameter_dictionary):
+  """
+  <Purpose>
+    Internal dictionary that ensures that the parameter dictionary:
```
vladimir-v-diaz: Should this be "Internal function" ?  I'm not sure what is meant by an internal dictionary.

in_toto/verifylib.py
```diff
@@ -48,6 +49,35 @@
...
+  <Side Effects>
+    None.
+
+  <Returns>
+    None.
+  """
+  for key in parameter_dictionary:
+    in_toto.formats.PARAMETER_DICTIONARY_KEY.check_match(key)
+    in_toto.formats.PARAMETER_DICTIONARY_VALUE.check_match(parameter_dictionary[key])
+
```
vladimir-v-diaz: Is this checking that the dict keys and values are a specific format? 
Why not use SCHEMA.DictOf(), which does exactly that?   See the following example:

https://github.com/in-toto/in-toto/blob/a14f79bf8b81728d4a2e2c1a8795f3ec42dc2c96/in_toto/formats.py#L39-L41

in_toto/verifylib.py
```diff
@@ -48,6 +49,35 @@
...
+      - All values are strings
+      - None of the keys have any characters outside of the allowed character
+        set ([a-zA-Z0-9-_]).
+
+  <Arguments>
+    parameter_dictionary:
+      The dictionary to verify.
+
+  <Exceptions>
+    SchemaMismatchError if the dictionary doesn't comply with the requirements
```
vladimir-v-diaz: Is this supposed to be FormatError?

in_toto/verifylib.py
```diff
@@ -240,6 +270,64 @@ def verify_layout_expiration(layout):
...
+
+  <Arguments>
+    layout:
+            The Layout object to process.
+
+      parameter_dictionary:
+            A dictionary containing key-value pairs for substitution.
+
+  <Exceptions>
+    securesystemslib.FormatException:
```
vladimir-v-diaz: I'm not sure this is the correct exception.  Isn't it FormatError?  Please verify that the expected exceptions are documented correctly.  The _sanitize_parameter_dictionary() call on line 295 is expected to raise SchemaMismatchError (according to its docstring), but that's not documented here.

in_toto/verifylib.py
```diff
@@ -240,6 +270,64 @@ def verify_layout_expiration(layout):
...
+      if the parameter dictionary is malformed.
+
+  <Side Effects>
+    The layout object will have any tags replaced with the corresponding
+    values defined in the parameter dictionary.
+  """
+  _sanitize_parameter_dictionary(parameter_dictionary)
+
+  for step in layout.steps:
```
vladimir-v-diaz: Why isn't the `layout` argument validated like `parameter_dictionary`?  Is it assumed, and why isn't that explicitly stated in the docstring?

in_toto/verifylib.py
```diff
@@ -240,6 +270,64 @@ def verify_layout_expiration(layout):
...
+        for m in inspection.expected_materials
+    ]
+
+    inspection.expected_products = [
+        [ s.format(**parameter_dictionary) for s in p ] \
+        for p in inspection.expected_products
+    ]
+
+    new_run = [x.format(**parameter_dictionary) for x in inspection.run]
+    inspection.run = new_run
```
vladimir-v-diaz: Per our guidelines against list comprehensions, please consider using s.replace() instead, and being explicit about what is being replaced.

I suggest that the `parameter_dictionary` argument be of the form:
```
{"{COMMAND}", "vim"}
```
It might be easier for others to see that what you want to replace is `{COMMAND}`  and not `COMMAND`.

in_toto/verifylib.py
```diff
@@ -240,6 +270,64 @@ def verify_layout_expiration(layout):
     raise LayoutExpiredError("Layout expired")
 
 
+def substitute_parameters(layout, parameter_dictionary):
+  """
+  <Purpose>
+    This function is a transitionary measure for parameter substitution (or
+    any other solution defined by the in-toto team). As of now, it acts as 
+    a very simple replacement layer for python-like parameters
```
vladimir-v-diaz, line 250: It might be good to discuss the use cases for this functionality.  Why and when should an adopter use the substitution layer?  

Also, is this functionality usable from the command-line tools?  This pull request appears to only support it from `verifylib.py`.

tests/test_param_substitution.py
```diff
@@ -0,0 +1,201 @@
...
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+
+<Started>
+  Nov 07, 2016
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Test verifylib functions.
```
vladimir-v-diaz: This purpose seems wrong.  It's not really testing the verifylib functions, but only param substitution.  There already is a test_verifylib.py.  Shouldn't the substitute function be tested there, in `test_verifylib.py`?

vladimir-v-diaz (commented):

in_toto/verifylib.py
SantiagoTorres: This is how it works right now. That's why we are using the built-in `format` call, rather than `replace`

tests/test_param_substitution.py
SantiagoTorres: I did this on purpose, as to avoid a feature that we may later migrate of. I.e., it makes easier to refactor the code this way

in_toto/verifylib.py
SantiagoTorres: No, I didn't add the functionality from the command line tools yet, as the only people interested in this (potentially temporary) feature is doing a direct call to verifylib.

in_toto/verifylib.py
SantiagoTorres: I'm not completely sure, this is a PR that adds `parameter_dictionary`. I'll dig up on that on a separate PR (I'm almost sure that it's checked at the time of de-serialization somewhere else in the code).

in_toto/verifylib.py
SantiagoTorres: Yes, I left this as a placeholder. Let me make sure I have the name right.

in_toto/verifylib.py
SantiagoTorres: Good catch. I refactored this function out, as it appears it just does what `schema.DictOf` already does. Thanks!

SantiagoTorres:
I see most of the feedback is about documenting the feature itself. Let me take care of this and post it here. In the meantime, I'll fix one of the issues and remove the WIP label

in_toto/verifylib.py
vladimir-v-diaz: Let me use the unit test example to be clear.
```
def test_substitute(self):
"""Do a simple substitution on the expected_command field"""
substitute_parameters(self.layout, {"EDITOR": "vim"})
```

I think it might be more readable if the second argument to `substitute_parameters()` was required as follows:
```Python
{"{EDITOR}", "vim"}
```

Just a suggestion.

SantiagoTorres: Hmm, I'm not completely convinced to having the argument named like that. It just adds complexity on the codebase and I think the change in clarity is very little. We can discuss offline to decide something.

SantiagoTorres:
Note: builds will fail until we fix https://github.com/secure-systems-lab/securesystemslib/issues/153

vladimir-v-diaz:
The master/develop branch seems to be passing.  Has this pull request been sync'd with the develop branch?

in_toto/verifylib.py
```diff
@@ -240,6 +242,84 @@ def verify_layout_expiration(layout):
...
+
+    KeyError:
+      if one of the keys in the parameter dictionary are not present for
+      substitution
+
+  <Side Effects>
+    The layout object will have any tags replaced with the corresponding
+    values defined in the parameter dictionary.
+  """
+  in_toto.formats.PARAMETER_DICTIONARY_SCHEMA.check_match(parameter_dictionary)
```
vladimir-v-diaz, line 271: nit: Consider including an example of a parameter_dictionary in a comment (especially since the strings being replaces are expected to contain the curly braces).  I had to refer back to the unit tests and in_toto.formats to confirm the expected format.

vladimir-v-diaz:
I tried executing some of the functions and unit tests and came across the following error (it might be related to GPG?):

```Bash
======================================================================
ERROR: test_verify_link_signature_thresholds__S_M_M (tests.test_verifylib.TestInTotoVerifyThresholdsGpgSubkeys)
Allowed trust delegation.
----------------------------------------------------------------------
  File "/Users/vlad/projects/in-toto/tests/test_verifylib.py", line 1423, in test_verify_link_signature_thresholds__S_M_M
    verify_link_signature_thresholds(layout, chain_link_dict)
  File "/Users/vlad/projects/in-toto/in_toto/verifylib.py", line 481, in verify_link_signature_thresholds
    " found '{}'".format(step.threshold, valid_authorized_links_cnt))
in_toto.exceptions.ThresholdVerificationError: Step requires at least '1' links validly signed by different authorized functionaries. Only found '0'
```

tests/test_param_substitution.py
```diff
@@ -0,0 +1,215 @@
...
+    self.assertEquals(self.layout.steps[0].expected_products[0][1], "new_thing")
+    self.assertEquals(self.layout.inspect[0].expected_materials[0][1], "vim")
+    self.assertEquals(self.layout.inspect[0].expected_materials[0][5], "source_step")
+    self.assertEquals(self.layout.inspect[0].expected_products[0][1], "new_thing")
+
+
+
+  def test_substitute_no_var(self):
+    """Raise an error if the parameter is not filled-in"""
+    with self.assertRaises(KeyError):
```
vladimir-v-diaz, line 99: Is there a unit test for a parameter dictionary that doesn't contain the expected format?

Another thing you  might want to do is raise a warning if the rule to be substituted doesn't contain the curly braces, which someone might forget to include.

in_toto/verifylib.py
```diff
@@ -240,6 +242,84 @@ def verify_layout_expiration(layout):
...
+
+  for step in layout.steps:
+
+    new_material_rules = []
+    for rule in step.expected_materials:
+      new_rule = []
+      for stanza in rule:
```
vladimir-v-diaz, line 278: Why did you choose the name `stanza`?  Is this terminology used when dealing with variable substitutions?

vladimir-v-diaz:
I just had a few more comments for the follow-up review of this pull request.  Feel free to merge when you wish.

in_toto/verifylib.py
SantiagoTorres: No, but I couldn't think of anything better :confounded:

Issue #171, opened 2018-01-17
Tox test build failing (can't reproduce, maybe related to GPG)


lukpueh:
**Description of issue or feature request**:
I keep getting build errors for different tox jobs, locally and on Travis CI. But when I re-run tox (locally) or restart the Travis build (via web UI) the builds pass.

Here's parts of a local tox build for commit 7016b95 (PR #170) , where py3.6 failed:

```
py36 runtests: commands[2] | coverage run test/runtests.py

[OUTPUT STRIPPED FOR READABILITY]

----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/lukp/in-toto/in-toto/in_toto/in_toto_verify.py", line 105, in in_toto_verify
    verifylib.in_toto_verify(layout, layout_key_dict)
  File "/Users/lukp/in-toto/in-toto/in_toto/verifylib.py", line 1315, in in_toto_verify
    verify_layout_signatures(layout, layout_key_dict)
  File "/Users/lukp/in-toto/in-toto/in_toto/verifylib.py", line 251, in verify_layout_signatures
    layout.verify_signatures(keys_dict)
  File "/Users/lukp/in-toto/in-toto/in_toto/models/metadata.py", line 255, in verify_signatures
    raise SignatureVerificationError("Invalid signature")
in_toto.exceptions.SignatureVerificationError: Invalid signature

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/lukp/in-toto/in-toto/test/test_in_toto_verify.py", line 209, in test_gpg_signed_layout_with_gpg_functionary_keys
    in_toto_verify_main()
  File "/Users/lukp/in-toto/in-toto/in_toto/in_toto_verify.py", line 161, in main
    in_toto_verify(args.layout, args.layout_keys, args.gpg, args.gpg_home)
  File "/Users/lukp/in-toto/in-toto/in_toto/in_toto_verify.py", line 109, in in_toto_verify
    sys.exit(1)
SystemExit: 1

----------------------------------------------------------------------
Ran 189 tests in 14.527s

[OUTPUT STRIPPED FOR READABILITY]

ERROR: InvocationError: '/Users/lukp/in-toto/in-toto/.tox/py36/bin/coverage run test/runtests.py'
___________________________________ summary ____________________________________
  py27: commands succeeded
  py34: commands succeeded
  py35: commands succeeded
ERROR:   py36: commands failed
```

I tried the following things, none of which could reproduce the problem:

[...]

PR #209, opened 2018-05-24
Add keytype support

yashsriv:
**Fixes pull request #205**:

Based on #205 to add support for `ed25519` as well as allow specifying type of keys via command line args.

requirements**:


in_toto/in_toto_keygen.py
```diff
@@ -99,11 +104,17 @@ def main():
...
+      elif args.type == in_toto.util.KeyType.ED25519:
+        in_toto.util.prompt_generate_and_write_ed25519_keypair(args.name)
       sys.exit(0)
     else:
-      in_toto.util.generate_and_write_rsa_keypair(args.name, args.bits,
-        password=None)
+      if args.type == in_toto.util.KeyType.RSA:
+        in_toto.util.generate_and_write_rsa_keypair(args.name)
+      elif args.type == in_toto.util.KeyType.ED25519:
+        in_toto.util.generate_and_write_ed25519_keypair(args.name)
```
yashsriv: This dangling `elif` causes coverage to think that there is a case which I haven't tested. How should I tackle it?

in_toto/util.py
```diff
@@ -1,62 +1,53 @@
...
+import securesystemslib.interface
 import securesystemslib.keys
 import securesystemslib.exceptions
 
+from securesystemslib.interface import (import_ed25519_privatekey_from_file,
+                                        import_ed25519_publickey_from_file)
+
 DEFAULT_RSA_KEY_BITS = 3072
 
-def generate_and_write_rsa_keypair(filepath, bits=DEFAULT_RSA_KEY_BITS,
```
yashsriv: This was a completely redundant copy of the same function in `securesystemslib.interface`

in_toto/in_toto_keygen.py
SantiagoTorres: a `#pragma: nocover` may be what you want in this case.

SantiagoTorres:
@vladimir-v-diaz can you take a look please? it's probably worth discussing if this single commit can be rebased off of `develop` rather than #205

in_toto/in_toto_keygen.py
vladimir-v-diaz: There are few more lines of code that are not hit by coverage.
https://travis-ci.org/in-toto/in-toto/jobs/383287249

I recommend you only use `#pragma: no cover` if you know for sure that Travis cannot execute a specific line.  In this case, shouldn't line 116 be covered by a unit test?

in_toto/in_toto_keygen.py
```diff
@@ -99,11 +104,17 @@ def main():
 
   try:
     if args.prompt:
-      in_toto.util.prompt_generate_and_write_rsa_keypair(args.name, args.bits)
+      if args.type == in_toto.util.KeyType.RSA:
+        in_toto.util.prompt_generate_and_write_rsa_keypair(args.name,
+                                                           args.bits)
+      elif args.type == in_toto.util.KeyType.ED25519:    # pragma: no branch
```
yashsriv: Using `# pragma: no branch` worked as we know for a fact that this is essentially an else statement.

tox.ini
```diff
@@ -27,7 +28,7 @@ commands =
     # https://docs.openstack.org/bandit/latest/blacklists/blacklist_imports.html#b404-import-subprocess
     # https://docs.openstack.org/bandit/latest/plugins/subprocess_popen_with_shell_equals_true.html
     # https://docs.openstack.org/bandit/latest/plugins/subprocess_without_shell_equals_true.html
-    bandit -r in_toto --skip B404,B603
+    bandit -r in_toto --skip B107,B404,B603
```
yashsriv: I had to add skipping of B107 because otherwise bandit would complain that I'm hardcoding a password wherever I have used `""` as the default parameter for password (in order to prevent prompts for password). Any workarounds around this? Using `None` as the default parameter doesn't throw errors but this does? :thinking:

in_toto/in_toto_keygen.py
```diff
@@ -20,16 +20,16 @@
...
   Example Usage:
-  Suppose Bob wants to create the keys of size 2048 bits and dump them with
+  Suppose Bob wants to create the rsa keys of size 2048 bits and dump them with
   file name "bob_keys". He also wants to encrypt the so created private key
   with his choice of passphrase. The keys would then be created, the private key
   would be encrypted and dumped as "bob_keys" and public key would be dumped
   as "bob_keys.pub". Bob will use the following command:
 
-  in-toto-keygen -p bob_keys 2048
+  in-toto-keygen -p -t rsa -b 2048 bob_keys
```
vladimir-v-diaz: A one-line example for Ed25519 might help to reinforce that the `-t ed25519` option does not require --bits.

in_toto/in_toto_keygen.py
vladimir-v-diaz: I don't quite understand why this line isn't being covered.  Shouldn't the unit test for `arg.prompt=True` and `args.type=ed25519` cover it?

yashsriv: The coverage error was for this `110 -> 112`. Basically it was unable to find any case where both of the ifs were skipped (its not possible).

in_toto/util.py
```diff
@@ -1,62 +1,53 @@
 import sys
 import getpass
+from enum import Enum
```
vladimir-v-diaz: The `enum` module was [added in Python 3.4](https://docs.python.org/3/library/enum.html).  It isn't available in Python 2 unless you use one of its backports (e.g., enum34).

I think this import statement is only working because enum34 was added as a third-party dependency (via cryptography?).

You should add `enum34` as dependency for `in-toto`, especially if we ever decide to make 'cryptography' an optional dependency. 

Another option is to avoid Enum entirely, and just use a simple list or dictionary.  We should store constants of this sort in gpg/constants.py, or similar.

in_toto/in_toto_keygen.py
vladimir-v-diaz: Ahh, all this time I've been thinking it was line coverage that was missing - it's branch coverage.

What I normally do is explain in a comment why the # pragma is being added.  It's often beneficial to document any assumptions made by the code.  In this case, you want to state that argparse enforces the values specified in --keytype.  However, will the if-elif construction still work if the `choices` parameter of add_argument() is later updated?

You might also want to use an actual else case.  For instance:

```Python


if arg.keytype == 'ed25519':
   ...
elif arg.keytype == 'rsa':
   ...
else: # pragma: no branch
    raise Error('Unsupported keytype: ' + str(arg.keytype))
```

Or raise the exception up front, as [we recommend in our code style guidelines](https://github.com/secure-systems-lab/code-style-guidelines#nesting).

in_toto/in_toto_record.py
```diff
@@ -172,6 +176,12 @@ def main():
       " link metadata."
       " (passing one of '--key' or '--gpg' is required)"))
 
+  parent_parser.add_argument("-t", "--key-type", dest="key_type",
+      type=in_toto.util.KeyType, choices=list(in_toto.util.KeyType),
```
vladimir-v-diaz: I was expecting the type to be `str` (e.g., 'rsa').  Isn't `in_toto.util.KeyType` a `class KeyType(enum.Enum)`?

in_toto/in_toto_record.py
```diff
@@ -242,7 +252,8 @@ def main():
     # case the key is encrypted. Something that should not happen in the lib.
     key = None
     if args.key:
-      key = in_toto.util.prompt_import_rsa_key_from_file(args.key)
+      key = in_toto.util.prompt_import_private_key_from_file(args.key,
+                                                             args.key_type)
```
vladimir-v-diaz: Make sure that args.key_type is indented with 4 spaces.  See our guideline on wrapping long lines.

in_toto/util.py
```diff
@@ -134,8 +141,37 @@ def prompt_password(prompt="Enter password: "):
   return getpass.getpass(prompt, sys.stderr)
 
 
+def prompt_import_private_key_from_file(filepath, key_type):
+  """Trys to load a key with/without password. If a CryptoError
```
vladimir-v-diaz: typo: "Trys"
I've seen this typo in a few other places.

in_toto/in_toto_record.py
vladimir-v-diaz: Forgot to include a link: https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length

in_toto/in_toto_record.py
yashsriv: I'd searched for how to support enums in argparse and this was the recommended solution. This works flawlessly but seeing that I'll anyways revert back to using strings, it won't make much of a difference.

yashsriv:
Hi, I have updated most of the things in a new commit so that its easier to review my fixes. I'll squash all my changes into one once review is done.

tox.ini
vladimir-v-diaz: A shortcoming of running Bandit as above (`bandit -r in_toto --skip B107,B404,...`) is that --skip is used for every single unit test.  Assuming that you only need to exclude B107 for certain tests, it might be better to create a Bandit profile for unit tests that require custom settings.  

You can take a look at the Bandit docs to see how profiles can be created in a configuration file.
https://bandit.readthedocs.io/en/latest/config.html?highlight=profile

Here is an example configuration file for Bandit that uses profiles:
https://github.com/chair6/bandit/blob/4004374a894bb41414ce07546518ac131353ec36/bandit.yaml#L21-L26

tox.ini
```diff
@@ -15,6 +15,7 @@ deps =
     bandit
 
 commands =
+    pip install -U https://github.com/secure-systems-lab/securesystemslib/archive/master.zip
```
vladimir-v-diaz: How do you plan to remove this temporary command?  Should you add a comment so we don't forget? See https://github.com/in-toto/in-toto/pull/205#discussion_r186854751.

in_toto/in_toto_keygen.py
```diff
@@ -20,16 +20,18 @@
   generated. By default it is set as 3072.
 
   General Usage:
-  in-toto-keygen [-p] <filename> [bits]
+  in-toto-keygen [-p] [-t {ed25519, rsa}] [--bits=<bits>] <filename>
 
   Example Usage:
-  Suppose Bob wants to create the keys of size 2048 bits and dump them with
+  Suppose Bob wants to create the rsa keys of size 2048 bits and dump them with
   file name "bob_keys". He also wants to encrypt the so created private key
```
vladimir-v-diaz: The following comment is unrelated to this PR, but you can address it here if you wish.

Where are keys dumped?  In the current working directory?  I think the comment / help text should make it clear.

in_toto/in_toto_record.py
```diff
@@ -242,7 +252,8 @@ def main():
     # case the key is encrypted. Something that should not happen in the lib.
     key = None
     if args.key:
-      key = in_toto.util.prompt_import_rsa_key_from_file(args.key)
+      key = in_toto.util.prompt_import_private_key_from_file(args.key,
+          args.key_type)
```
vladimir-v-diaz: Why does `in-toto-record` not provide a --prompt command-line option, similar to `in-toto-keygen`?

Do we always want to to prompt for a password when importing a private key with this script?
@SantiagoTorres

in_toto/in_toto_run.py
```diff
@@ -240,7 +249,7 @@ def main():
     # case the key is encrypted. Something that should not happen in the lib.
     key = None
     if args.key:
-      key = util.prompt_import_rsa_key_from_file(args.key)
+      key = util.prompt_import_private_key_from_file(args.key, args.key_type)
```
vladimir-v-diaz: Like my comment in https://github.com/in-toto/in-toto/pull/209/files#r192196029, why is it prompting for a passphrase for the default action.  What if the private key file is unencrypted?

in_toto/in_toto_record.py
yashsriv: This doesn't always prompt. It first checks if it can open without a password and if not, then only prompt the user.

tox.ini
yashsriv: Bandit is a linter? It doesn't run on specific tests. What we are doing right now is recursively running it on all files in the in_toto folder. The rule I'm skipping is this one which itself claims to throw a lot of false positives:
https://docs.openstack.org/bandit/latest/plugins/b107_hardcoded_password_funcdef.html

I'm adding a note regarding that.

in_toto/in_toto_sign.py
```diff
@@ -316,6 +327,13 @@ def main():
       "Path(s) to PEM formatted key file(s), used to sign the passed link or"
       " layout metadata or to verify its signatures."))
 
+  parser.add_argument("-t", "--key-type", dest="key_type",
+      type=str, choices=in_toto.util.SUPPORTED_KEY_TYPES,
+      nargs="+", help=(
+      "Specify the key-type of the keys specified by the '--key'"
```
vladimir-v-diaz, line 338: Some users might misinterpret the help description to mean that one single key type is expected for one or more keys.

Is the following invocation of in-toto-sign supported?  I'm not sure the help description makes it clear enough that it is.
```
$ in-toto-sign -k key1 key2 key3 -t rsa ed25519 rsa
```

tox.ini
vladimir-v-diaz: Bandit is a security linter, if I'm not mistaken.  It says so on their project page: "A security linter from OpenStack Security"

https://pypi.org/project/bandit/

I was recommending that you run Bandit + a profile on specific tests that need to avoid B107, or whatever.  For instance,

```
$ bandit tests/test_in_toto_verify.py -p ignore_pw_test
(other bandit invocations go here, with perhaps different profiles)
```

in_toto/in_toto_sign.py
```diff
@@ -151,8 +151,18 @@ def _sign_and_dump_metadata(metadata, args):
...
-      for key_path in args.key:
-        key = util.prompt_import_rsa_key_from_file(key_path)
+
+      if args.key_type is None:
+        args.key_type = [util.KEY_TYPE_RSA] * len(args.key)
+
+      if len(args.key_type) != len(args.key):
+        raise securesystemslib.exceptions.FormatError(
+          "number of key_types should match with the number"
+          " of keys specified")
```
vladimir-v-diaz, line 167: The help description for --key-type states that "rsa" is used if one isn't given.  If more than one --key is given, is the user forced to specify all the key types?  I'm not sure that's clear from the help description.

in_toto/in_toto_record.py
vladimir-v-diaz: Oh, the function name `prompt_import_from_private_key()` led me to believe that it does.

This is unexpected because it doesn't quite match the other similar functions:
in_toto.util.generate_and_write_rsa_keypair()
in_toto.util.prompt_generate_and_write_rsa_keypair()

tox.ini
vladimir-v-diaz: Note: You don't have to use profiles if you wish, but I only recommend it so that we don't ignore the B107 test across the entire codebase.  I'm not sure if there is a better way to do it, but using profiles is one possible way to deal with an overreaching --skip.

in_toto/in_toto_record.py
yashsriv: I was using the already existing function name. What do you recommend? `try_import_*` ?

vladimir-v-diaz: I guess it's fine to just call it `import_from_private_key()` (or if you want to be verbose, `import_private_or_encrypted_key()`), but it might be better to confirm with @SantiagoTorres.

yashsriv:
@vladimir-v-diaz I have updated the following stuff:
* Fix bandit test to only skip B107 for in_toto/util.py
* Improve documentation of key-type for in_toto_sign.py

tox.ini
```diff
@@ -7,6 +7,9 @@
 [tox]
 envlist = py{27,34,35,36}
 
+[flake8]
+ignore = E111, E114
+
```
vladimir-v-diaz: You will need to add flake8 to tox.ini's dependencies entry.

However, I'm not sure why flake8 is appearing here all of a sudden. 
 Doesn't in-toto use Pylint?

tox.ini
```diff
@@ -34,10 +34,11 @@ commands =
     # https://docs.openstack.org/bandit/latest/blacklists/blacklist_imports.html#b404-import-subprocess
     # https://docs.openstack.org/bandit/latest/plugins/subprocess_popen_with_shell_equals_true.html
     # https://docs.openstack.org/bandit/latest/plugins/subprocess_without_shell_equals_true.html
-    # Note: We exclude tests that fail on specifying the default value of a variable resembling a password.
-    # We actually never use hardcoded passwords.
+    bandit -r in_toto --skip B404,B603 -x in_toto/util.py
```
vladimir-v-diaz: 👍 
This is approach is better/simpler than using Bandit's profiles.

tox.ini
yashsriv: Oversight on my part. I use flake8 personally and I had added these settings to prevent errors. I was careful not to commit them (till now) :sweat_smile:

in_toto/util.py
```diff
@@ -5,58 +5,42 @@
...
 
-  with open(filepath, "wb") as fo_private:
-    fo_private.write(private_pem.encode("utf-8"))
+def generate_and_write_ed25519_keypair(filepath, password=""):
+  """
+  Calls securesystemslib.interface.generate_and_write_ed25519_keypair with
+  a default password of "" so as to prevent the library from prompting for
+  password. If prompting is needed, it's already handled here and then
+  forwarded.
+  """
```
vladimir-v-diaz: Thanks for including the docstring and for providing good info!  The only other advice I can give is to specify the exceptions that this function raises, expected return values, etc.

We also have a guideline for docstrings [here](
https://github.com/secure-systems-lab/code-style-guidelines#documentation-strings).

in_toto/util.py
```diff
@@ -103,12 +87,32 @@ def import_rsa_key_from_file(filepath, password=None):
   return rsa_key
 
 
-def import_rsa_public_keys_from_files_as_dict(filepaths):
+def import_public_keys_from_files_as_dict(filepaths, key_types=None):
   """Takes a list of filepaths to RSA public keys and returns them as a
   dictionary conformant with securesystemslib.formats.KEYDICT_SCHEMA."""
+  # are key_types needed?
```
vladimir-v-diaz: Good point.  In the future, we'll revise the SSLib functions to better handle key files.

in_toto/util.py
```diff
@@ -134,12 +138,44 @@ def prompt_password(prompt="Enter password: "):
...
+  try:
+    if key_type == KEY_TYPE_ED25519:
+      key = prompt_import_ed25519_privatekey_from_file(filepath)
+    elif key_type == KEY_TYPE_RSA:
+      key = prompt_import_rsa_key_from_file(filepath)
+    else:  # pragma: no cover
+      # This branch is never possible as argparse already checks valid keys
+      # via the choices parameter.
+      raise TypeError('Unsupported keytype: ' + str(key_type))
+  except:
```
vladimir-v-diaz: We should avoid using a bare except.  Our code style guidelines provides more info on why to avoid them.
https://github.com/secure-systems-lab/code-style-guidelines#use-class-based-exceptions

in_toto/util.py
```diff
@@ -134,12 +138,44 @@ def prompt_password(prompt="Enter password: "):
...
+    raise TypeError('Either invalid key type (neither ED25519 nor RSA), '
```
vladimir-v-diaz: Should we raise an in-toto-specific exception here?  It would also be helpful to document the expected exceptions in the doc string.

vladimir-v-diaz: Instead of explicitly listing the supported key types, I suggest that you append `in-toto.util.KEY_TYPES` (or the equivalent).  This alternate approach might better handle future changes to the supported key types.

in_toto/util.py
```diff
@@ -134,12 +138,44 @@ def prompt_password(prompt="Enter password: "):
...
+        'or wrong password.')
+
+  return key
+
+
+def prompt_import_ed25519_privatekey_from_file(filepath):
+  """Tries to load an Ed25519 private key without password. If a CryptoError
+  occurs, prompts the user for a password and trys to load the the key again.
```
vladimir-v-diaz: nit: "trys" and duplicate "the".

tests/test_in_toto_keygen.py
```diff
@@ -63,7 +64,14 @@ def test_main_optional_args(self):
       patch("getpass.getpass", return_value=password), self.assertRaises(
       SystemExit):
       in_toto_keygen_main()
-    with patch.object(sys, 'argv', args + ["-p", "bob", "3072"]), \
+    with patch.object(sys, 'argv', args + ["-t", "ed25519", "bob"]), \
+      self.assertRaises(SystemExit):
+      in_toto_keygen_main()
+    with patch.object(sys, 'argv', args + ["-p", "-t", "ed25519", "bob"]), \
+      patch("getpass.getpass", return_value=password), self.assertRaises(
```
vladimir-v-diaz, line 75: Is there a test where "rsa" is explicitly specified?

vladimir-v-diaz (commented):

in_toto/in_toto_record.py
SantiagoTorres: > I guess it's fine to just call it import_from_private_key()

Yeah, I think that `import_private_key_from_file` would do it.

yashsriv:
@vladimir-v-diaz I think I have fixed most of the issues in my recent commits. ptal and if its ok, I'll squash all my commits into one to avoid cluttering the git log.

P.S.: Any idea why tests fail randomly on travis?

vladimir-v-diaz:
You can go ahead and squash the commits.

The unit tests are randomly failing due to unexpected GPG behavior.  See https://github.com/in-toto/in-toto/issues/171.

trishankatdatadog:
Very good-looking PR, guys, thanks in particular to @yashsriv for all his hard work in so little time!

tox.ini
```diff
@@ -15,6 +15,10 @@ deps =
     bandit
 
 commands =
+    # Need master version of securesystemslib till this commit is in a release:
+    # https://github.com/secure-systems-lab/securesystemslib/commit/ba5f52118d308d2bc67ad11b5947a4ea7ef8f2ed
```
vladimir-v-diaz: I released v0.11.2, which contains the commits needed by this pull request.  Can you update this line to use v.0.11.2?  Let me know if you encounter any issues.
https://github.com/secure-systems-lab/securesystemslib/releases/tag/sslibv0.11.2

vladimir-v-diaz: A default installation of v0.11.2 only supports ed25519 in pure python.  To enable rsa, ecdsa, c extensions, etc.,  you need to install the optional dependencies...

```
pip install securesystemslib[crypto]
pip install securesystemslib[pynacl]
```

vladimir-v-diaz: @yashsriv You don't have to worry about getting `in-toto` to work in pure Python.  For now, you just have to update the requirements.txt and setup.py files to include the `crypto` and `pynacl`.

In setup.py, it's something like:
```
setup(
    ...
    install_requires=["securesystemslib[crypto]"],
    ...
)
```

In requirements.txt, it's something like:

```
securesystemslib[security] == 0.11.2 
```

yashsriv:
@SantiagoTorres should I update the debian package stuff as well?

SantiagoTorres:
Thanks everyone!

PR #218, opened 2018-09-14
Assume and test that gpg2 exists. Otherwise, assume gpg exists.

trishankatdatadog:
**Fixes issue #**:

N/A.

On systems such as Debian / Ubuntu, or Fedora, we want to use `gpg2` if it exists. Otherwise, assume `gpg` exists (which is the norm on macOS and Windows).

Cc: @ofek @SantiagoTorres 

requirements**:


SantiagoTorres:
This is looks simpler and kosher enough. Thanks!

PR #211, opened 2018-09-06
ENH: add windows support

RubyLiu206:
Windows support was added mostly by:

- Fixing path separator issues
- Making sure test executables existed within a regular windows install
- Adding support python programs if this wasn't the case.

Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #**:

**Description of the changes being introduced by the pull request**:

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #225, opened 2018-09-20
Fix path normalization for Windows

ofek:
https://github.com/in-toto/in-toto/pull/211 missed things


trishankatdatadog:
@ofek is the man

Issue #219, opened 2018-09-18
subprocess.check_output()


trishankatdatadog:
**Description of issue or feature request**:

We should use [subprocess.checkoutput()](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) by default. Using `subprocessPopen()` without checking return codes sometimes leads to misleading errors (such as apparently malformed GPG signature packets, when actually GPG did not run as intended, and return the expected output).

This also means that we should just install [subprocess32](https://pypi.org/project/subprocess32/) by default.

**Current behavior**:

We use `subprocess.Popen()`.

**Expected behavior**:

We should use `subprocess.check_output()`.

trishankatdatadog:
I have a PR for this I'll send when I figure out why git inside my Fedora VM breaks when SSHing over NAT but not bridge...

PR #227, opened 2018-09-26
use subprocess32 on python2 by default

trishankatdatadog:
that is provided, the better.

**Fixes issue #**: #219 

- install subprocess32 by default
- always check error code when calling gpg
- hide stderr from users
- always timeout when calling subprocesses

requirements**:

### DIFF
--- a/in_toto/gpg/constants.py
+++ b/in_toto/gpg/constants.py
@@ -16,11 +16,10 @@
   handling
 """
 
-import shlex
-import subprocess
-
 import in_toto.gpg.rsa as rsa
 import in_toto.gpg.dsa as dsa
+
+import in_toto.process as process
 
 # By default, we assume and test that gpg2 exists. Otherwise, we assume gpg
 # exists.
@@ -30,7 +29,7 @@ GPG_VERSION_COMMAND = GPG_COMMAND + " --version"
 FULLY_SUPPORTED_MIN_VERSION = "2.1.0"
 
 try:
-  subprocess.check_call(shlex.split(GPG_VERSION_COMMAND))
+  process.check_call_no_stdout_no_stderr(GPG_VERSION_COMMAND)
 except OSError: # pragma: no cover
   GPG_COMMAND = "gpg"
   GPG_VERSION_COMMAND = GPG_COMMAND + " --version"
--- a/in_toto/gpg/constants.py
+++ b/in_toto/gpg/constants.py
@@ -29,7 +29,8 @@ GPG_VERSION_COMMAND = GPG_COMMAND + " --version"
 FULLY_SUPPORTED_MIN_VERSION = "2.1.0"
 
 try:
-  process.check_call_no_stdout_no_stderr(GPG_VERSION_COMMAND)
+  process.run(GPG_VERSION_COMMAND, stdout=process.DEVNULL,
+    stderr=process.DEVNULL)
 except OSError: # pragma: no cover
   GPG_COMMAND = "gpg"
   GPG_VERSION_COMMAND = GPG_COMMAND + " --version"
--- a/in_toto/process.py
+++ b/in_toto/process.py
@@ -12,44 +12,20 @@ else: # pragma: no cover
 
 import in_toto.formats as formats
 
+
+# Constants.
 from in_toto.settings import SUBPROCESS_TIMEOUT
+DEVNULL = subprocess.DEVNULL
+PIPE = subprocess.PIPE
 
 
 # Inherits from in_toto base logger (c.f. in_toto.log)
 log = logging.getLogger(__name__)
 
 
-def shlex_split(cmd):
-  """
-  <Purpose>
-    Splits a string specifying a command and its argument into a list of
-    substrings, if necessary.
-
-  <Arguments>
-    cmd:
-            The command and its arguments. (str | list)
-
-  <Exceptions>
-    securesystemslib.exceptions.FormatError:
-            If the cmd as list does not match
-            in_toto.formats.LIST_OF_ANY_STRING_SCHEMA.
-
-  <Side Effects>
-    None.
-
-  <Returns>
-    A list of strings matching in_toto.formats.LIST_OF_ANY_STRING_SCHEMA.
-
-  """
-  if isinstance(cmd, six.string_types):
-    cmd = shlex.split(cmd)
-  else:
-    formats.LIST_OF_ANY_STRING_SCHEMA.check_match(cmd)
-  return cmd
-
-
 # TODO: Properly duplicate standard streams (issue #11)
-def run_pipe_stdout_pipe_stderr(cmd):
+def run(cmd, check=True, _input=None, stdin=None, stdout=None, stderr=None,
+        timeout=SUBPROCESS_TIMEOUT, universal_newlines=False):
   """
   <Purpose>
     Run the specified command, WITHOUT writing to standard input, and WITHOUT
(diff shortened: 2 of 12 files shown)


trishankatdatadog:
We should argue about whether `in_toto.process` has the right API, but I think this is a good enough start.

trishankatdatadog:
Does anyone know why the tests are failing on different Python versions on different platforms? I'm too tired to look at it now...

ofek:
Check out https://github.com/DataDog/integrations-core/blob/master/datadog_checks_dev/datadog_checks/dev/subprocess.py

trishankatdatadog:
@ofek That is very good. My only concern is that `subprocess32` (a backport of Python 3 `subprocess`) is needed to dispel concurrency bugs on Python 2. According to [Google](https://github.com/google/python-subprocess32/tree/560f1a92db18c2d2bebe4049756528ce827aa366#usage): "Users will thank you by not filing concurrency bugs."

And we have seen some weird Heisenbug issues running unit tests, with GPG passing one time, and failing another...

trishankatdatadog:
Oh, but @ofek, I guess you meant the API. Yeah, sure, I think the `datadog-checks-dev` subprocess API is a very good one, and we can borrow that :) I thought of making it more bulletproof by less configuration, but that's a good way to do it.

trishankatdatadog:
Simpler API now.

Sadly, GPG tests still fail mysteriously once in a while, but this is a known issue.

trishankatdatadog:
Does anyone know why subprocess does not work on Python 3.4 and / or AppVeyor?

ofek:
`run` is 3.5+

trishankatdatadog:
_doh_

Will think of something on Monday, when I am not jet-lagged...

in_toto/process.py
```diff
@@ -0,0 +1,124 @@
...
+  # The reason why we are not allowed to even specify stdin=None when input
+  # is specified is due to this overly stringent code in subprocess:
+  # https://github.com/google/python-subprocess32/blob/560f1a92db18c2d2bebe4049756528ce827aa366/subprocess32.py#L402
+  if _input:
+    log.debug('Ignoring stdin: '+str(stdin))
+    return subprocess.run(cmd, check=check, input=_input,
+      stdout=stdout, stderr=stderr, timeout=timeout,
+      universal_newlines=universal_newlines)
+  else:
+    log.debug('Ignoring input: '+str(input))
```
lukpueh, line 117: This is not a proper review yet, I just noticed something while skimming through your PR:
Above `input` should be an `_input`, right?

As for Python3.4 compatibility, why not use `check_output`, as you suggested in the corresponding issue (#219)? Let me know if you want me to take a more thorough look at it.

trishankatdatadog:
I don't yet understand why, but the in-toto tests sometimes fail nondeterministically on Windows 🤦‍♂️

trishankatdatadog:
It would be nice to get this merged next week, too, so we know exactly what went wrong if GPG fails for whatever reason.

in_toto/process.py
```diff
@@ -0,0 +1,122 @@
+import logging
```
lukpueh, line 1: Could you please add a [header block](https://github.com/secure-systems-lab/code-style-guidelines#documentation-strings) that explains the purpose of this module? AFAICS, this is an interface for python’s subprocess module that takes care of importing subprocess/subprocess32 (good!), namespaces `subprocess.PIPE` and `subprocess.DEVNULL`, and provides a wrapper for the `subprocess.run` function. Where the wrapper differs from the Python function in that it
- only passes the most common arguments. See how `subprocess.run` normally takes [arbitrary keyword arguments](https://github.com/python/cpython/blob/3.5/Lib/subprocess.py#L352) and [passes them on to `Popen`](https://github.com/python/cpython/blob/3.5/Lib/subprocess.py#L383) (why?)
- sets a configurable default for `timeout` argument (probably a good idea)
- sets a `True` default for `check` argument (why?)
- allows passing both `stdin` and `input` favoring latter, instead of raising an exception (why?)
- accepts a string (will be shlexed) as well as a list of command arguments and performs basic validation (seems convenient)


Overall I welcome the idea of providing a single interface for subprocess. Especially to handle things like the subprocess32 backport or the duplicating streams Issue (#160) in one place. But what do you think about above concerns? Maybe I’m missing something.

in_toto/process.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+# TODO: Properly duplicate standard streams (issue #11)
+def run(cmd, check=True, _input=None, stdin=None, stdout=None, stderr=None,
+        timeout=SUBPROCESS_TIMEOUT, universal_newlines=False):
+  """
+  <Purpose>
+    Run the specified command, WITHOUT writing to standard input, and WITHOUT
+    writing to standard output and error.  Note that we do NOT check whether
+    the command returned a non-zero code.
```
lukpueh, line 31: Is this accurate? Per default (`stdout = stdin = None`) this does write to stdout, streams, in addition the user can parametrize those streams.

setup.py
```diff
@@ -61,17 +61,17 @@
...
+    'Programming Language :: Python :: 3.7',
     'Programming Language :: Python :: Implementation :: CPython',
     'Topic :: Security',
     'Topic :: Software Development'
   ],
   packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests",
       "debian"]),
   install_requires=["six", "securesystemslib[crypto,pynacl]>=0.11.3", "attrs",
-                    "python-dateutil", "iso8601", "pathspec"],
+                    "python-dateutil", "iso8601", "pathspec", "subprocess32"],
```
lukpueh, line 74: I guess we could install subprocess32 only in Python 2 (e.g. `"subprocess32;python_version<'3'"`), since we don't use it in Python 3. If we do, please also update `requirements.txt`.

setup.py
```diff
@@ -61,17 +61,17 @@
     'Programming Language :: Python :: 2',
     'Programming Language :: Python :: 2.7',
     'Programming Language :: Python :: 3',
-    'Programming Language :: Python :: 3.4',
```
lukpueh, line 64: Are we fine with dropping support for 3.4? I don't have a strong opinion. However, `subprocess.run` is only a [tiny wrapper](https://github.com/python/cpython/blob/3.5/Lib/subprocess.py#L378-L399) around `subprocess.Popen`, so not sure if we really need it.

Also note that in order to duplicate streams (#160), we probably have to use `Popen` anyway, because AFAICS `run` only returns after the process has finished (blocking), but for stream duplication, we need to perform some read/write while the process is running.

lukpueh (commented):
Let me know what you think of my comments. If we decide to drop 3.4 in order to use `subprocess.run`, I probably wouldn't restrict the keyword arguments and write it something like this.

```python
def run(cmd, <kwarg_with_custom_default>=<custom_default>, **kwargs):
  cmd = validate_and_optionally_shlex_cmd_arg()
  validate_and_optionally_update_kwargs_as_desired()
  subprocess.run(cmd, **kwargs)
```

lukpueh:
We should also use your interface in [`in_toto.runtests.py`](https://github.com/in-toto/in-toto/blob/develop/tests/runtests.py#L24-L31)

in_toto/gpg/constants.py
```diff
@@ -30,7 +29,8 @@
 FULLY_SUPPORTED_MIN_VERSION = "2.1.0"
 
 try:
-  subprocess.check_call(shlex.split(GPG_VERSION_COMMAND))
+  process.run(GPG_VERSION_COMMAND, stdout=process.DEVNULL,
+    stderr=process.DEVNULL)
```
lukpueh, line 33: Just realized that this of course fixes the issue I mentioned yesterday that there is gpg noise on stdout if you run in-toto from the develop branch. 👍

setup.py
ofek: Please do this before merging

in_toto/gpg/constants.py
trishankatdatadog: That noise is still there (see the unit test), I need someone to help me look at where this is coming from, I don't have the headspace for this now...

lukpueh:
I took over @trishankatdatadog's subprocess branch and addressed the comments I made on this PR. The new PR can be found at #231. Please review/comment and merge there.

in_toto/gpg/constants.py
lukpueh: Oh yeah, you're right, but it was only in the unit tests (see [`runtests.py`](https://github.com/in-toto/in-toto/blob/a46d2f84703c00e08936960d1746bac1964488df/tests/runtests.py#L31)). I fixed also this in #231.

PR #229, opened 2018-09-26
Normalize line endings for cross-platform support

ofek:
https://github.com/secure-systems-lab/securesystemslib/pull/156


awwad:
With a new version of securesystemslib released, I re-ran these. Looks like the PR build passed, at least.

PR #231, opened 2018-10-04
Add interface for subprocess

lukpueh:
Takes over #227 and addresses most of the comments discussed there.

**Fixes issue #**:
#219

  Provide a common interface for Python's `subprocess` module to:

  - require the Py3 subprocess backport `subprocess32` on Python2 in a single place,
  - in-toto namespace subprocess constants (DEVNULL, PIPE) and
  - provide a custom `subprocess.run` wrapper

See commit messages for more details.

**THIS PR ALSO CHANGES SUPPORTED PYTHON VERSIONS**
Drops Python 3.4 support and adds Python 3.7. Former is for easy use of `subprocess.run` which was added in Python 3.5.
If we realize that we need 3.4, we would have to find an alternative for `subprocess.run` or provide a custom  implementation of it, which should not be hard because it is only a [tiny wrapper](https://github.com/python/cpython/blob/3.5/Lib/subprocess.py#L378-L399) around `subprocess.Popen`. However we would also need to adopt [`subprocess.CompletedProcess`](https://github.com/python/cpython/blob/3.5/Lib/subprocess.py#L319-L349) and [`subprocess.CalledProcessError`](https://github.com/python/cpython/blob/3.5/Lib/subprocess.py#L60-L85), which are new in Python 3.5 and required by `subprocess.run`.

requirements**:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ matrix:
           env: TOXENV=py36
         - python: "3.7"
           env: TOXENV=py37
+          # NOTE: temporary workaround; see https://github.com/travis-ci/travis-ci/issues/9815#issue-336465122
+          dist: xenial
+          sudo: true
 
 install:
     - pip install -U tox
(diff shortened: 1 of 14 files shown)


trishankatdatadog:
Thanks, @lukpueh, LGTM! ❤️

ofek (approved):
LGTM!

PR #230, opened 2018-10-02
Add normalize_line_endings argument to in_toto_run/record

lukpueh:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #**:

**Description of the changes being introduced by the pull request**:

Add optional boolean `normalize_line_endings` argument to `in_toto_run`, `in_toto_record_start` and `in_toto_record_stop` in `in_toto.runlib`.
If `True`, replaces windows and mac line endings with unix line endings before hashing the content of the passed files, for cross-platform support. Default is `False`.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

Issue #212, opened 2018-09-12
CLI should honor .gitignore by default


trishankatdatadog:
**Description of issue or feature request**:

CLI tools like `in-toto-run` should honor `.gitignore`, and not record files matching these patterns as materials.

**Current behavior**:

CLI tools do not honor `.gitignore` by default.

**Expected behavior**:

CLI tools honor `.gitignore` by default (unless overridden by the user).

aalsabag:
I'm looking to tackle this here. 
My plan is to first introduce a flag "exclude-with-gitignore" which is by default set to false.
This will use the existing pathspec package to ignore through the gitignore.

PR #232, opened 2018-10-05
Add file type detection (WIP)

lukpueh:
**DISCLAIMER:** This PR is WIP submitted for discussion/review (do not merge!) 

Port [`git`'s file type detection](https://github.com/git/git/blob/master/convert.c#L46-L104) used to decide whether line endings should be normalized when checking files in.

I tested it with the [test files](https://github.com/audreyr/binaryornot/tree/master/tests/files) of the Python `binaryornot` library and it seems to work.

**Todo:**
 - Actually use in [in-toto](https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L75-L76) or maybe even better in [securesystemslib](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/hash.py#L128) (see #230 or [secure-systems-lab/securesystemslib#156](https://github.com/secure-systems-lab/securesystemslib/pull/156/files))
- Write tests and see if it behaves like git's file type detection
- Consider performance. Looking at every byte of potentially many files (this is meant to be used with `in-toto-run` when recording artifacts) could be costly. Maybe it's enough to just [inspect a chunk](https://eli.thegreenplace.net/2011/10/19/perls-guess-if-file-is-text-or-binary-implemented-in-python/)?

**Future work:**
In the future we should also look at git configuration (`core.autocrlf` and `text` attribute in `.gitignores`), which may disable file type detection or line end normalization.


lukpueh:
#212 could be added to the module introduced in this PR

lukpueh:
Closing due to inactivity

Issue #228, opened 2018-09-26
Better error messages during verification of rules


trishankatdatadog:
**Description of issue or feature request**:

Presently, it takes a bit of detective work to understand exactly why artifacts failed to match given rules.

For example, supposing an artifact matched a filename pattern, but was supposed to MATCH a product or material from some other step, and was instead observed to be MODIFIED, then this is not clear at all, and requires manual inspection by the user.

At the same time, these diagnostic error messages should only be reported on stderr for artifacts that violate given rules.

**Current behavior**:

Error messages are misleading.

**Expected behavior**:

Error messages should be as precise as possible.

michizhou:
@trishankatdatadog I would be more than happy to take on this issue. I just need a bit more context in terms of which file (with full repo filepath) generates the error messages and any supporting advice or documentation that could help me make the messages more specific. Please let me know. Thanks!

lukpueh:
Artifact rules are processed in [in-toto/verifylib.py](https://github.com/in-toto/in-toto/blob/develop/in_toto/verifylib.py) (see `verify_item_rules` and `verify_*_rule` functions). Please also take a look #238 and in-toto/docs#11, which should give you a good overview of the current way artifact rules are processed.

michizhou:
@lukpueh @trishankatdatadog Just as a heads up, I just made a new PR that should at least partially resolve this issue. I also noticed that changing the `verify_delete_rule` function implementation (which is in alignment with the `FIXME` goals of in-toto/in-toto#204) is causing issues in `test_verifylib.py` since some of the tests in that file rely on RuleVerificationError exceptions being raised by the `verify_delete_rule` function. We might want to rethink changing the `verify_delete_rule` function implementation if that is the case. Feel free to share your thoughts on this and other issues.

michizhou:
@lukpueh My apologies for not getting back to you earlier, but I have recently addressed your most recent concerns on my prior PR (#242) though a new PR (#243) I created. Please look it over and let me know if you need me to make any further changes. Thanks!

michizhou:
@lukpueh I just pushed some recent changes that should give more useful contextual information for the tracebacks on rule processing in the verification error messages. Please let me know if there are any additional changes I would need to make in your review. Thanks!

lukpueh:
Fixed with #243 and #262

PR #242, opened 2018-11-08
Updated rule verification error messages and error workflow

michizhou:
that is provided, the better.

**Fixes issue #228**:

Modified the rule verification error messages to be more specific about which artifacts failed to match which given rules by listing out the artifacts that failed to be processed by any previous rules. Also restructured the error workflow so only the DISALLOW rule can fail overall rule verification, with the other rules only able to remove artifacts from source artifact queues on success or leave the queues unchanged on failure, in alignment with in-toto/in-toto#204.

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -563,11 +563,15 @@ def verify_all_steps_command_alignment(layout, chain_link_dict):
 def verify_match_rule(rule, source_artifacts_queue, source_artifacts, links):
   """
   <Purpose>
-    Verifies that for each queued source artifact filtered by the specified
-    source pattern there is a destination artifact filtered by the specified
-    destination pattern and they are equal in terms of path and file hash.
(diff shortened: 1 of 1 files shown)


lukpueh (changes requested):
Thanks for the PR! Here are a few general comments:

- Always update the upstream branch (in-toto/in-toto@develop) before you create your feature branch (michizhou/in-toto@update-rule-docs). There is no reason why this PR should contain a merge commit.

- Use a feature branch name that describes your feature. You are not updating the the docs but the error messages, right?

- The diff seems off, it contains changes that should already be in upstream develop, i.e. the doc changes that were merged with ac225cd918ce381512f7dfa71cb05b66263f969c.

- Don't change the rule processing (DELETE rule) just now. This is a different issue that we will address in a different PR.

- Consider the prescribed maximum line length in your commit messages.

I will follow up with a review once you have addressed above issues.

PR #243, opened 2018-11-08
Updated rule verification error messages and error workflow

michizhou:
that is provided, the better.

**Fixes issue #228**:

Modified the rule verification error messages to be more specific about which artifacts failed to match which given rules by listing out the artifacts that failed to be processed by any previous rules. Also restructured the error workflow so only the DISALLOW and DELETE rules are able to fail overall rule verification, with the other rules only able to remove artifacts from source artifact queues on success or leave the queues unchanged on failure, in alignment with #204.

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -28,7 +28,6 @@ import iso8601
 import fnmatch
 import six
 import logging
-import pprint
 from dateutil import tz
 
 import securesystemslib.exceptions
@@ -1052,7 +1051,7 @@ def verify_disallow_rule(rule, source_artifacts_queue):
   if len(matched_artifacts):
     raise RuleVerificationError("Rule '{0}' failed.\nRule pattern matches the"
         " following artifacts of the artifact queue, which is disallowed:\n"
-        " '{1}' \n'{2}'".format(" ".join(rule), matched_artifacts,
+        " {1} \n{2}".format(" ".join(rule), matched_artifacts,
             get_traceback_info()))
 
 
@@ -1113,7 +1112,7 @@ def verify_item_rules(source_name, source_type, rules, links):
         queue.
 
   <Side Effects>
-    None.
+    Clears and populates the global RULE_TRACE data structure.
 
   """
 
@@ -1350,14 +1349,14 @@ def get_traceback_info():
   traceback_str = "Trace for earlier '{0}' rules of item '{1}':\n".format(
       RULE_TRACE["type"], RULE_TRACE["name"])
 
-  traceback_str += "Initial state of queues:\n"
-  traceback_str += pprint.pformat(RULE_TRACE["trace"].pop(0)) + "\n"
-
-  for trace_entry in RULE_TRACE["trace"]:
-    traceback_str += "Queues after processing of rule '{0}':\n".format(
-        trace_entry["rule"])
-    queues = {k: trace_entry[k] for k in ('artifacts', 'materials', 'products')}
-    traceback_str += pprint.pformat(queues) + "\n"
+  for count, trace_entry in enumerate(RULE_TRACE["trace"]):
+    if count == 0:
+      traceback_str += "Initial state of queues:\n"
+    else:
+      traceback_str += "Queues after processing of rule '{0}':\n".format(
+          trace_entry["rule"])
+    for q_name in ["artifacts", "materials", "products"]:
+      traceback_str += "\t{}: {}\n".format(q_name, trace_entry[q_name])
 
   return traceback_str
 
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -49,6 +49,8 @@ import in_toto.rulelib
 # Inherits from in_toto base logger (c.f. in_toto.log)
 log = logging.getLogger(__name__)
 
+RULE_TRACEBACK_QUEUE = list()
+
 
 def _raise_on_bad_retval(return_value, command=None):
   """
@@ -1049,7 +1051,8 @@ def verify_disallow_rule(rule, source_artifacts_queue):
   if len(matched_artifacts):
     raise RuleVerificationError("Rule '{0}' failed, rule pattern matches the"
         " following artifacts of the artifact queue, which is disallowed:"
-        " '{1}' ".format(" ".join(rule), matched_artifacts))
+        " '{1}' . Here is the traceback for all earlier rule processing:\n"
+        " '{2}' ".format(" ".join(rule), matched_artifacts, RULE_TRACEBACK_QUEUE))
 
 
 def verify_item_rules(source_name, source_type, rules, links):
@@ -1149,9 +1152,11 @@ def verify_item_rules(source_name, source_type, rules, links):
     if rule_type == "match":
       source_artifacts_queue = verify_match_rule(
           rule, source_artifacts_queue, source_artifacts, links)
+      RULE_TRACEBACK_QUEUE.append(source_artifacts_queue)
 
     elif rule_type == "allow":
       source_artifacts_queue = verify_allow_rule(rule, source_artifacts_queue)
+      RULE_TRACEBACK_QUEUE.append(source_artifacts_queue)
 
     elif rule_type == "disallow":
       verify_disallow_rule(rule, source_artifacts_queue)
@@ -1162,15 +1167,17 @@ def verify_item_rules(source_name, source_type, rules, links):
     elif rule_type == "create":
       source_products_queue = verify_create_rule(
           rule, source_materials_queue, source_products_queue)
+      RULE_TRACEBACK_QUEUE.append(source_products_queue)
 
       # The create rule only updates the products_queue, which in turn
       # only affects the generic artifacts queue if source_type is "products"
       if source_type == "products":
         source_artifacts_queue = source_products_queue
-
+        
     elif rule_type == "delete":
       source_materials_queue = verify_delete_rule(
           rule, source_materials_queue, source_products_queue)
+      RULE_TRACEBACK_QUEUE.append(source_materials_queue)
 
       # The delete rule only updates the materials_queue, which in turn
       # only affects the generic artifacts queue if source_type is "materials"
@@ -1195,6 +1202,8 @@ def verify_item_rules(source_name, source_type, rules, links):
             rule, source_materials_queue, source_artifacts_queue,
             source_materials, source_products)
         source_artifacts_queue = source_products_queue
+
+      RULE_TRACEBACK_QUEUE.append(source_artifacts_queue)
 
 
 def verify_all_item_rules(items, links):
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1355,8 +1355,9 @@ def get_traceback_info():
 
   for trace_entry in RULE_TRACE["trace"]:
     traceback_str += "Queues after processing of rule '{0}':\n".format(
-        trace_entry.pop("rule"))
-    traceback_str += pprint.pformat(trace_entry) + "\n"
+        trace_entry["rule"])
+    queues = {k: trace_entry[k] for k in ('artifacts', 'materials', 'products')}
+    traceback_str += pprint.pformat(queues) + "\n"
 
   return traceback_str
 
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1050,11 +1050,16 @@ def verify_disallow_rule(rule, source_artifacts_queue):
       source_artifacts_queue, rule_data["pattern"])
 
   if len(matched_artifacts):
-    rule_traceback = pprint.pformat(RULE_TRACE)
+    rule_traceback = "Initial state of queues:\n"
+    rule_traceback += pprint.pformat(RULE_TRACE["trace"][0])
+    for num in range(1,len(RULE_TRACE["trace"])):
+      queue_prompt = "\nQueues after processing of rule '{0}':\n".format(num)
+      rule_traceback += queue_prompt + pprint.pformat(RULE_TRACE["trace"][num])
     raise RuleVerificationError("Rule '{0}' failed.\nRule pattern matches the"
         " following artifacts of the artifact queue, which is disallowed:\n"
-        " '{1}' \nHere is the traceback for all earlier rule processing:\n"
-        " '{2}' ".format(" ".join(rule), matched_artifacts, rule_traceback))
+        " '{1}' \nHere is the trace for earlier '{2}' rules of item '{3}':\n"
+        " '{4}' ".format(" ".join(rule), matched_artifacts, RULE_TRACE["type"],
+        RULE_TRACE["name"], rule_traceback))
 
 
 def verify_item_rules(source_name, source_type, rules, links):
@@ -1145,9 +1150,11 @@ def verify_item_rules(source_name, source_type, rules, links):
   # Initialize rule traceback dictionary
   RULE_TRACE["name"] = source_name
   RULE_TRACE["type"] = source_type
-  RULE_TRACE["trace"] = list()
+  RULE_TRACE["trace"] = {0: {'artifacts': source_artifacts_queue, 
+      'materials': source_materials_queue, 'products': source_products_queue}}
 
   # Apply (verify) all rule
+  rule_count = 1
   for rule in rules:
     results_dict = {}
 
@@ -1214,7 +1221,8 @@ def verify_item_rules(source_name, source_type, rules, links):
     results_dict["artifacts"] = source_artifacts_queue
     results_dict["materials"] = source_materials_queue
     results_dict["products"] = source_products_queue
-    RULE_TRACE["trace"].append(results_dict)
+    RULE_TRACE["trace"][rule_count] = results_dict
+    rule_count += 1
 
 
 def verify_all_item_rules(items, links):
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1050,16 +1050,10 @@ def verify_disallow_rule(rule, source_artifacts_queue):
       source_artifacts_queue, rule_data["pattern"])
 
   if len(matched_artifacts):
-    rule_traceback = "Initial state of queues:\n"
-    rule_traceback += pprint.pformat(RULE_TRACE["trace"][0])
-    for num in range(1,len(RULE_TRACE["trace"])):
-      curr_rnd = RULE_TRACE["trace"][num]
-      rule_name = curr_rnd["rule"]
-      queues = {k: curr_rnd[k] for k in ('artifacts', 'materials', 'products')}
-      q_prompt = "\nQueues after processing of rule '{0}':\n".format(rule_name)
-      rule_traceback += q_prompt + pprint.pformat(queues)
-    raise RuleVerificationError(print_error_message(rule, matched_artifacts,
-        RULE_TRACE, rule_traceback))
+    raise RuleVerificationError("Rule '{0}' failed.\nRule pattern matches the"
+        " following artifacts of the artifact queue, which is disallowed:\n"
+        " '{1}' \n'{2}'".format(" ".join(rule), matched_artifacts,
+            get_traceback_info()))
 
 
 def verify_item_rules(source_name, source_type, rules, links):
@@ -1332,30 +1326,14 @@ def verify_threshold_constraints(layout, chain_link_dict):
                     step_name=step.name, keyid=keyid)))
 
 
-def print_error_message(rule, matched_artifacts, rule_trace, error_traceback):
+def get_traceback_info():
   """
   <Purpose>
-    Formats the queue for the error message meant to be printed out when a
-    RuleVerificationError is raised using the rule that triggers the error,
-    mismatched artifacts, elements of the rule trace and the error traceback.
+    Retrieves traceback information for the error message meant to be printed 
+    out when a RuleVerificationError is raised using elements of the rule trace.
 
   <Arguments>
-    rule:
-            The rule that triggers the error at the time of its processing
-            due to being matched to existing artifacts.
-
-    matched_artifacts:
-            Artifacts with which the error-inducing rule has been matched to
-            within the current artifacts queue at time of processing.
-
-    rule_trace:
-            Rule trace that stores information about rule processing like the
-            name of the processed item and the type of the processed rule list.
-
-    error_traceback:
-            Traceback of all processed rules and the state of their queues
-            after each round of rule processing that is printed to provide a
-            comprehensive traceback for error handling.
+    None.
 
   <Exceptions>
     None.
@@ -1364,19 +1342,23 @@ def print_error_message(rule, matched_artifacts, rule_trace, error_traceback):
     None.
 
   <Returns>
-    A string formatted as an error message containing the name of the processed
-    item, type of the processed rule list, and a traceback for all processed 
-    rules with rule names and states of the queues after each processed rule.
+    A string containing a traceback for all processed rules with rule names 
+    and the states of the queues after each round of rule processing.
 
   """
 
-  error_message = ("Rule '{0}' failed.\nRule pattern matches the following"
-        " artifacts of the artifact queue, which is disallowed:\n '{1}'"
-        "  \nHere is the trace for earlier '{2}' rules of item '{3}':\n"
-        " '{4}' ").format(" ".join(rule), matched_artifacts, rule_trace["type"],
-        rule_trace["name"], error_traceback)
+  traceback_str = "Trace for earlier '{0}' rules of item '{1}':\n".format(
+      RULE_TRACE["type"], RULE_TRACE["name"])
 
-  return error_message
+  traceback_str += "Initial state of queues:\n"
+  traceback_str += pprint.pformat(RULE_TRACE["trace"].pop(0)) + "\n"
+
+  for trace_entry in RULE_TRACE["trace"]:
+    traceback_str += "Queues after processing of rule '{0}':\n".format(
+        trace_entry.pop("rule"))
+    traceback_str += pprint.pformat(trace_entry) + "\n"
+
+  return traceback_str
 
 
 def reduce_chain_links(chain_link_dict):
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1051,9 +1051,9 @@ def verify_disallow_rule(rule, source_artifacts_queue):
 
   if len(matched_artifacts):
     rule_traceback = pprint.pformat(RULE_TRACE)
-    raise RuleVerificationError("Rule '{0}' failed, rule pattern matches the"
-        " following artifacts of the artifact queue, which is disallowed:"
-        " '{1}' . Here is the traceback for all earlier rule processing:\n"
+    raise RuleVerificationError("Rule '{0}' failed.\nRule pattern matches the"
+        " following artifacts of the artifact queue, which is disallowed:\n"
+        " '{1}' \nHere is the traceback for all earlier rule processing:\n"
         " '{2}' ".format(" ".join(rule), matched_artifacts, rule_traceback))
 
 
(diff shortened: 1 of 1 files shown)


lukpueh:
Please take another look at my comments in https://github.com/in-toto/in-toto/pull/242#pullrequestreview-172901029 and try to address all issues I pointed out. Let me know if something is unclear.

michizhou:
@lukpueh I just did take another look, and I have addressed all the issues you raised in [#242 (review)](https://github.com/in-toto/in-toto/pull/242#pullrequestreview-172901029). The merge commit in this PR resolves a merge conflict that arose after rebase. I tried to squash this commit together with the first one, but it's not possible. Please review this PR accordingly. Thanks.

lukpueh (changes requested):
Please address my concerns raised in https://github.com/in-toto/in-toto/pull/242#pullrequestreview-172901029:

> ... There is no reason why this PR should contain a merge commit.

You still have a merge commit in this PR (see https://github.com/in-toto/in-toto/pull/243/commits). It is possible to get rid it.

> Don't change the rule processing (DELETE rule) just now. This is a different issue that we will address in a different PR.

You are changing the rule processing logic (see https://github.com/in-toto/in-toto/pull/243/files#diff-a3d946f0bec6db9af878388e7086fbccL675), which is also the reason why unit tests fail.

lukpueh:
Re merge commits in your PRs (plus some more background):

You never need to perform any merges locally. Before you start to work on a new feature, always make sure that the main branch of [your fork is in sync with the main branch of the upstream repo](https://help.github.com/articles/syncing-a-fork/). (Note, we call our main branch `develop` and not `master`.

Updating your fork's main branch should never require a *regular merge* but instead work with a *fast forward merge* (keeping the history linear).

If the main branch evolves while you are working on your feature, your feature branch's original base (originally determined by `git checkout -b <feature branch name>`) will become outdated. This is not necessarily a problem, only if the main branch's updates conflict with the changes you made on your feature branch.

In order to update your feature branch's base, I want you to always use (being on your feature branch) `git rebase [--onto] <updated main branch name>` and fix conflicts manually as they appear.

in_toto/verifylib.py
```diff
@@ -797,8 +797,7 @@ def verify_create_rule(rule, source_materials_queue, source_products_queue):
 
   """
   rule_data = in_toto.rulelib.unpack_rule(rule)
-
-
+
```
lukpueh: Please remove trailing whitespace.

in_toto/verifylib.py
```diff
@@ -1026,8 +1025,9 @@ def verify_disallow_rule(rule, source_artifacts_queue):
       source_artifacts_queue, rule_data["pattern"])
 
   if len(matched_artifacts):
-    raise RuleVerificationError("Rule '{0}' failed, pattern matched disallowed"
-        " artifacts: '{1}' ".format(" ".join(rule), matched_artifacts))
+    raise RuleVerificationError("Rule '{0}' failed, rule pattern matches the"
+        " following artifacts of the artifact queue, which is disallowed:"
+        " '{1}' ".format(" ".join(rule), matched_artifacts))
```
lukpueh: TBH, I don't think this changes a lot. It does rephrase the error message slightly, but does not add any information content as requested in #228.

Since `len(matched_artifacts)`, which decides whether `RuleVerificationError` is raised or not, heavily depends on earlier rule processing, it might be helpful to cache the results of the earlier rule processing and add them to the error message. What do you think? (cc'ing @trishankatdatadog)

lukpueh (changes requested):

in_toto/verifylib.py
michizhou: @lukpueh I think that would be a good idea. How would we fetch the results of the earlier rule processing? Should we just make a call to one of the `verify_*_rule` functions? If so, which one do we fetch from and would we use its `matched_artifacts`?

lukpueh: The calls to `verify_*_rule` are already happening. You'd have to cache their results and present it to the user in case an error occurs. You can either use a global variable which you reset in each run of `verify_*_rule`. Or you keep such a variable local in `verify_item_rules`, append the results from the calls to the `verify_*_rule` functions and always pass it to `verify_delete_rule` so that it can be displayed there in case of an error.

In general I am not a fan of using global variables, but it might be okay in this case. Use whatever seems cleaner to you.

in_toto/verifylib.py
```diff
@@ -672,9 +663,7 @@ def verify_match_rule(rule, source_artifacts_queue, source_artifacts, links):
   try:
     dest_link = links[dest_name]
   except KeyError:
-    raise RuleVerificationError("Rule '{rule}' failed, destination link"
-        " '{dest_link}' not found in link dictionary".format(
-            rule=" ".join(rule), dest_link=dest_name))
+    return source_artifacts_queue
```
lukpueh: It seems to me that you had reverted this unwanted change in a previous version of this PR. Did you force push back in? (Please also leave in the fixme comment above (L576-584).

michizhou: I thought that I had done so when I forced pushed to get rid of the merge commit. I have since rectified my mistake and rebased accordingly.

in_toto/verifylib.py
michizhou: @lukpueh Okay I think I understand the gist of what you are saying. I just committed more changes that should hopefully implement the caching of earlier rule processing results. I am a bit unsure about whether I got the right idea from your comments so please let me know if you this is not what you had in mind.

lukpueh: Yes, this looks good pretty much like what I had in mind. However, you should not add extra calls to `verify_disallow_rule`, this completely changes the rule processing logic, which we don't want.

Also, just adding the updated queues to your traceback list might not be informative enough.
Please try to add some more context. You don't have to code it right away, maybe you can start by jotting down a data structure here (can be pseudo-code) with all the info you consider relevant for someone you receives the error.

michizhou: @lukpueh Okay I get your point. Just to clarify, what do the updated queues represent? Do they contain all the artifact patterns that were successfully or were failed to be processed? Would making the traceback list into a traceback dictionary (with the keys being the rules as they are called and the values as the updated queues) provide enough context? Or should I add even more info?

By the way I already removed the extra calls to `verify_disallow_rule` once I realized that they were causing the builds to fail, so there should be no logic issues now.

lukpueh: I like your thinking, adding the rules is definitely a good idea. I would, however, keep the list as base structure to preserve the order. But it could be a list of lists or dictionaries. Please take a closer look at the various `verify_*_rule` functions and the dispatching `verify_item_rules` function to really understand what they are doing and what information is relevant and should therefor be propagated.

trishankatdatadog: Anything to improve the error message will be great, guys! I don't think this PR alone should be enough to close #228, but it's definitely a good start 👍

michizhou:
@lukpueh Sorry for bothering you again, but I just wanted to follow up with you on the recent changes I made toward modifying the contents of the rule verification error messages from a week ago. I would greatly appreciate it if you could review my changes at your earliest convenience. Thanks!

in_toto/verifylib.py
```diff
@@ -572,7 +574,7 @@ def verify_match_rule(rule, source_artifacts_queue, source_artifacts, links):
     The rule only modifies the source artifacts queue, by removing artifacts
     that were successfully consumed by the rule, i.e. if there was a match with
     a target artifact.
-
+
```
lukpueh: Please remove trailing whitespace.

in_toto/verifylib.py
```diff
@@ -1140,15 +1147,19 @@ def verify_item_rules(source_name, source_type, rules, links):
...
           rule, source_materials_queue, source_products_queue)
+      results_dict = {'Rule Used': rule, 'Remaining Products': source_products_queue}
+      RULE_TRACEBACK_QUEUE.append(results_dict)
 
       # The create rule only updates the products_queue, which in turn
       # only affects the generic artifacts queue if source_type is "products"
       if source_type == "products":
         source_artifacts_queue = source_products_queue
-
+
```
lukpueh: Please remove trailing whitespace.

lukpueh (changes requested):
Looks good overall. A few minor suggestions:
- Reset the queue each time you enter `verify_item_rules`.
- It might be helpful to add some additional context, i.e. what step/inspection you're on (`source_name`) and whether it's a batch of material rules or product rules (`source_type`).
- For maximum transparency, I'd just always add all three available queues, after having processed any rule. So `RULE_TRACEBACK_QUEUE` could look something like:
```python
{
  "name": source_name,
  "type": source_type,
  "trace": [{
      "rule": rule,
      "artifacts": source_artifacts_queue,
      "materials": source_materials_queue,
      "products": source_products_queue
    }, ...]
}
```

- You can then write a little function that formats that data in a nice way, which you call before passing it to the exception message.

- Minor nitpick: the name `RULE_TRACEBACK_QUEUE` suggests that it holds a `queue` data structure, I think `RULE_QUEUE_TRACEBACK` would be better, or just `RULE_TRACE`. :)

- Please also make sure to not exceed [line length](https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length).

- And remove trailing whitespace.

michizhou:
@lukpueh I recently made changes to the code in order to reflect the suggestions from your review. If you have the chance either this weekend or early next week, I would greatly appreciate it if you could look over these most recent updates and share your thoughts. 

I also want to apologize for all the extra force-pushing earlier yesterday; those were from my attempts to rebase my feature branch in order to sync up with the base branch (i.e. `in-toto:develop`). The last one was done in order to fix a line length formatting error in the commit message for my last commit. I promise to greatly limit such force-pushing in the future. Thanks for understanding.

michizhou:
@lukpueh I just wanted to follow up with you on the fact that I recently pushed new changes that help address the comments of your review last week. When you have the chance, I would appreciate it if you could look over these most recent changes and share your thoughts. Thanks!

in_toto/verifylib.py
```diff
@@ -1112,6 +1120,11 @@ def verify_item_rules(source_name, source_type, rules, links):
         " one of 'materials' or 'products.'\n"
         "Got:\n\t'{}'".format(source_type))
 
+  # Initialize rule traceback dictionary
+  RULE_TRACE["name"] = source_name
+  RULE_TRACE["type"] = source_type
+  RULE_TRACE["trace"] = list()
+  results_dict = {}
```
lukpueh: You must re-initialize `results_dict` inside the loop for each iteration, otherwise you'll update the same dict over and over again and all elements in your list (references) will point to the same dict:

```python
# What you are doing
foo = {}
baz = []
for i in range(5):
  foo["bar"] = i
  baz.append(foo)
print(baz)
# Output: [{'bar': 4}, {'bar': 4}, {'bar': 4}, {'bar': 4}, {'bar': 4}]
```
vs.

```python
# What you should do
baz = []
for i in range(5):
  foo = {}
  foo["bar"] = i
  baz.append(foo)
print(baz)
# Output: [{'bar': 0}, {'bar': 1}, {'bar': 2}, {'bar': 3}, {'bar': 4}]
```

lukpueh (changes requested):
Thanks for the updates! Please fix the issue I pointed out below. Also, have you tested what the `RuleVerificationError` message looks like? Do you think it is readable/helpful? Can you post a sample output? 

You can take a look at `test_verifylib.TestVerifyAllItemRules` for a minimal rule processing example.

michizhou:
@lukpueh Thanks for the quick review. I have fixed the issue you pointed out and have modified the organization of the content in the error messages for improved readability from the user perspective.

I have also included a sample output of the error message (with most recent changes) below:
```python
in_toto.exceptions.RuleVerificationError: Rule 'DISALLOW in_toto/gpg' failed.
Rule pattern matches the following artifacts of the artifact queue, which is disallowed:
 '['in_toto/gpg']' 
Here is the traceback for all earlier rule processing:
 '{}' 
```
I look forward to your next review of my changes. Please inform me of any other concerns. Thanks!

lukpueh:
Thanks for the fix. Regarding the sample output, it does not really help us to assess the readability/usefulness of your additions. I suggest you try triggering an exception that shows a non-empty trace.

michizhou:
@lukpueh It was my pleasure. In light of your concerns, I have included a more comprehensive sample output for you to analyze in order to better assess the readability/usefulness of my changes to how the rule verification error messages are organized in content. Here is the sample output:

```python
in_toto.exceptions.RuleVerificationError: Rule 'DISALLOW foo*' failed.
Rule pattern matches the following artifacts of the artifact queue, which is disallowed:
 '['foo', 'foobar']' 
Here is the traceback for all earlier rule processing:
 '{'name': 'item',
 'trace': [{'artifacts': ['bar'],
            'materials': ['bar'],
            'products': ['baz', 'foo', 'bar'],
            'rule': ['DELETE', 'foobar']},
           {'artifacts': ['bar'],
            'materials': ['bar'],
            'products': ['foo', 'bar'],
            'rule': ['CREATE', 'baz']},
           {'artifacts': ['bar'],
            'materials': ['bar'],
            'products': ['foo'],
            'rule': ['MODIFY', 'bar']},
           {'artifacts': ['bar'],
            'materials': ['bar'],
            'products': ['foo'],
            'rule': ['MATCH', 'foo', 'WITH', 'MATERIALS', 'FROM', 'item']},
           {'artifacts': ['bar'],
            'materials': ['bar'],
            'products': ['foo'],
            'rule': ['DISALLOW', 'barfoo']},
           {'artifacts': [],
            'materials': ['bar'],
            'products': ['foo'],
            'rule': ['ALLOW', '*']}],
 'type': 'materials'}' 
```
Please share your thoughts on this matter and let me know if any more changes are needed. Thanks!

trishankatdatadog:
Hi everyone, I think I already see all this by turning on debugging output
and looking at the stack trace, Not to discount the work, but I don't
understand it: what's new here?

michizhou:
@trishankatdatadog There is some new additional context that has been added with regards to rule processing. Specifically, the new information includes the following: 
- What step/inspection you're on (`name`)
- Whether it's a batch of material rules or product rules (`type`)
- After having processed a rule, all three available queues (`artifacts`, `materials`, `products`) are added to the trace along with the respective rule used

If you believe there should be any additional pieces of information or if the existing information as seen in the sample output should be modified (i.e. different names for dictionary keys or values), please let me know so that I can make improvements to the error messages. Thanks for the feedback!

lukpueh:
@trishankatdatadog, you're right of course that we could get the same insights by writing all this information to the log and asking the caller to set the corresponding log level in order to make more sense of a rule verification exception. This is a UX design decision and I think Michael's approach fits well with the initial request in #228. What do the others think? (@SantiagoTorres, @trishankatdatadog, @michizhou).

That being said, I think we could still try to make the trace a bit clearer for the user, e.g. by
- putting the name of the processed item and the type of the processed rule list before the trace,
- adding the initial state of the queues to the trace,
- making it clear that the queues show the artifacts after the rule was processed.

...so that it provides the following content in a way that is easy to parse by the user (not word by word!!):
```
Trace for preceding <rule type> rules of item <item name>:
Initial state of queues
State of queues after processing rule 0
State of queues after processing rule 1
...
State of queues after processing rule n-1
```

lukpueh:
@michizhou, I'm curious btw how you got the above output. Because the rule before the failing `DISALLOW foo*` rule is an `ALLOW *` rule, which leaves the artifact queue empty. So I don't really understand where the matched `foo` and `foobar` are coming from.

michizhou:
@lukpueh I got the output with `bpython` by first processing the code below, derived from the `setUp` method and sample list of material rules from the `TestVerifyItemRules` class of `test_verifylib.py`:

```python
sha256_1 = \
    "d65165279105ca6773180500688df4bdc69a2c7b771752f0a46ef120b7fd8ec3"
         
     
sha256_2 = \
	"cfdaaf1ab2e4661952a9dec5e8fa3c360c1b06b1a073e8493a7c46d2af8c504b"
         
     
from in_toto.models.metadata import Metablock
from in_toto.models.link import Link, FILENAME_FORMAT
item_name = "item"
links = {
        "item": Metablock(signed=Link(name="item",
            materials={
                "foo": {"sha256": sha256_1},
                "foobar": {"sha256": sha256_1},
                 "bar": {"sha256": sha256_1}
             },
             products={
                 "baz" : {"sha256": sha256_1},
                 "foo": {"sha256": sha256_1},
                 "bar": {"sha256": sha256_2}
             }
        ))
      }
       
   
rules = [
      ["DELETE", "foobar"],
      ["CREATE", "baz"],
      ["MODIFY", "bar"],
      ["MATCH", "foo", "WITH", "MATERIALS", "FROM", "item"], # match with self
      ["DISALLOW", "barfoo"],
      ["ALLOW", "*"],
    ]
```

I then called `v.verify_item_rules(item_name, "materials", rules, links)` to populate the rule trace (`v.RULE_TRACE`) before executing the following lines to introduce the failing `DISALLOW foo*` rule:

```python
rule = ["DISALLOW", "foo*"]
queue = ["foo", "bar", "foobar"]
v.verify_disallow_rule(rule, queue)
```

The output after these commands were run helped to generate the above output for the rule trace. I might also need to update the error message to reflect the fact that rule verification could fail if a rule pattern matches elements of the other queues (depending on the rule), not just the artifact queue, if you think that it would make the content of the error messages more accurate.

michizhou:
@SantiagoTorres Do you think the rule trace and the associated error message in its current state would suffice? Or should I try to reformat the rule trace according to the suggestions of @lukpueh?

in_toto/verifylib.py
```diff
@@ -1048,8 +1050,16 @@ def verify_disallow_rule(rule, source_artifacts_queue):
       source_artifacts_queue, rule_data["pattern"])
 
   if len(matched_artifacts):
-    raise RuleVerificationError("Rule '{0}' failed, pattern matched disallowed"
-        " artifacts: '{1}' ".format(" ".join(rule), matched_artifacts))
+    rule_traceback = "Initial state of queues:\n"
+    rule_traceback += pprint.pformat(RULE_TRACE["trace"][0])
+    for num in range(1,len(RULE_TRACE["trace"])):
+      queue_prompt = "\nQueues after processing of rule '{0}':\n".format(num)
```
lukpueh: I think it would be better if it showed the rule and not the rule index here, e.g.:

```
Queues after processing of rule ['DELETE', 'foobar']:
 'artifacts': ['foo', 'bar'],
 'materials': ['foo', 'bar'],
 'products': ['baz', 'foo', 'bar']
```

in_toto/verifylib.py
```diff
@@ -894,8 +896,8 @@ def verify_delete_rule(rule, source_materials_queue, source_products_queue):
 
   for matched_material in matched_materials:
     if matched_material in source_products_queue:
-      raise RuleVerificationError("Rule '{0}' failed, material '{1}' was found"
-          " in products but should have been deleted."
+      raise RuleVerificationError("Rule '{0}' failed, material '{1}' that should"
```
lukpueh: Line length. Also, is the message so much better now?

in_toto/verifylib.py
```diff
@@ -1048,8 +1050,16 @@ def verify_disallow_rule(rule, source_artifacts_queue):
...
+    rule_traceback = "Initial state of queues:\n"
+    rule_traceback += pprint.pformat(RULE_TRACE["trace"][0])
+    for num in range(1,len(RULE_TRACE["trace"])):
+      queue_prompt = "\nQueues after processing of rule '{0}':\n".format(num)
+      rule_traceback += queue_prompt + pprint.pformat(RULE_TRACE["trace"][num])
+    raise RuleVerificationError("Rule '{0}' failed.\nRule pattern matches the"
+        " following artifacts of the artifact queue, which is disallowed:\n"
+        " '{1}' \nHere is the trace for earlier '{2}' rules of item '{3}':\n"
+        " '{4}' ".format(" ".join(rule), matched_artifacts, RULE_TRACE["type"],
+        RULE_TRACE["name"], rule_traceback))
```
lukpueh: Would you mind adding a helper function that returns the formatted queue, instead of formatting it here?

in_toto/verifylib.py
```diff
@@ -1134,9 +1147,16 @@ def verify_item_rules(source_name, source_type, rules, links):
         " one of 'materials' or 'products.'\n"
         "Got:\n\t'{}'".format(source_type))
 
+  # Initialize rule traceback dictionary
+  RULE_TRACE["name"] = source_name
+  RULE_TRACE["type"] = source_type
+  RULE_TRACE["trace"] = {0: {'artifacts': source_artifacts_queue, 
+      'materials': source_materials_queue, 'products': source_products_queue}}
```
lukpueh: Why change the trace list to a trace dict, if the dict keys are just incrementing indices?

in_toto/verifylib.py
```diff
@@ -1134,9 +1147,16 @@ def verify_item_rules(source_name, source_type, rules, links):
...
 
   # Apply (verify) all rule
+  rule_count = 1
```
lukpueh: Check out Python's `enumerate` built-in, it comes in handy when you iterate over a list but need an index too. That said, you don't need the runner variable here (see my comment about dict vs. list above).

lukpueh (changes requested):
Cool stuff, this just made me discover a bug! I will follow up with a PR shortly. Please, address my comments and we can merge. Btw. this has gotten pretty verbose, @trishankatdatadog, is this what you had in mind with #228?

lukpueh:
Here's the sample output (using a slightly simplified version of above snippet):

```python
in_toto.exceptions.RuleVerificationError: Rule 'DISALLOW *' failed.
Rule pattern matches the following artifacts of the artifact queue, which is disallowed:
 '['bar']'
Here is the trace for earlier 'materials' rules of item 'item':
 'Initial state of queues:
{'artifacts': ['foo', 'foobar', 'bar'],
 'materials': ['foo', 'foobar', 'bar'],
 'products': ['baz', 'foo', 'bar']}
Queues after processing of rule '1':
{'artifacts': ['foo', 'bar'],
 'materials': ['foo', 'bar'],
 'products': ['baz', 'foo', 'bar'],
 'rule': ['DELETE', 'foobar']}
Queues after processing of rule '2':
{'artifacts': ['foo', 'bar'],
 'materials': ['foo', 'bar'],
 'products': ['foo', 'bar'],
 'rule': ['CREATE', 'baz']}
Queues after processing of rule '3':
{'artifacts': ['foo', 'bar'],
 'materials': ['foo', 'bar'],
 'products': ['foo'],
 'rule': ['MODIFY', 'bar']}
Queues after processing of rule '4':
{'artifacts': ['bar'],
 'materials': ['foo', 'bar'],
 'products': ['foo'],
 'rule': ['MATCH', 'foo', 'WITH', 'MATERIALS', 'FROM', 'item']}'
```

lukpueh:
Here's the promised PR (draft) https://github.com/in-toto/in-toto/pull/259, to fix the above mentioned issue. I'll phrase out the description and add a test next week.

trishankatdatadog:
I think it will go a looong way! Thanks, everyone :)

in_toto/verifylib.py
```diff
@@ -1048,8 +1050,16 @@ def verify_disallow_rule(rule, source_artifacts_queue):
...
-    raise RuleVerificationError("Rule '{0}' failed, pattern matched disallowed"
-        " artifacts: '{1}' ".format(" ".join(rule), matched_artifacts))
+    rule_traceback = "Initial state of queues:\n"
+    rule_traceback += pprint.pformat(RULE_TRACE["trace"][0])
+    for num in range(1,len(RULE_TRACE["trace"])):
+      curr_rnd = RULE_TRACE["trace"][num]
+      rule_name = curr_rnd["rule"]
+      queues = {k: curr_rnd[k] for k in ('artifacts', 'materials', 'products')}
+      q_prompt = "\nQueues after processing of rule '{0}':\n".format(rule_name)
+      rule_traceback += q_prompt + pprint.pformat(queues)
```
lukpueh: Oh, I must have been unclear. I was thinking of a helper function for L1053-L1060 so as to not pollute `verify_disallow_rule` with all that dict parsing and string constructing code.

The error message that you moved out now could IMHO still be left here, and suffixed by the return value from said helper function,  which might look something liked this:
```python
def _get_rule_traceback():
    trace_str = "Trace of {} rules for item '{}':\n".format(
        RULE_TRACE["type"], RULE_TRACE["name"])

    trace_str += "Initial state of queues:\n"
    trace_str += pprint.pformat(RULE_TRACE["trace"].pop(0)) + "\n"

    for trace_line in RULE_TRACE["trace"]:
      trace_str += "Queues after processing of rule '{0}':\n".format(
          trace_line.pop("rule"))
      trace_str += pprint.pformat(trace_line) + "\n"

  return trace_str
```

What do you think?

in_toto/verifylib.py
```diff
@@ -1306,6 +1332,53 @@ def verify_threshold_constraints(layout, chain_link_dict):
                     step_name=step.name, keyid=keyid)))
 
 
+def print_error_message(rule, matched_artifacts, rule_trace, error_traceback):
```
lukpueh: Please be mindful when you choose function names. Judging from the name this looks like a generic function to print error messages, but it is neither generic nor does it print something (see suggestion above for an alternative name).

lukpueh (changes requested):

in_toto/verifylib.py
michizhou: I agree with your suggestions. It was my mistake for misunderstanding your intentions. I will be sure to update the functionality of the helper function to reflect this.

in_toto/verifylib.py
```diff
@@ -1048,8 +1050,10 @@ def verify_disallow_rule(rule, source_artifacts_queue):
       source_artifacts_queue, rule_data["pattern"])
 
   if len(matched_artifacts):
-    raise RuleVerificationError("Rule '{0}' failed, pattern matched disallowed"
-        " artifacts: '{1}' ".format(" ".join(rule), matched_artifacts))
+    raise RuleVerificationError("Rule '{0}' failed.\nRule pattern matches the"
+        " following artifacts of the artifact queue, which is disallowed:\n"
+        " '{1}' \n'{2}'".format(" ".join(rule), matched_artifacts,
```
lukpueh: Can you remove the single quotes "`'`" in `'{1}'` and `'{2}'`, they look a bit weird in the output.

in_toto/verifylib.py
```diff
@@ -1119,6 +1123,9 @@ def verify_item_rules(source_name, source_type, rules, links):
   source_materials_queue = list(source_materials.keys())
   source_products_queue = list(source_products.keys())
 
+  # Reset rule traceback dictionary
+  RULE_TRACE.clear()
```
lukpueh, line 1126: Can you mention clearing and populating the global `RULE_TRACE` in the docstring's side effects section?

in_toto/verifylib.py
```diff
@@ -1306,6 +1326,42 @@ def verify_threshold_constraints(layout, chain_link_dict):
...
+  traceback_str = "Trace for earlier '{0}' rules of item '{1}':\n".format(
+      RULE_TRACE["type"], RULE_TRACE["name"])
+
+  traceback_str += "Initial state of queues:\n"
+  traceback_str += pprint.pformat(RULE_TRACE["trace"].pop(0)) + "\n"
+
+  for trace_entry in RULE_TRACE["trace"]:
+    traceback_str += "Queues after processing of rule '{0}':\n".format(
+        trace_entry["rule"])
+    queues = {k: trace_entry[k] for k in ('artifacts', 'materials', 'products')}
```
lukpueh: Our style guidelines [discourage the use of dict comprehensions](https://github.com/secure-systems-lab/code-style-guidelines#no-lambda-functions-or-lisp-esque-code). Personally, I think they are really neat, but let's not use them.

What about a second regular for loop, e.g....?

```python
for name in ["artifacts", "materials", "products"]:
  traceback_str += "\t{}: {}\n".format(name, trace_entry[name])
```
You could format the "initial state of queues" in that for loop too.  Then you wouldn't need `pprint` anymore.  As an alternative to `pop`ing the first line you can use an index from `enumerate` (remember?) so that you can distinguish between the first and the rest of the trace lines.

If you do use `pop`, please mention it in the docstring's side effects.

lukpueh (changes requested):

lukpueh (approved):

PR #244, opened 2018-11-19
FIX: gpg/rsa: fix #170 signature length bug

SantiagoTorres:
It appears the specification for PKCSv1.5 doesn't say how long should a
signature should be (i.e., should we zero-pad the signature or should we
use the minimum number of bits required). In this case, signatures
generated using gnupg would sometimes be 8*n bits shorter than the
public key length, which would result in signatures that fail to verify
due to missing leading zeroes (the leading zeroes existed at the bit
level however).

Zero-left-pad signature objects that do not match the length of the exponent
`n` on a gpg rsa public key object before verification.

that is provided, the better.

**Fixes issue #**:

requirements**:

- [ ] Tests have been added for the bug fix or new feature:
   I don't want to add a test for this on the interest of runtime of the test-suite, but I'm open to ideas on how to test this..
- [x] Docs have been added for the bug fix or new feature:
   I left a big comment warning future in-toto developers in their pilgrimage through the innards of RFC3447 and RFC4880...

### DIFF
--- a/in_toto/gpg/rsa.py
+++ b/in_toto/gpg/rsa.py
@@ -164,6 +164,18 @@ def gpg_verify_signature(signature_object, pubkey_info, content):
 
   pubkey_object = create_pubkey(pubkey_info)
 
+  # zero-pad the signature due to a discrepancy between the openssl backend
+  # and the gnupg interpretation of PKCSv1.5. Read more at:
+  # https://github.com/in-toto/in-toto/issues/171#issuecomment-440039256
+  # we are skipping this if on the tests because well, how would one test this
+  # deterministically.
+  pubkey_length = len(pubkey_info['keyval']['public']['n'])
+  signature_length = len(signature_object['signature'])
+  if pubkey_length != signature_length: # pragma: no cover
+    zero_pad = "0"*(pubkey_length - signature_length)
+    signature_object['signature'] = "{}{}".format(zero_pad,
+        signature_object['signature'])
+
   digest = in_toto.gpg.util.hash_object(
       binascii.unhexlify(signature_object['other_headers']),
       hashing.SHA256(), content)
--- a/in_toto/gpg/constants.py
+++ b/in_toto/gpg/constants.py
@@ -36,9 +36,6 @@ try:
   proc = process.run(GPG_VERSION_COMMAND, stdout=process.PIPE,
     stderr=process.PIPE)
 
-  # TODO: Remove debug statements after fixing in-toto/in-toto#171
-  log.debug("{0} (stdout):{1}".format(GPG_VERSION_COMMAND, proc.stdout))
-  log.debug("{0} (stderr):{1}".format(GPG_VERSION_COMMAND, proc.stderr))
 except OSError: # pragma: no cover
   GPG_COMMAND = "gpg"
   GPG_VERSION_COMMAND = GPG_COMMAND + " --version"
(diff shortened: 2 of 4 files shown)


SantiagoTorres:
I swear to god if tests fail due to some cryptic gpg verification bug **I'll kill and eat a turkey by the end of this week**.

SantiagoTorres:
related to #171

alyptik (approved):
Fix looks fine to me, ACK.

alyptik:
So I think I am starting to kind of see what's going on here. The signatures you are trying to verify with `gpg` actually aren't gpg-generated, just regular PKCS#15 ones from `openssl`? The PGP RFC does require those leading zeros, but I'm guessing PKCS#15 standard doesn't?

Or am I misunderstanding something (originally I had thought these were only gpg-generated keys) here still?

in_toto/gpg/rsa.py
```diff
@@ -164,6 +164,18 @@ def gpg_verify_signature(signature_object, pubkey_info, content):
 
   pubkey_object = create_pubkey(pubkey_info)
 
+  # zero-pad the signature due to a discrepancy between the openssl backend
+  # and the gnupg interpretation of PKCSv1.5. Read more at:
+  # https://github.com/in-toto/in-toto/issues/171#issuecomment-440039256
+  # we are skipping this if on the tests because w ell, how would one test this
+  # deterministically.
```
lukpueh, line 171: Can you fix spelling errors in the comment.
And would you mind describing how you figured out the padding discrepancy? PGP RFC? cryptography code base?

in_toto/gpg/rsa.py
```diff
@@ -164,6 +164,18 @@ def gpg_verify_signature(signature_object, pubkey_info, content):
...
+  pubkey_length = len(pubkey_info['keyval']['public']['n'])
+  signature_length = len(signature_object['signature'])
+  if pubkey_length != signature_length: # pragma: no-cover
```
lukpueh: /no-cover/no cover/

lukpueh (changes requested):
@SantiagoTorres, you're a hero! I am still not sure how you figured this out, but I'm impressed. :)

Should we remove the `log.debug` statements in `develop` now that we have debugged this? 
https://github.com/in-toto/in-toto/blob/22ebbef7a5b92c5ec527a1c693240c402d8441fe/in_toto/gpg/constants.py#L39-L41
https://github.com/in-toto/in-toto/blob/22ebbef7a5b92c5ec527a1c693240c402d8441fe/in_toto/gpg/functions.py#L94-L96
https://github.com/in-toto/in-toto/blob/22ebbef7a5b92c5ec527a1c693240c402d8441fe/in_toto/gpg/functions.py#L237-L239
https://github.com/in-toto/in-toto/blob/22ebbef7a5b92c5ec527a1c693240c402d8441fe/in_toto/gpg/util.py#L224-L226

alyptik:
@lukpueh said:
> @SantiagoTorres, you're a hero! I am still not sure how you figured this out, but I'm impressed. :)

Seconded, never would have thought to check this, ha. The whole subpacket 33 thing seems so silly now.

lukpueh:
@alyptik, for clarification: The keys and the signatures we are discussing here are indeed generated using gpg. We call the gpg2 command line tool for signing, using a key in the keychain, identified via keyid (or the default key, if no keyid is specified). And we also use the gpg2 command line tool to export public keys so that they can be e.g. added to an in-toto layout.

However, signature verification is performed without gpg. For that we use `pyca/cryptography`.

alyptik:
@lukpueh Aha, I see. Thanks for the clarification.

SantiagoTorres:
@lukpueh Let me know if this is what you had in mind. Cheers!

lukpueh (approved):

PR #234, opened 2018-10-10
Modified command line tool descriptions in README.md

michizhou:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #159**: Revisit command line tool descriptions in README.md

**Description of the changes being introduced by the pull request**:
Addressed some of @vladimir-v-diaz comments about issues with the README's command line tool descriptions in his gpg support PR review (#150). 

Current behavior/Expected behavior:

Changes that have been implemented (need review):
- Rephrased the general description of `in-toto-run`, shortening its purpose to a single sentence
- Specified that command-line options as what is passed to `in-toto-run` command line tool
- Added link to usage example from demo application for `in-toto-run` to help reduce emphasis on supported command-line options for `in-toto-run`
- Added description for `in-toto-sign` to README

Changes that need to be made (not implemented):
- Rethinking of command hierarchy, e.g. have `in-toto` as single base command with subcommands (`run`, `verify`, `record`, ...), instead of `in-toto-run`, `in-toto-verify`, `in-toto-record`, ...
- Possible addition of more simple usage examples to all commands to reduce emphasis on supported command-line options

**Please verify and check that the pull request fulfills the following
requirements**:

- [x] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [x] Docs have been added for the bug fix or new feature

Issue #159, opened 2017-12-08
Revisit command line tool descriptions in README.md


lukpueh:
Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.

**Description of issue or feature request**:
@vladimir-v-diaz mentioned several issues with the README's command line tool descriptions in his gpg support PR review (#150). We should address his comments in a separate PR.

**Current behavior**/**Expected behavior**:
From Vlad's comments:

- Rephrase the general description of `in-toto-run` especially its purpose https://github.com/in-toto/in-toto/pull/150#discussion_r155297699
- Explain what *"passing a command to `in-toto-run`"* means https://github.com/in-toto/in-toto/pull/150#discussion_r155337421
- Reduce emphasis on supported command-line options and add usage examples  instead https://github.com/in-toto/in-toto/pull/150#discussion_r155332638
- Rethink the command hierarchy, e.g. have `in-toto` as single base command with subcommands (`run`, `verify`, `record`, ...), instead of `in-toto-run`, `in-toto-verify`, `in-toto-record`, ... https://github.com/in-toto/in-toto/pull/150#discussion_r155337421
- Add `in-toto-sign` to README (https://github.com/in-toto/in-toto/pull/150#discussion_r155654194)

michizhou:
@lukpueh @alyptik If it's alright with you guys, I would be happy to take up this issue. I would really appreciate it if you guys could provide some supporting docs and reference materials so that I could get started with addressing the comments raised by @vladimir-v-diaz's review in #150. I will open a new PR to address some of the concerns in this issue for review.

alyptik:
@michizhou Great :) It's a tad sparse, but you can find examples and other supporting documentation on the website at [in-toto.github.io](https://in-toto.github.io/metadata-examples.html) or in the [in-toto/docs](https://github.com/in-toto/docs/tree/master/examples) repository. As I suggested in another comment, a possible rewording for #150 would be "executes the **specified** command..." which I feel is a bit clearer.

lukpueh:
[...]

PR #248, opened 2019-01-09
Update command descriptions in README

adityasaky:
**Fixes issue**: The base-path argument was missing from the documentation for in-toto-run and in-toto-record in the README file.

**Description of the changes being introduced by the pull request**: The argument has been added to the relevant sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,8 @@ To learn more about the different rule types, their guarantees and how they are 
 `in-toto-run` generates link metadata for the given command-line option and 
 runs it as its own command. See the [this simple usage example from the demo 
 application for more details](https://github.com/in-toto/demo#tampering-with-the-software-supply-chain). 
-For a detailed list of all the arguments, run `in-toto-run --help` or look at 
-the documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py).
+For a detailed list of all the command line arguments, run `in-toto-run --help` 
+or look at the code documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py).
 
 ##### in-toto-record
 `in-toto-record` works similar to `in-toto-run` but can be used for
@@ -79,9 +79,9 @@ by a single command. Use `in-toto-record start ...` to create a
 preliminary link file that only records the *materials*, then run the
 commands of that step or edit files manually and finally use
 `in-toto-record stop ...` to record the *products* and generate the actual
-link metadata file. For a detailed list of all arguments, run `in-toto-record start --help`
-or `in-toto-record stop --help`. You can also find the documentation 
-[here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_record.py).
+link metadata file. For a detailed list of all command line arguments and their usage,
+run `in-toto-record start --help` or `in-toto-record stop --help`. You can also 
+find the code documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_record.py).
 
 #### Release final product
 
@@ -97,8 +97,9 @@ Use `in-toto-verify` on the final product to verify that
 - materials and products of each step were in place as defined by the rules, and
 - run the defined inspections
 
-For a detailed list of arguments and usage, run `in-toto-verify --help` or look 
-at the documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_verify.py).
+For a detailed list of all command line arguments and their usage, run 
+`in-toto-verify --help` or look at the code documentation 
+[here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_verify.py).
 
 #### Signatures
 `in-toto-sign`, unlike `in-toto-verify`, can only add, replace, and 
@@ -114,8 +115,9 @@ This tool is intended to sign layouts created by the layout web wizard, but
 also serves well to re-sign test and demo data. For example, it can be used 
 if metadata formats or signing routines change.
 
-For a detailed list of arguments and usage, run `in-toto-sign --help` or look 
-at the documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_sign.py).
+For a detailed list of all command line arguments and their usage, run 
+`in-toto-sign --help` or look at the code documentation 
+[here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_sign.py).
 
 #### Settings
 Settings can be configured in [`in_toto.settings`](https://github.com/in-toto/in-toto/blob/develop/in_toto/settings.py), via prefixed environment variables or in RCfiles in one of the following
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ To learn more about the different rule types, their guarantees and how they are 
 ##### in-toto-run
 `in-toto-run` generates link metadata for the given command-line option and 
 runs it as its own command. See [this simple usage example from the demo application 
-for more details](https://github.com/in-toto/demo#tampering-with-the-software-supply-chain). 
+for more details](https://github.com/in-toto/demo). 
 For a detailed list of all the command line arguments, run `in-toto-run --help` 
 or look at the code documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py).
 
--- a/README.md
+++ b/README.md
@@ -66,9 +66,9 @@ To learn more about the different rule types, their guarantees and how they are 
 #### Carrying out software supply chain steps
 
 ##### in-toto-run
-`in-toto-run` is used to execute a step in the software supply chain. This can be
-anything relevant to the project such as tagging a release with `git`, running
-a test, or building a binary. The relevant step name and command are
+`in-toto-run` is used to execute a step in the software supply chain. This can
+be anything relevant to the project such as tagging a release with `git`,
+running a test, or building a binary. The relevant step name and command are
 passed as arguments, along with materials, which are files required for that
 step's command to execute, and products which are files expected as a result
 of the execution of that command. These, and other relevant details
--- a/README.md
+++ b/README.md
@@ -66,20 +66,7 @@ To learn more about the different rule types, their guarantees and how they are 
 #### Carrying out software supply chain steps
 
 ##### in-toto-run
-`in-toto-run` generates link metadata for the given command-line option and runs it as its own command. See the [this simple usage example from the demo application for more details](https://github.com/in-toto/demo#tampering-with-the-software-supply-chain).
-
-```shell
-in-toto-run  --step-name <unique step name>
-            {--key <functionary signing key path>,  --gpg [<functionary gpg signing key id>]}
-            [--gpg-home <path to gpg keyring>]
-            [--base-path <filepath>]
-            [--materials <filepath>[ <filepath> ...]]
-            [--products <filepath>[ <filepath> ...]]
-            [--record-streams]
-            [--no-command]
-            [--verbose] -- <cmd> [args]
-```
-
+`in-toto-run` generates link metadata for the given command-line option and runs it as its own command. See the [this simple usage example from the demo application for more details](https://github.com/in-toto/demo#tampering-with-the-software-supply-chain). For a detailed list of all the arguments, run `in-toto-run -help` or look at the documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py).
 
 ##### in-toto-record
 `in-toto-record` works similar to `in-toto-run` but can be used for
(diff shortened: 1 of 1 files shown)


lukpueh:
Thanks for the PR, @adityasaky! I just saw that ` base-path` is not the only missing argument.  And that the other commands in the README are out of sync with the implementation too.

What do you think about removing the detailed help message dumps from the README and just instruct the reader to use the respective command with the `--help` flag for more details?
I assume that the exact command line arguments are above all interesting for someone who is about to use the command and thus has it installed.

It would probably be more helpful to revise the prose descriptions of the commands in the README. Do you feel like doing that?

adityasaky:
> What do you think about removing the detailed help message dumps from the README and just instruct the reader to use the respective command with the --help flag for more details?

I think that's a good idea actually. The documentation is available via `--help` as you say and it's easier to maintain one set too. On the repo itself, we also have it in `runib.py` file under the respective function, so I think that's enough of a description too?

> It would probably be more helpful to revise the prose descriptions of the commands in the README. Do you feel like doing that?

I think I can after I work on the `lstrip` issue we discussed. Does that work?

lukpueh:
Yes, finishing #194 and then working on the README is a good plan. And yes, you can also reference to the source code documentation.

adityasaky:
I'm probably going to start work on updating README now. Is it okay if I stick to this branch and continue pushing here, so that it can be a part of this PR?

adityasaky:
I've updated the sections for `in-toto-run` and `in-toto-record`. I'll do the same thing for `verify` and `sign` once I confirm with the respective source files.

lukpueh (changes requested):
Awesome! Here are a few suggestions:
- Writing "*command line arguments and usage examples*" instead of *"arguments and usage"* might be clearer
- Please wrap lines at 80 characters. I know most of the README does not follow that rule, but we should slowly enforce it when changing parts. It makes future diffs smaller.
- It's generally good (accessibility) practice to have link texts describe what they are linking to, e.g. 
 *"...[code documentation]()..."* instead of *"..documentation [here]()..."*.
- Would you mind taking a stab at the description text of `in-toto-run`? #234 supposedly addressed some suggestions made in #159. I think this could be better. Besides there is a mistake (*"...See **the [this** simple..."*) and I am not sure why it links to the "tampering with the software supply chain" part of the demo instead of the entire demo.

adityasaky:
I’ll get on those changes, @lukpueh, thanks!

adityasaky:
I fixed the first three issues, @lukpueh. Regarding the last one:

- For now, I fixed the grammatical error ("the this")
- I also updated the link to point to the demo instead of that particular subsection

When you say it can be better, do you mean a rewrite of that section? I'm looking over the initial threads that started this - #159 and comments stemming from #150.

lukpueh:
Yes, please take another look at #159, especially the first three bullet points. `in-toto-run` is one of the two most central tools of the in-toto framework (the other one is `in-toto-verify`). The README should summarize its purpose and usage and point to additional documentation (docstrings, cli help message, demo).

And yes, @vladimir-v-diaz's original comments in #150 and by @aaaaalbert [comment in #234](https://github.com/in-toto/in-toto/pull/234#issuecomment-454793725) may give you some ideas.

README.md
```diff
@@ -115,24 +97,27 @@ Use `in-toto-verify` on the final product to verify that
...
-`in-toto-sign`, unlike `in-toto-verify`, can only add, replace, and verify signatures within in-toto Link or Layout metadata, with options to
-- replace (default) or add signature(s), with layout metadata able to be signed by multiple keys at once while link metadata can only be signed by one key at a time
-- write signed metadata to a specified path (if no output path is specified, layout metadata is written to the path of the input file while link metadata is written to '<name>.<keyid prefix>.link')
+`in-toto-sign`, unlike `in-toto-verify`, can only add, replace, and 
+verify signatures within in-toto Link or Layout metadata, with options to
+- replace (default) or add signature(s), with layout metadata able to be 
+signed by multiple keys at once while link metadata can only be signed by one key at a time
+- write signed metadata to a specified path (if no output path is specified, 
+layout metadata is written to the path of the input file while link metadata 
+is written to '<name>.<keyid prefix>.link')
```
lukpueh: Please use `&lt;` and `&gt;` for  &lt;  and &gt; or put the whole filename inside backticks.

README.md
```diff
@@ -115,24 +97,27 @@ Use `in-toto-verify` on the final product to verify that
...
-```
+For a detailed list of all command line arguments and their usage, run 
+`in-toto-verify --help` or look at the code documentation 
+[here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_verify.py).
 
 #### Signatures
-`in-toto-sign`, unlike `in-toto-verify`, can only add, replace, and verify signatures within in-toto Link or Layout metadata, with options to
-- replace (default) or add signature(s), with layout metadata able to be signed by multiple keys at once while link metadata can only be signed by one key at a time
-- write signed metadata to a specified path (if no output path is specified, layout metadata is written to the path of the input file while link metadata is written to '<name>.<keyid prefix>.link')
+`in-toto-sign`, unlike `in-toto-verify`, can only add, replace, and
```
lukpueh: `in-toto-sign` is a *only* a metadata helper tool, whereas `in-toto-verify` is the command line interface for the entire in-toto verification routine. 

Please rephrase to something like ...
*`in-toto-sign` is a metadata signature helper tool to add, replace ....*

adityasaky: Should the document outright say that `verify` is the command for the entire verification routine?

README.md
```diff
@@ -66,19 +66,18 @@ To learn more about the different rule types, their guarantees and how they are
...
-            [--record-streams]
-            [--no-command]
-            [--verbose] -- <cmd> [args]
-```
+`in-toto-run` is used to execute one of the software supply chain steps. The 
+relevant step name and command are passed as arguments, along with materials, 
+which are files required for command execution, and products which are files
+expected as a result of the execution of the command. These, and other 
+relevant details pertaining to the step are stored in a link file, which is 
+signed using the functionary's key.
```
lukpueh: I think it could still be made a little clearer that `in-toto-run` should be used to executes any command corresponding to a step of the software supply chain, such as tagging a software release with git, running tests or building a binaries.

Also the explanation about materials might be a little confusing. Which command requires them for execution? `in-toto-run` or the passed command? And will execution fail if they are not passed? (*It does not, but the materials are simply not recorded in the corresponding link file, if not passed*).

README.md
```diff
@@ -66,19 +66,18 @@ To learn more about the different rule types, their guarantees and how they are
...
 
+See [this simple usage example from the demo application 
+for more details](https://github.com/in-toto/demo). 
+For a detailed list of all the command line arguments, run `in-toto-run --help` 
+or look at the code documentation [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py).
```
lukpueh: Please consider the comment I left about link tests (here and below).

README.md
```diff
@@ -115,24 +104,27 @@ Use `in-toto-verify` on the final product to verify that
...
+is written to '&lt;name&gt;.&lt;keyid prefix&gt;.link')
 - verify signatures
-This tool is intended to sign layouts created by the layout web wizard, but also serves well to re-sign test and demo data. For example, it can be used if metadata formats or signing routines change.
-```shell
-usage: in-toto-sign [-h] -f <path> [-k <path> [<path> ...]]
-                    [-t <key_type> [<key_type> ...]] [-g [<id> [<id> ...]]]
-                    [--gpg-home <path>] [-o <path>] [-a] [--verify] [-v | -q]
-```
+
+This tool is intended to sign layouts created by the layout web wizard, but
```
lukpueh: Maybe we should link to the layout web wizard here:
https://in-toto.engineering.nyu.edu/

lukpueh (changes requested):
I like this a lot better. Can you give it another try and refine? Please also strip the trailing whitespace that you added.

README.md
lukpueh: No, I think that's not necessary.

README.md
adityasaky: Do you mean the web layout wizard specified below?

adityasaky: Oh do you mean this: 

> It's generally good (accessibility) practice to have link texts describe what they are linking to, e.g.
"...code documentation..." instead of "..documentation here...".

adityasaky: Fixed. I was confused by tests.

README.md
```diff
@@ -66,19 +66,20 @@ To learn more about the different rule types, their guarantees and how they are
...
-
+`in-toto-run` is used to execute a step in the software supply chain. This can be 
+anything relevant to the project such as tagging a release with `git`, running 
+a test, or building a binary to ship. The relevant step name and command are 
+passed as arguments, along with materials, which are files required for that 
+step's command to execute, and products which are files expected as a result 
+of the execution of that command. These, and other relevant details 
+pertaining to the step are stored in a link file, which is signed using the 
+functionary's key. If materials are not passed to the command, the link file 
+generated just doesn't record them.
```
lukpueh: Same goes for products.

README.md
```diff
@@ -66,19 +66,20 @@ To learn more about the different rule types, their guarantees and how they are
...
+
+See [this simple usage example from the demo application 
+for more details](https://github.com/in-toto/demo). 
+For a detailed list of all the command line arguments, run `in-toto-run --help` 
+or look at the [code documentation](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_run.py) here.
```
lukpueh: I think you can remove the *"here"* now (here and below).

lukpueh (changes requested):
So I have two more small nitpicks in my inline comments.

And the trailing whitespace I mentioned in my last comment, was about the single trailing space on every line that you modified (see https://github.com/secure-systems-lab/code-style-guidelines/issues/5 for a discussion about it). But this is just a suggestion, not a blocker.

I'm also requesting a quick review from someone else, preferably not from the in-toto team. (@aaaaalbert? @alyptik?)

README.md
```diff
@@ -66,19 +66,22 @@ To learn more about the different rule types, their guarantees and how they are
...
-            [--materials <filepath>[ <filepath> ...]]
-            [--products <filepath>[ <filepath> ...]]
-            [--record-streams]
-            [--no-command]
-            [--verbose] -- <cmd> [args]
-```
-
+`in-toto-run` is used to execute a step in the software supply chain. This can be
+anything relevant to the project such as tagging a release with `git`, running
+a test, or building a binary to ship. The relevant step name and command are
```
SantiagoTorres: I'd drop "to ship here"

README.md
```diff
@@ -115,24 +108,28 @@ Use `in-toto-verify` on the final product to verify that
...
-`in-toto-sign`, unlike `in-toto-verify`, can only add, replace, and verify signatures within in-toto Link or Layout metadata, with options to
-- replace (default) or add signature(s), with layout metadata able to be signed by multiple keys at once while link metadata can only be signed by one key at a time
-- write signed metadata to a specified path (if no output path is specified, layout metadata is written to the path of the input file while link metadata is written to '<name>.<keyid prefix>.link')
+`in-toto-sign` is a metadata signature helper tool to add, replace, and
+verify signatures within in-toto Link or Layout metadata, with options to:
+- replace (default) or add signature(s), with layout metadata able to be
+signed by multiple keys at once while link metadata can only be signed by one key at a time
+- write signed metadata to a specified path (if no output path is specified,
+layout metadata is written to the path of the input file while link metadata
+is written to '&lt;name&gt;.&lt;keyid prefix&gt;.link')
```
SantiagoTorres: can' we change these inline html encoded <,> markers to actual symbols?

SantiagoTorres (changes requested):
Two very minor nits.

Thanks!

README.md
```diff
@@ -66,19 +66,22 @@ To learn more about the different rule types, their guarantees and how they are
...
+anything relevant to the project such as tagging a release with `git`, running
+a test, or building a binary. The relevant step name and command are
+passed as arguments, along with materials, which are files required for that
+step's command to execute, and products which are files expected as a result
+of the execution of that command. These, and other relevant details
+pertaining to the step are stored in a link file, which is signed using the
+functionary's key. If materials are not passed to the command, the link file
+generated just doesn't record them. Similarly, if the execution of a command
+via `in-toto-run` doesn't result in any products, they're not recorded in the
+link file.
```
lukpueh: IMHO the parts about materials and products are still ambiguous. The command reads/modifies/creates files no matter what the functionary passed as `--materials` or `--products`. And the reverse is true, in-toto records the hashes of passed `--materials` and `--products` no matter which files the command touches. So what is passed here is not immediately relevant for the command to execute, but it allows the functionary to explicitly provide information about the command execution, which in turn allows us to verify eventually that the steps of the supply chain are properly linked together. Does this make sense? Could you try once more to distill this information? :) Thanks!

README.md
```diff
@@ -87,19 +90,9 @@ by a single command. Use `in-toto-record start ...` to create a
...
-                            [--materials <material path> [<material path> ...]]
-
-usage: in-toto-record stop -step-name <unique step name>
-                           (--key <signing key path> | --gpg [<gpg keyid>])
-                           [--gpg-home <gpg keyring path>] [-v]
-                           [--products <product path> [<product path> ...]]
-```
+link metadata file. For a detailed list of all command line arguments and their usage,
+run `in-toto-record start --help` or `in-toto-record stop --help`. You can also
+find the [code documentation](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_record.py).
```
lukpueh: This last sentence sounds a bit odd. Can you rephrase it to something like above or below.

lukpueh (changes requested):

README.md
adityasaky: I'll give it a shot.

SantiagoTorres (approved):

PR #250, opened 2019-01-15
Add capabilities to support left stripping from paths when recording artifacts

adityasaky:
**Fixes issue #**: #194 (partially)

The command line argument `--lstrip-paths` now accepts a single prefix which will then be left stripped from all matching paths of artifacts. This left stripped path is then used as the key to the dictionary when recording artifacts.

### DIFF
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -51,6 +51,7 @@ LSTRIP_PATHS_ARGS = ["--lstrip-paths"]
 LSTRIP_PATHS_KWARGS = {
   "dest": "lstrip_paths",
   "required": False,
+  "nargs": "+",
   "metavar": "<path>",
   "help": ("Record the path of artifacts in link metadata after left"
           " stripping the specified <path> from the full path. If"
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -51,7 +51,6 @@ LSTRIP_PATHS_ARGS = ["--lstrip-paths"]
 LSTRIP_PATHS_KWARGS = {
   "dest": "lstrip_paths",
   "required": False,
-  "nargs": 1,
   "metavar": "<path>",
   "help": ("Record the path of artifacts in link metadata after left"
           " stripping the specified <path> from the full path. If"
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -55,6 +55,8 @@ LSTRIP_PATHS_KWARGS = {
   "metavar": "<path>",
   "help": ("Record the path of artifacts in link metadata after left"
           " stripping the specified <path> from the full path. If"
-          " there are multiple prefixes specified, the longest match"
-          " is what is actually stripped from the full path.")
+          " there are multiple prefixes specified, only a single "
+           "prefix can match the path of any artifact and that is "
+           "then left stripped. All prefixes are checked to ensure none "
+           "of them are a left substring of another.")
 }
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -46,3 +46,15 @@ BASE_PATH_KWARGS = {
   "help": ("Record 'materials/products' relative to <path>. If not set,"
           " current working directory is used as base path.")
   }
+
+LSTRIP_PATHS_ARGS = ["--lstrip-paths"]
+LSTRIP_PATHS_KWARGS = {
+  "dest": "lstrip_paths",
+  "required": False,
+  "nargs": "+",
+  "metavar": "<path>",
+  "help": ("Record the path of artifacts in link metadata after left"
+          " stripping the specified <path> from the full path. If"
+          " there are multiple prefixes specified, the longest match"
+          " is what is actually stripped from the full path.")
+}
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -51,7 +51,7 @@ LSTRIP_PATHS_ARGS = ["--lstrip-paths"]
 LSTRIP_PATHS_KWARGS = {
   "dest": "lstrip_paths",
   "required": False,
-  "nargs": "+",
+  "nargs": 1,
   "metavar": "<path>",
   "help": ("Record the path of artifacts in link metadata after left"
           " stripping the specified <path> from the full path. If"
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -260,14 +260,14 @@ examples:
       in_toto.runlib.in_toto_record_start(args.step_name, args.materials,
           signing_key=key, gpg_keyid=gpg_keyid,
           gpg_use_default=gpg_use_default, gpg_home=args.gpg_home,
-          exclude_patterns=args.exclude_patterns, base_path=args.base_path)
+          exclude_patterns=args.exclude_patterns, base_path=args.base_path, lstrip_paths=args.lstrip_paths)
 
     # Mutually exclusiveness is guaranteed by argparser
     else: # args.command == "stop":
       in_toto.runlib.in_toto_record_stop(args.step_name, args.products,
           signing_key=key, gpg_keyid=gpg_keyid,
           gpg_use_default=gpg_use_default, gpg_home=args.gpg_home,
-          exclude_patterns=args.exclude_patterns, base_path=args.base_path)
+          exclude_patterns=args.exclude_patterns, base_path=args.base_path, lstrip_paths=args.lstrip_paths)
 
   except Exception as e:
     log.error("(in-toto-record {0}) {1}: {2}"
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -68,6 +68,9 @@ optional arguments:
                         for additional info.
   --base-path <path>    Record 'materials/products' relative to <path>. If not
                         set, current working directory is used as base path.
+  --lstrip-paths <path> If a prefix path is passed, the prefix is left stripped
+                        from the path of every artifact that contains the prefix.
+                        Currently, the prefix is a single path.
   -v, --verbose         Verbose execution.
   -q, --quiet           Suppress all output.
 
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -68,8 +68,13 @@ optional arguments:
                         for additional info.
   --base-path <path>    Record 'materials/products' relative to <path>. If not
                         set, current working directory is used as base path.
-  --lstrip-paths <path> If a prefix path is passed, the prefix is left stripped
-                        from the path of every artifact that contains the prefix.
+  --lstrip-paths <path> [<path> ...]
+                        Record the path of artifacts in link metadata after left
+                        stripping the specified <path> from the full path. If
+                        there are multiple prefixes specified, only a single
+                        prefix can match the path of any artifact and that is
+                        then left stripped. All prefixes are checked to ensure none
+                        of them are a left substring of another.
   -v, --verbose         Verbose execution.
   -q, --quiet           Suppress all output.
 
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -69,12 +69,12 @@ optional arguments:
   --base-path <path>    Record 'materials/products' relative to <path>. If not
                         set, current working directory is used as base path.
   --lstrip-paths <path> [<path> ...]
-                        Record the path of artifacts in link metadata after left
-                        stripping the specified <path> from the full path. If
-                        there are multiple prefixes specified, only a single
-                        prefix can match the path of any artifact and that is
-                        then left stripped. All prefixes are checked to ensure none
-                        of them are a left substring of another.
+                        Record the path of artifacts in link metadata after
+                        left stripping the specified <path> from the full
+                        path. If there are multiple prefixes specified, only a
+                        single prefix can match the path of any artifact and
+                        that is then left stripped. All prefixes are checked
+                        to ensure none of them are a left substring of another.
   -v, --verbose         Verbose execution.
   -q, --quiet           Suppress all output.
 
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -115,7 +115,8 @@ import in_toto.user_settings
 import in_toto.runlib
 
 from in_toto.common_args import (EXCLUDE_ARGS, EXCLUDE_KWARGS,
-    BASE_PATH_ARGS, BASE_PATH_KWARGS)
+    BASE_PATH_ARGS, BASE_PATH_KWARGS, LSTRIP_PATHS_ARGS,
+    LSTRIP_PATHS_KWARGS)
 
 # Command line interfaces should use in_toto base logger (c.f. in_toto.log)
 log = logging.getLogger("in_toto")
@@ -195,6 +196,7 @@ examples:
 
   parent_parser.add_argument(*EXCLUDE_ARGS, **EXCLUDE_KWARGS)
   parent_parser.add_argument(*BASE_PATH_ARGS, **BASE_PATH_KWARGS)
+  parent_parser.add_argument(*LSTRIP_PATHS_ARGS, **LSTRIP_PATHS_KWARGS)
 
 
   verbosity_args = parent_parser.add_mutually_exclusive_group(required=False)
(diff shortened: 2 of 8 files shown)


adityasaky:
Reviewing why the checks failed.

adityasaky:
The issue, for reference, is how `string.lstrip` works. It works off a character set rather than the string itself.

tests/test_runlib.py
```diff
@@ -188,6 +188,36 @@ def test_base_path_is_parent_dir(self):
 
     os.chdir(self.test_dir)
 
+
+  def test_lstrip_paths_valid_prefix_directory(self):
+    lstrip_paths = "subdir/subsubdir/"
```
ofek: Let's also try some explicit unicode strings

ofek (changes requested):

tests/test_runlib.py
adityasaky: In the prefix? The left strip would just fail silently in that instance as the test setup doesn't have such files. Should a file also be added then, which is explicitly unicode?

lukpueh: It's okay to change the test case setup or write your own test cases.

adityasaky: https://github.com/adityasaky/in-toto/commit/778aee7c7d40601967b98a0f00438f9f9b2fa049

I have this commit in my repo that adds such a test. The reason I haven't pushed to the branch here is because it requires me to explicitly declare the encoding. `tox` was complaining when I didn't, and I just wanted to check before I pushed it here.

lukpueh: Declaring the encoding in that test module is fine. Please also check why tests are failing on appveyor.

adityasaky: Whoops, didn't notice the Windows build failed, apologies.

Looking it up, I see that I need to add `encoding="utf-8"` in `open()` for Python 3 on Windows. However, `encoding` isn't valid in Python 2, so adding that will cause the Python 2 test to fail. Thoughts?

adityasaky: Also, worth noting that the underlying system may support unicode file paths and still have this issue, as it seems to be an issue with file handling, and printing to stdout (boiling down to stdout's support for unicode).

in_toto/in_toto_record.py
```diff
@@ -258,14 +263,14 @@ def main():
       in_toto.runlib.in_toto_record_start(args.step_name, args.materials,
           signing_key=key, gpg_keyid=gpg_keyid,
           gpg_use_default=gpg_use_default, gpg_home=args.gpg_home,
-          exclude_patterns=args.exclude_patterns, base_path=args.base_path)
+          exclude_patterns=args.exclude_patterns, base_path=args.base_path, lstrip_paths=args.lstrip_paths)
```
SantiagoTorres: This line isn't wrapped to 78 cols.

in_toto/in_toto_record.py
```diff
@@ -258,14 +263,14 @@ def main():
...
 
     # Mutually exclusiveness is guaranteed by argparser
     else: # args.command == "stop":
       in_toto.runlib.in_toto_record_stop(args.step_name, args.products,
           signing_key=key, gpg_keyid=gpg_keyid,
           gpg_use_default=gpg_use_default, gpg_home=args.gpg_home,
-          exclude_patterns=args.exclude_patterns, base_path=args.base_path)
+          exclude_patterns=args.exclude_patterns, base_path=args.base_path, lstrip_paths=args.lstrip_paths)
```
SantiagoTorres: Neither this

in_toto/runlib.py
```diff
@@ -149,6 +149,11 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
             endings before hashing the content of the passed files, for
             cross-platform support.
 
+    lstrip_paths: (optional)
+            If a prefix path is passed, the prefix is left stripped from
+            the path of every artifact that contains the prefix. Currently,
```
SantiagoTorres: I'd drop the `currently, the prefix is a single path.` as it just clutters the description (let's mention that more can be provided in the future instead).

in_toto/runlib.py
```diff
@@ -214,9 +219,21 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
-          normalize_line_endings=normalize_line_endings)
+      if lstrip_paths:
+        # If a prefix is passed using the argument --lstrip-paths,
+        # that prefix is left stripped from the filepath passed.
+        # Currently, it's set to only take a single prefix in common_args.
+        # Note: if the prefix doesn't include a trailing /, the dictionary key
+        # may include an unexpected /.
+        # Path was already normalized above
+        if artifact.startswith(lstrip_paths[0]):
+          stripped_artifact = artifact[len(lstrip_paths[0]):]
```
SantiagoTorres: Should we handle this as a single string for now? (i.e., make the arg a list and then turn it into a list once we settle for the multiple strip paths)

in_toto/runlib.py
```diff
@@ -214,9 +219,21 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
+        artifacts_dict[stripped_artifact] = _hash_artifact(artifact,
+            normalize_line_endings=normalize_line_endings)
+      else:
+        # Path was already normalized above
+        artifacts_dict[artifact] = _hash_artifact(artifact,
+            normalize_line_endings=normalize_line_endings)
```
SantiagoTorres: It feels that this else clause is repetitive, is there a reason we can't move 231-232 out of the if and ignore the else?

in_toto/runlib.py
```diff
@@ -264,8 +281,19 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
-              normalize_line_endings=normalize_line_endings)
+          if lstrip_paths:
+            # If a prefix is passed using the argument --lstrip-paths,
+            # that prefix is left stripped from the filepath passed.
+            # Currently, it's set to only take a single prefix in common_args.
+            # Note: if the prefix doesn't include a trailing /, the dictionary key
+            # may include an unexpected /.
+            if normalized_filepath.startswith(lstrip_paths[0]):
+              normalized_filepath = normalized_filepath[len(lstrip_paths[0]):]
+            artifacts_dict[normalized_filepath] = _hash_artifact(filepath,
```
SantiagoTorres: same goes here.

in_toto/runlib.py
```diff
@@ -476,7 +508,7 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
 
   materials_dict = record_artifacts_as_dict(material_list,
       exclude_patterns=exclude_patterns, base_path=base_path,
-      follow_symlink_dirs=True, normalize_line_endings=normalize_line_endings)
+      follow_symlink_dirs=True, normalize_line_endings=normalize_line_endings, lstrip_paths=lstrip_paths)
```
SantiagoTorres: Please wrap this around.

in_toto/runlib.py
```diff
@@ -490,7 +522,7 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
 
   products_dict = record_artifacts_as_dict(product_list,
       exclude_patterns=exclude_patterns, base_path=base_path,
-      follow_symlink_dirs=True, normalize_line_endings=normalize_line_endings)
+      follow_symlink_dirs=True, normalize_line_endings=normalize_line_endings, lstrip_paths=lstrip_paths)
```
SantiagoTorres: please wrap this

in_toto/runlib.py
```diff
@@ -615,7 +651,7 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
 
   materials_dict = record_artifacts_as_dict(material_list,
       exclude_patterns=exclude_patterns, base_path=base_path,
-      follow_symlink_dirs=True, normalize_line_endings=normalize_line_endings)
+      follow_symlink_dirs=True, normalize_line_endings=normalize_line_endings, lstrip_paths=lstrip_paths)
```
SantiagoTorres: Please wrap this

tests/test_runlib.py
```diff
@@ -125,6 +126,7 @@ def setUpClass(self):
         fp.write(path)
 
 
+
```
SantiagoTorres: is this stray newline needed?

tests/test_runlib.py
```diff
@@ -188,6 +190,79 @@ def test_base_path_is_parent_dir(self):
...
+      pass
+
+  @unittest.skipUnless(os.getenv("TEST_SKIP_WINDOWS_PY3"), "test only applies to windows and py3")
+  def test_lstrip_paths_valid_unicode_prefix_file_windows_py_3(self):
+    # Try to create a file with unicode character
+    try:
+      os.mkdir("ಠ")
+      path = "ಠ/foobar"
+      with open(path, "w", encoding="utf-8") as fp:
+        fp.write(path)
```
SantiagoTorres: I wonder if you'd just want to use `shutil.copy` to simplify all of this

SantiagoTorres (changes requested):

SantiagoTorres:
LGTM now, @ofek , @lukpueh could you do the honors? :100:

in_toto/common_args.py
```diff
@@ -46,3 +46,15 @@
...
+LSTRIP_PATHS_ARGS = ["--lstrip-paths"]
+LSTRIP_PATHS_KWARGS = {
+  "dest": "lstrip_paths",
+  "required": False,
+  "nargs": "+",
+  "metavar": "<path>",
+  "help": ("Record the path of artifacts in link metadata after left"
+          " stripping the specified <path> from the full path. If"
+          " there are multiple prefixes specified, the longest match"
+          " is what is actually stripped from the full path.")
```
lukpueh: Please update help message

in_toto/in_toto_record.py
```diff
@@ -68,6 +68,8 @@
                         for additional info.
   --base-path <path>    Record 'materials/products' relative to <path>. If not
                         set, current working directory is used as base path.
+  --lstrip-paths <path> If a prefix path is passed, the prefix is left stripped
+                        from the path of every artifact that contains the prefix.
```
lukpueh: This is not what the actual help message says (see comment above). They should match.

in_toto/runlib.py
```diff
@@ -204,6 +210,15 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
     norm_artifacts = _apply_exclude_patterns(norm_artifacts, exclude_patterns)
 
+  # Check if any of the prefixes passed for left stripping is a left substring
+  # of another
+  if lstrip_paths:
+    for prefix_one, prefix_two in itertools.combinations(lstrip_paths, 2):
+      if prefix_one.startswith(prefix_two) or \
+          prefix_two.startswith(prefix_one):
+        raise in_toto.exceptions.PrefixError("'{}' and '{}'"
+            "triggered a left substring error".format(prefix_one, prefix_two))
```
lukpueh, line 241: Missing space after string continuation: `'{}'triggered`

tests/test_runlib.py
```diff
@@ -188,6 +189,108 @@ def test_base_path_is_parent_dir(self):
...
+      path = "ಠ/foo"
+      shutil.copy("foo", path)
+
+      # Attempt to left strip the path now that the file has been created
+      lstrip_paths = ["ಠ/"]
+      expected_artifacts = sorted(["foo"])
+      artifacts_dict = record_artifacts_as_dict(["./ಠ/"],
+                                                lstrip_paths=lstrip_paths)
+      self.assertListEqual(sorted(list(artifacts_dict.keys())),
+                           expected_artifacts)
```
lukpueh: Please use double indent for line continuation.

in_toto/runlib.py
```diff
@@ -264,7 +293,23 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
+            for prefix in lstrip_paths:
+              if normalized_filepath.startswith(prefix):
+                key = normalized_filepath[len(prefix):]
+                break
+
+          if key in artifacts_dict:
+            raise in_toto.exceptions.PrefixError("Prefix selection has resulted"
+                " in non unique dictionary key '{}'".format(key))
+
+          artifacts_dict[key] = _hash_artifact(filepath,
```
lukpueh, line 303: Looks like there is potential for DRY here. Mind creating a helper function?

lukpueh (changes requested):
I have a few more requests...

in_toto/runlib.py
adityasaky: For the PrefixError check? It happens twice only (for now), but I suppose it could be broken off.

lukpueh: I meant the entire thing. [L232-L247](https://github.com/in-toto/in-toto/blob/744129161f8fef1569dc27db5b0b146b39d94670/in_toto/runlib.py#L232-L247) and [L299-L312](https://github.com/in-toto/in-toto/blob/744129161f8fef1569dc27db5b0b146b39d94670/in_toto/runlib.py#L299-L312) are exactly the same, right?

in_toto/in_toto_record.py
```diff
@@ -68,6 +68,13 @@
                         for additional info.
   --base-path <path>    Record 'materials/products' relative to <path>. If not
                         set, current working directory is used as base path.
+  --lstrip-paths <path> [<path> ...]
+                        Record the path of artifacts in link metadata after left
+                        stripping the specified <path> from the full path. If
+                        there are multiple prefixes specified, only a single
+                        prefix can match the path of any artifact and that is
+                        then left stripped. All prefixes are checked to ensure none
+                        of them are a left substring of another.
```
lukpueh: Please check [max line lengths](https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length).

in_toto/in_toto_run.py
```diff
@@ -60,6 +60,13 @@
                         for additional info.
   --base-path <path>    Record 'materials/products' relative to <path>. If not
                         set, current working directory is used as base path.
+  --lstrip-paths <path> [<path> ...]
+                        Record the path of artifacts in link metadata after left
+                        stripping the specified <path> from the full path. If
+                        there are multiple prefixes specified, only a single
+                        prefix can match the path of any artifact and that is
+                        then left stripped. All prefixes are checked to ensure none
+                        of them are a left substring of another.
```
lukpueh: Please check [max line lengths](https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length).

in_toto/runlib.py
```diff
@@ -214,8 +229,24 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
 
-      # Path was already normalized above
-      artifacts_dict[artifact] = _hash_artifact(artifact,
+      key = artifact
+      if lstrip_paths:
+        # If a prefix is passed using the argument --lstrip-paths,
+        # that prefix is left stripped from the filepath passed.
+        # Note: if the prefix doesn't include a trailing /, the dictionary key
+        # may include an unexpected /.
+        for prefix in lstrip_paths: # pragma: no cover
```
lukpueh: `no cover` excludes the entire loop from coverage. Please use [`no branch`](https://coverage.readthedocs.io/en/coverage-4.3.4/branch.html#branch) which marks it as partial.

tests/test_runlib.py
```diff
@@ -188,6 +189,108 @@ def test_base_path_is_parent_dir(self):
...
+      expected_artifacts = sorted(["foo"])
+      artifacts_dict = record_artifacts_as_dict(["./ಠ/"],
+          lstrip_paths=lstrip_paths)
+      self.assertListEqual(sorted(list(artifacts_dict.keys())),
+          expected_artifacts)
+      os.remove(path)
+      os.rmdir("ಠ")
+    except OSError:
+      # OS doesn't support unicode explicit files
+      pass
```
lukpueh: Why you distinguish between environments and then run the exact same test in both environments?

lukpueh (changes requested):
I did a full pass now. All the logic and corresponding tests look great. I have a few minor comments. Please address and we can merge. Thanks!

tests/test_runlib.py
adityasaky: Good point! 

Initially, the file was being written and not copied, and there was a difference there. When I switched to `shutil.copy`, the tests became the same. I’ll take one out and remove the environment check.

lukpueh (approved):

lukpueh:
Thanks, @adityasaky!

PR #251, opened 2019-01-21
Fix gpg subkey link threshold bug (#202)

### LINKED ISSUE #202: PGP subkeys count towards threshold
**Description of issue**:
After @vladimir-v-diaz 's review of [PR](https://github.com/in-toto/in-toto/pull/201#discussion_r174275398) #201, it was found that the threshold check with pgp keys can be bypassed by signing different pieces of link metadata with different subkeys.

**Current behavior**: A malicious functionary is able to bypass threshold verification by signing multiple pieces of link metadata with different subkeys on his private keyring. To do this, all the different subkeys need to be input by the project owner at the time of signing this root layout.

**Expected behavior**: Instead, any subkey and the master key signatures should count as a single signature towards the threshold.

lukpueh:
**Fixes issue #**:
Fixes #202 

This PR modifies the signature threshold verification of in-toto's final product verification to not count links signed by different GPG subkeys of the same GPG master key repeatedly towards a step's link threshold.

It also updates some threshold related comments.

requirements**:


in_toto/verifylib.py
```diff
@@ -399,32 +406,36 @@ def verify_link_signature_thresholds(layout, chain_link_dict):
 
   """
   # Create an inverse keys-subkeys dictionary, with subkey keyids as
-  # dictionary keys and main keys as dictionary values.
-  # NOTE: We assume that a given subkey can only belong to one master key
+  # dictionary keys and main keys as dictionary values. This will be
+  # required below to assess main-subkey trust delegations.
+  # We assume that a given subkey can only belong to one master key
```
SantiagoTorres, line 411: This is a good question. I believe it can be artificially done (actually, let me try it when I'm done with the paper). Spec-wise, I think the answer is yes, and I think we should assume so on our implementation.

I'd leave it as is and then add an assert later though.

tests/test_verifylib.py
```diff
@@ -1456,6 +1459,38 @@ def test_verify_link_signature_thresholds__S_S_S(self):
...
+        ],
+      keys={
+          masterkey: in_toto.gpg.functions.gpg_export_pubkey(
+              masterkey, self.gnupg_home)
+        }
+      )
+    with self.assertRaises(ThresholdVerificationError):
+      verify_link_signature_thresholds(layout, chain_link_dict)
+
+
```
SantiagoTorres, line 1493: it may be overkill, but can we add another one in which we sign with the masterkey and either subkey1 or subkey2?

SantiagoTorres (commented):
minor nits. Thanks for taking care of this.

in_toto/verifylib.py
lukpueh: okay

tests/test_verifylib.py
lukpueh: I briefly considered this too, but then I realized that we currently can't sign with a master key, if it has any attached subkeys with signing capabilities, of which gpg uses the most recently added, even if you pass the master key id.  :)

It's possible to forcefully use the masterkey, by appending a `!` to the keyid. However, our keyid schema (`[a-fA-F0-9]`) does not allow that.

SantiagoTorres: Ah, fair point. Let's leave this test as-is for now.

SantiagoTorres:
LGTM.

PR #252, opened 2019-01-22
Output and record standard streams in in-toto-run

### LINKED ISSUE #11: Properly duplicate standard streams

lukpueh:
**Fixes issue #**:
Fixes #11 

Add `subprocess.run`-like function to `in_toto.process` module that redirects standard output and standard error of the executed command to temporary files and both writes the contents to the corresponding streams of the parent process during execution and also returns them along
with the child process's return code upon its termination.

This function is called `in_toto_run` when passing `record_streams=True`.

The PR is a friendly takeover of #160. Thanks for the solid ground work, @ng1355! 

requirements**:

### DIFF
--- a/in_toto/process.py
+++ b/in_toto/process.py
@@ -178,11 +178,13 @@ def run_duplicate_streams(cmd, timeout=SUBPROCESS_TIMEOUT):
   stderr_fd, stderr_name = tempfile.mkstemp()
   try:
     with io.open(stdout_name, "r") as stdout_reader, \
-        io.open(stderr_name, "r") as stderr_reader:
+        os.fdopen(stdout_fd, "w") as stdout_writer, \
+        io.open(stderr_name, "r") as stderr_reader, \
+        os.fdopen(stderr_fd, "w") as stderr_writer:
 
       # Start child , writing standard streams to temporary files
-      proc = subprocess.Popen(cmd, stdout=stdout_fd,
-          stderr=stderr_fd, universal_newlines=True)
+      proc = subprocess.Popen(cmd, stdout=stdout_writer,
+          stderr=stderr_writer, universal_newlines=True)
       proc_start_time = time.time()
 
       stdout_str = stderr_str = ""
--- a/in_toto/process.py
+++ b/in_toto/process.py
@@ -19,11 +19,16 @@
   - require the Py3 subprocess backport `subprocess32` on Python2,
   - in-toto namespace subprocess constants (DEVNULL, PIPE) and
   - provide a custom `subprocess.run` wrapper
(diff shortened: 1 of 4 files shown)


in_toto/process.py
```diff
@@ -40,11 +45,11 @@
 PIPE = subprocess.PIPE
 
 
+
```
SantiagoTorres: Probably this stray newline can go away.

in_toto/process.py
```diff
@@ -118,3 +123,97 @@ def run(cmd, check=True, timeout=SUBPROCESS_TIMEOUT, **kwargs):
...
+          raise subprocess.TimeoutExpired(cmd, timeout)
+
+        # Read from child process's redirected streams, write to parent
+        # process's standard streams and construct retuirn values
+        stdout_part = stdout_reader.read()
+        stderr_part = stderr_reader.read()
+        sys.stdout.write(stdout_part)
+        sys.stderr.write(stderr_part)
+        sys.stdout.flush()
+        sys.stderr.flush()
```
SantiagoTorres, line 210: silly comment, but are all of these IPC calls available on windows too?

tests/test_process.py
```diff
@@ -50,5 +54,67 @@ def test_input_vs_stdin(self):
...
+
+    for cmd in [cmd_str, cmd_list]:
+      return_code, _, _ = in_toto.process.run_duplicate_streams(cmd)
+      self.assertEqual(return_code, 100)
+
+
+  def test_run_duplicate_streams_timeout(self):
+    """Test raise TimeoutExpired. """
+    with self.assertRaises(in_toto.process.subprocess.TimeoutExpired):
+      in_toto.process.run_duplicate_streams("python --version", timeout=-1)
```
SantiagoTorres, line 117: how is this timing out?

in_toto/process.py
lukpueh: According to the passing windows tests on appveyor they do.

tests/test_process.py
lukpueh: It times out a second before it started, and thus returns immediately. `subprocess`' timeout has the same behavior.

Issue #254, opened 2019-02-15
Fix summary link naming


lukpueh:
**Description of issue or feature request**:
As per the [specification](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#451-artifact-rules-in-sublayouts), *"when using sublayouts, the artifact rules that apply to the equivalent step [in the super-layout] are applied to the materials of the first [...] and the products of the last step of the sublayout"*.   To achieve this we [create summary links](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/verifylib.py#L1367) for sublayouts to be used for artifact rule verification in the super layout. 

To locate a link during artifact rule verification, we use its corresponding name. E.g. for some step of the super layout, whose artifacts should match the products of a sublayout, pertaining to a step `foo`, a rule could be:  `MATCH * WITH PRODUCTS FROM foo`.

However, currently we assign the [name of the sublayout's first step](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/verifylib.py#L1412) to its summary link, which besides posing a problem for match rules, is semantically inaccurate and may lead to namespace collisions (duplicate step names). 

Kudos to @aditya for pointing out that [the summary link naming seemed off](https://github.com/in-toto/in-toto-golang/blob/ba65c84a773271d05c16d1d0a718e98f1bbeeee2/in_toto/verifylib.go#L554-L562). 

**Current behavior**:
- Summary links for sublayouts are assigned the name of its (the sublayout's) first step
- Artifact rule verification for layouts with sublayouts is not tested

**Expected behavior**:
- Summary links for sublayouts should be assigned the name of the corresponding step in the super-layout
- Add tests

adityasaky:
I'm working on a fix for this on the golang repository and can probably apply it here too.

adityasaky:
in-toto/in-toto-golang#19 fixes this issue. I'll apply a similar fix here shortly.

adityasaky:
This can be closed now.

PR #256, opened 2019-02-25
Fix summary link naming

adityasaky:
that is provided, the better.

**Fixes issue #**: #254 

I've implemented the same fix used in the golang repository - see in-toto/in-toto-golang#19.

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1432,7 +1432,7 @@ def verify_sublayouts(layout, chain_link_dict, superlayout_link_dir_path):
   return chain_link_dict
 
 
-def get_summary_link(layout, reduced_chain_link_dict, step_name):
+def get_summary_link(layout, reduced_chain_link_dict, name):
   """
   <Purpose>
     Merges the materials of the first step (as mentioned in the layout)
@@ -1455,9 +1455,8 @@ def get_summary_link(layout, reduced_chain_link_dict, step_name):
               <link name> : <Metablock containing a Link object>,
               ...
             }
-    step_name:
-            The step that the layout corresponds to, only used in sublayout
-            verification.
+    name:
+            The name that the summary link will be associated with.
 
   <Exceptions>
     None.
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
What is a "summary link?" First time I ever heard about it.

adityasaky:
Summary links are used during sublayout verification  - a link summarizes the steps in the sublayout and takes its place. Right now, the name of a summary link is set as the name of the first link of the sublayout. When I was working on in-toto/in-toto-golang, @lukpueh and I realized this could be a potential problem, and this is a simple fix to use the corresponding step name instead. See: #254

in_toto/verifylib.py
```diff
@@ -1455,6 +1455,9 @@ def get_summary_link(layout, reduced_chain_link_dict):
               <link name> : <Metablock containing a Link object>,
               ...
             }
+    step_name:
+            The step that the layout corresponds to, only used in sublayout
+            verification.
```
lukpueh: Minor suggestions: I think the description and the name give more context about the *usual* way of calling `get_summary_link` than necessary. I'd just call the parameter `name` or `link_name` and  say that it is the name assigned to the link.

in_toto/verifylib.py
```diff
@@ -1568,6 +1571,9 @@ def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
               - artifact rules in step and inspection definitions in the layout
               - the run fields in the inspection definitions
               - the expected command in the step definitions
+    step_name: (optional)
+            The step that the layout corresponds to, used only for sublayout
+            verification to generate summary links.
```
lukpueh: Maybe it is worth pointing out that the parameter is used in recursive calls of `in_toto_verify` and to refer the reader to `verify_sublayouts` to see how it is used?

lukpueh (changes requested):
Thanks for the PR, @adityasaky! Please consider addressing my inline suggestions. Also, would you mind adding a test, where a MATCH rule of a super layout refers to the step, represented by the (now correctly named) summary link of a sublayout? (see #254)

SantiagoTorres (approved):
LGTM

lukpueh:
Thanks for addressing my concerns, @adityasaky. Can you please add the requested test?

adityasaky:
Yep, I'm working on that @lukpueh.

adityasaky:
I think this suffices, @lukpueh. Please review and let me know!

lukpueh (approved):

lukpueh:
Thanks for the awesome work, @adityasaky! In addition to reviewing your code I ran the test in debug mode and all looks good to me.

PR #255, opened 2019-02-19
Fix various minor gpg packet parsing issues

lukpueh:
**Fixes issue #**:
None

**Description of the changes being introduced by the pull request**:
-  Update gpg signature subpacket parsing to look for subpackets in both hashed and unhashed sections of a packet (ed1ccd5)
- Fix two subpacket header length parsing bugs (4841508)
- Support gpg new format packet length header  (29c781a)

Please see commit messages for details.

*__Noteworthy__*:
I discovered these bugs while working on gpg self-signature verification (related to #245), by running `in_toto.gpg.functions.gpg_export_pubkey(keyid)` over all public keys in my system gpg keyring.
```bash
# Use to get a list of all keyids
gpg --list-keys --with-colons 2>&1 | grep ^pub | grep -Eo --color=never [A-F0-9]\{16\,}
```

Although the gpg packets that triggered these bugs seem to be rare and are not relevant for our purposes (e.g. large user attribute packets), we still need to be able to parse them. Because if we fail to determine the end of any packet, we will fail to parse all subsequent potentially relevant packets in the stream of consecutive packets.

Due to the rareness of these gpg packets we currently deem the effort to manually craft them for testing purposes infeasible. 

**Please verify and check that the pull request fulfills the following
requirements**:
- [x] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [x] Docs have been added for the bug fix or new feature

PR #257, opened 2019-02-28
Add gpg self-signature verification support

lukpueh:
*I apologize for the size of this PR, but separating the commits into more (inter-dependent) PRs would only increase the PR handling effort and still leave at least one PR with a large diff. Let me know if there is anything I can do to facilitate code review. A guide to this PR is provided below.*

**Fixes issue #**:
- Depends on PR #255, please review and merge there first (ed1ccd5 - 050b7cb)
- Provides infrastructure required by #245

This PR adds signature verification on gpg public key export for two types of self-signatures:
- Self-certificates related to the primary key, which can be used to acquire additional certified information about the the primary key, such as the key expiration date.
- Subkey binding signatures, that bind each subkey to the primary key. So far we just assumed that a subkey attached to a primary key belongs to the primary key without verifying the binding signature. This PR discards all subkey that aren't verifiably bound to the primary key.

The lion share of this PR is provided by d3b6c44, which updates gpg public key parsing  to verify the corresponding self-signatures. 
Suggestions to re-organize the relevant functions, i.e. `get_pubkey_bundle `, `parse_pubkey_bundle`, `_assign_certified_key_info` and `_get_verified_subkeys`, are appreciated.

So far we only supported gpg binary type signatures (i.e. signatures over in-toto metadata) and restricted signature parsing accordingly. b7ce963 updates signature parsing to support the self-signature types described above. 

For binary type signature creation and verification we only support SHA256. 713929b, 9388560 add changes to support SHA1, and SHA512 too (for self-signatures only!), which, based on real-word keyring inspections, seem to be widely-used.  

4181469 updates signature parsing to return additional information, which is needed for further signature handling (as part of this PR), and also to get additional information about the primary key, such as the expiration date (required for #245). Note that this PR only provides certificate verification support, while the information extraction should be handled in #245 (see [inline `# TODO:` comment](https://github.com/lukpueh/in-toto/blob/d3b6c44fe76013568f3459f67ba2ed35db3935f0/in_toto/gpg/common.py#L342-L343)).

Further tangential clean-up/refactor commits include:
-  26c1fbf to remove an obsolete `else` clause
- ae4f5d0 to enhance some related error messages and fix comments and docstrings 
-  4d825be to slightly modify signature keyid handling

**Caveats/TODOs**
- Lacks proper testing (I'm inclined to do this in a separate PR out of sympathies for the reviewers)
- Does not handle contradictory self-certificates (see [inline `# TODO:` comment](https://github.com/lukpueh/in-toto/blob/d3b6c44fe76013568f3459f67ba2ed35db3935f0/in_toto/gpg/common.py#L343-L346), should be addressed with #245)
- Does not handle key revocation certificates (see https://github.com/in-toto/in-toto/issues/263)

requirements**:


in_toto/gpg/util.py
```diff
@@ -247,3 +310,37 @@ def is_version_fully_supported():
...
+    return hashing.SHA256
+
+  elif hash_algorithm_id == in_toto.gpg.constants.SHA512:
+    return hashing.SHA512
+
+  else:
+    raise ValueError("Hash algorithm '{}' not supported, must be one of '{}' "
+        "(see RFC4880 9.4. Hash Algorithms).".format(hash_algorithm_id,
+        {in_toto.gpg.constants.SHA1, in_toto.gpg.constants.SHA256,
+         in_toto.gpg.constants.SHA512}))
```
aaaaalbert: If this is never extended to support other hash types, my comment is moot, but there's a perhaps more concise way that avoids the `if` cascade:
```python
  supported_hashing_algorithms = [in_toto.gpg.constants.SHA1,
      in_toto.gpg.constants.SHA256, in_toto.gpg.constants.SHA512]

  corresponding_hashing_classes = [hashing.SHA1, hashing.SHA256, hashing.SHA512]

  hashing_class = dict(zip(supported_hashing_algorithms, corresponding_hashing_classes))

  try:
    return hashing_class[hash_algorihm_id]
  except KeyError:
    raise ValueError("Hash algorithm '{}' not supported, must be one of '{}' "
        "(see RFC4880 9.4. Hash Algorithms).".format(hash_algorithm_id,
        supported_hashing_algorithms))
```

(And had 938856026d091878883db9f837fe762dceaa4c03 not dropped `SUPPORTED_HASH_ALGORITHMS` from `gpg/constants.py`, the proposed code could be even DRYer.)

in_toto/gpg/common.py
```diff
@@ -127,62 +134,363 @@ def parse_pubkey_payload(data):
...
+      # - disallow duplicate packets
+      if packet_type != PACKET_TYPE_PRIMARY_KEY and \
+          not key_bundle[PACKET_TYPE_PRIMARY_KEY]["key"]:
+        raise Exception("First packet must be a primary key ('{}'), got '{}'."
+            .format(PACKET_TYPE_PRIMARY_KEY, packet_type))
+
+      elif packet_type == PACKET_TYPE_PRIMARY_KEY and \
+          key_bundle[PACKET_TYPE_PRIMARY_KEY]["key"]:
+        raise Exception("Unexpected primary key.")
+
```
aaaaalbert, line 205: I always find `if` cascades with complex causes difficult to read, as I'm never sure if I'm not missing a combination of conditions that's not handled. (I do think however that L193-L251 make sense.)

The `elif` in L198-199 could be rolled into L205+ as a sub-`if` clause.

in_toto/gpg/common.py
```diff
@@ -127,62 +134,363 @@ def parse_pubkey_payload(data):
...
+      # The first (and only the first) packet in the bundle must be the master
+      # key.  See RFC4880 12.1 Key Structures, V4 version keys
+      # TODO: Do we need additional key structure assertions? e.g.
+      # - there must be least one User ID packet, or
+      # - order and type of signatures, or
+      # - disallow duplicate packets
+      if packet_type != PACKET_TYPE_PRIMARY_KEY and \
+          not key_bundle[PACKET_TYPE_PRIMARY_KEY]["key"]:
+        raise Exception("First packet must be a primary key ('{}'), got '{}'."
```
aaaaalbert: Please raise more specific error types than plain `Exception`s. (See also L200)

in_toto/gpg/common.py
```diff
@@ -127,62 +134,363 @@ def parse_pubkey_payload(data):
...
+          key_bundle[PACKET_TYPE_PRIMARY_KEY]["signatures"].append(packet)
+
+      else:
+        log.info("Ignoring gpg key packet '{}', we only handle packets of "
+            "types '{}' (see RFC4880 4.3. Packet Tags).".format(packet_type,
+            [PACKET_TYPE_PRIMARY_KEY, PACKET_TYPE_USER_ID,
+            PACKET_TYPE_USER_ATTR, PACKET_TYPE_SUB_KEY,
+            PACKET_TYPE_SIGNATURE]))
+
+    except Exception as e:
```
aaaaalbert: There's a bunch of `except Exception`s across this PR. I wonder why the blanket `except` is desired (instead of a more specific exception type), and why it makes sense in a few places to `continue` after logging the exception.

in_toto/gpg/util.py
```diff
@@ -100,34 +101,96 @@ def parse_packet_header(data, expected_type=None):
...
   """
   data = bytearray(data)
-  packet_type = (data[0] & 0x3c ) >> 2
-  packet_length_bytes = data[0] & 0x03
+  header_len = None
+  body_len = None
+
+  # If Bit 6 of 1st octet is set we parse a New Format Packet Length, and
+  # an Old Format Packet Lengths otherwise
+  if data[0] & 0x40: # pragma: no cover
```
aaaaalbert: Consider using binary representations for this and L132, `0b0100_0000` and `0b0011_1111` and such. (I think this is what they're here for :-)

in_toto/gpg/common.py
lukpueh: This one (and L200) are caught and re-raised with more information as specific `PacketParsingError`s in L254, adding additional information about the packet position.

aaaaalbert: OK, got it. I wonder if it makes sense to raise the `PacketParsingError` on that level already, just so as to not catch other potential exceptions and report them in a confusing way.

(Then L254 would only `except PacketParsingError` and act upon it, no other unrelated exceptions.)

in_toto/gpg/common.py
lukpueh: Agreed! I do have this concern somewhere in my notes and I meant to add it to the PR description's "caveats/todos" section, as this issue applies to other places in the gpg module too, and I plan on addressing it in a separate PR. Thanks for pointing it out!

in_toto/gpg/common.py
lukpueh: Makes sense!

in_toto/gpg/common.py
lukpueh: Re `continue`: I considered that question too and decided that in some cases it is okay to continue the overall key export but ignore a self-certificate, if it
  - belongs to a malformed User ID packet (L307),
  - cannot be parsed  (L315),
  - is invalid (L337),

and, ignore a subkey, if it
  - cannot be parsed (L386, L389),
  - has a corresponding binding signature that cannot be parsed (L400),
  - has not exactly one binding signature (L420), or
  - the binding signature is invalid (L430).

But maybe it makes sense too do abort, since these errors are a sign that something is not fully supported or fishy.

in_toto/gpg/common.py
lukpueh: Agreed, that's why I added comprehensive comments.

> The `elif` in L198-199 could be rolled into L205+ as a sub-`if` clause.

That's how I wrote it initially, but then I thought it would be more readable the way it is now, as it is less arrow-head'ish.

aaaaalbert: > arrow-head'ish

Hmm, like, *two* levels in doesn't feel *that* spiky to me :-)

in_toto/gpg/util.py
lukpueh: Great idea, will do.

in_toto/gpg/common.py
lukpueh: It would be 4 levels already: `while`, `try`, `elif`, `if` ... 5 if you count the `def`... :P

aaaaalbert: ...which is *technically* correct, but I think the `def` and `while` and `try` only count as overarching brackets. So: two ;-)

in_toto/gpg/util.py
lukpueh: I opted for the if/else variant because it seemed a tad easier to read, and I am not sure if we need to support other algos in the near future. But your approach is definitely more elegant.

lukpueh:
Thanks for your comments @aaaaalbert, they are very much appreciated! ❤️❤️❤️

Do you have any advice regarding the organization of  `get_pubkey_bundle `, `parse_pubkey_bundle`, `_assign_certified_key_info` and `_get_verified_subkeys`? I had this as one long spaghetti function initially, but that was too much to bear, so I split them into separate functions, but the boundaries feel a little arbitrary. Any thoughts?

aaaaalbert:
My pleasure! 😄 

------

Re organizing these functions, I think it's very sensible to have them separated even if they are called from a single place only. Each is moderately complicated, each has side- and sub-conditions in its checks, there must be tons of docs in each.... I'm totally fine with the way you did this!

The only thing that comes to mind is the choice of *where* exactly you collate the return values of the  private / sub-functions into what the caller uses. For example, [`parse_...` combines a few values](https://github.com/in-toto/in-toto/blob/d3b6c44fe76013568f3459f67ba2ed35db3935f0/in_toto/gpg/common.py#L262-L265); meanwhile [`get_...` could do the same right here](https://github.com/in-toto/in-toto/blob/d3b6c44fe76013568f3459f67ba2ed35db3935f0/in_toto/gpg/common.py#L493) with similar effort, assuming that it got `parse_...`'s `key_bundle` instead of the tuple it actually returns.

This isn't super-important, it's just to consider the time spent for someone who reads the code and has to trace function calls.

lukpueh:
@aaaaalbert, funny that you would mention this. Changing the code from the way you are proposing to what it is now was one of the last things I did on my WIP branch. 

The (probably only semi-compelling) reason why I'm calling the helper functions from within `parse_...` is that I didn't want to *expose* the raw key bundle (i.e. primary key, user id, user attribute, and sub key packets plus their respective signature packets) in order to not have it confused it with the key bundle (i.e. parsed primary key with certified additional info and parsed and bound subkeys) that's returned from `get_...` eventually.

But `parse_...` could of course be seen as just another helper function that returns a *private* data structure to be used by other helper functions.

aaaaalbert:
That's interesting! Any way is fine though, I suggest no change.

in_toto/gpg/functions.py
```diff
@@ -154,6 +157,11 @@ def gpg_verify_signature(signature_object, pubkey_info, content):
 
     content:
             The content to be verified. (bytes)
+    hash_algorithm_id: (optional)
+            one of SHA1, SHA256, SHA512 (see in_toto.gpg.constants) used to
+            verify the signature. Default is SHA256.
+            NOTE: The hash algorithms specified in "pubkey_info"'s
+            "hashes" or "method" field are currently not considered.
```
SantiagoTorres: I'm curious as to why we want to do this instead of updating our signature-type class?

in_toto/gpg/util.py
```diff
@@ -187,12 +250,12 @@ def parse_subpackets(subpacket_octets):
     length = subpacket_octets[ptr]
     ptr += 1
 
-    if length > 192 and length < 255 : # pragma: no cover
+    if length >= 192 and length < 255 : # pragma: no cover
```
SantiagoTorres: Nice catch, thanks for correcting this.

in_toto/gpg/util.py
SantiagoTorres: I was going to suggest something similar. I'm fine either way though.

in_toto/gpg/constants.py
```diff
@@ -44,16 +44,18 @@
...
-
-# See sections 5.2.3 (signature) and 5.2.2 (public key) of RFC4880
+# See RFC4880 section 4.3. Packet Tags for a list of all packet types The
+# relevant packets defined below are described in sections 5.2 (signature),
+# 5.5.1.1 (primary pubkey) and 5.5.1.2 (pub subkey), 5.12 (user id) and 5.13
+# (user attribute)
+PACKET_TYPE_SIGNATURE = 0x02
+PACKET_TYPE_PRIMARY_KEY = 0x06
+PACKET_TYPE_USER_ID = 0x0D
+PACKET_TYPE_USER_ATTR = 0x11
```
SantiagoTorres, line 54: I'm assuming we need these two to concatenate them and verify the key-binding sig? If so, then ignore this. I just wanted to bring this up.

in_toto/gpg/util.py
aaaaalbert: Minimal [style issue](https://github.com/secure-systems-lab/code-style-guidelines#pet-peeves): No leading space for the `:` here.

lukpueh: @SantiagoTorres, it only got to my attention because I have a key in my system keyring, where that byte is exactly 192. :)
@aaaaalbert, thanks!

in_toto/gpg/constants.py
lukpueh: You mean User ID and User Attribute? Yes, well no, not key-binding sigs but rather self-certs (0x10 - 0x13).  See [`_assign_certified_key_info()`](https://github.com/in-toto/in-toto/blob/5f18f390b0b706b0a6899f3ca6f050277570f355/in_toto/gpg/common.py#L268-L282) for more details.

lukpueh:
Thanks for your feedback, @SantiagoTorres and @aaaaalbert! I'll address your comments and add tests soon.

in_toto/gpg/functions.py
lukpueh: What do you mean by "signature-type class"? The information that's on the pubkey, i.e. [`hashes`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/gpg/formats.py#L33) and [`method`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/gpg/formats.py#L50)? If yes, than the reason is that that's the hashes/method we support for binary signatures, i.e. signatures over documents, which is the main purpose of the keys, and which still is only sha256.

I added sha1 and sha512 for self-signatures only, which,  IIRC, is something we agreed on in a different conversation. Does this make sense, or did I misunderstand your question?

in_toto/gpg/common.py
lukpueh: Will be addressed in https://github.com/in-toto/in-toto/issues/126#issuecomment-473337539.

in_toto/gpg/common.py
lukpueh: Agreed, it's only two (...technically). I'll still keep this one.

lukpueh:
36085ad to 9a6141f are related to testing, including some minor modifications in the actual implementation  to make testing easier. Please review!

in_toto/gpg/rsa.py
```diff
@@ -26,7 +25,7 @@
 import in_toto.gpg.util
 import in_toto.gpg.exceptions
 import in_toto.gpg.formats
-
+import in_toto.gpg.constants
```
SantiagoTorres: I see that this removed one whitespace (we should have 2 here right?)

in_toto/gpg/util.py
```diff
@@ -26,6 +26,7 @@
 
 import in_toto.gpg.exceptions
 import in_toto.process
+import in_toto.gpg.constants
```
SantiagoTorres, line 30: same whitespace issue.

in_toto/gpg/functions.py
```diff
@@ -129,8 +131,8 @@ def gpg_sign_object(content, keyid=None, homedir=None):
     raise ValueError("Full keyid could not be determined for signature '{}'".
         format(signature))
 
-  # If we have a full keyid it is safe to remove the short keyid to save space
-  del signature["short_keyid"]
+  # It is okay now to remove the optional short keyid to save space
+  signature.pop("short_keyid", None)
```
SantiagoTorres, line 135: Curious: what's the rationale for this change?

in_toto/gpg/common.py
```diff
@@ -127,62 +134,363 @@ def parse_pubkey_payload(data):
...
+        }
+
+      # Other non-signature packets in the key bundle include User IDs and User
+      # Attributes, required to verify primary key certificates, and subkey
+      # packets. For each packet we create a new ordered dictionary entry. We
+      # use a dictionary to aggregate signatures by packet below,
+      # and it must be ordered because each signature packet belongs to the
+      # most recently parsed packet of a type.
+      elif packet_type in [PACKET_TYPE_USER_ID, PACKET_TYPE_USER_ATTR,
+          PACKET_TYPE_SUB_KEY]:
```
SantiagoTorres: nit, can we use a set instead?

in_toto/gpg/common.py
```diff
@@ -127,62 +134,363 @@ def parse_pubkey_payload(data):
...
+        key_bundle[packet_type][packet] = {
+          "body_len": body_len,
+          "signatures": []
+        }
+
+      # The remaining relevant packets are signatures, required to bind subkeys
+      # to the primary key, or to gather additional information about the
+      # primary key, e.g. expiration date.
+      # A signatures corresponds to the most recently parsed packet of a type,
```
SantiagoTorres: s/signatures/signature/g

in_toto/gpg/common.py
```diff
@@ -127,62 +134,363 @@ def parse_pubkey_payload(data):
...
+  # (see Notes about certification signatures in RFC 4880 5.2.3.3.)
+  for user_id_packet, packet_data in bundle[PACKET_TYPE_USER_ID].items():
+    try:
+      # Construct signed content (see RFC4880 section 5.2.4. paragraph 4)
+      signed_content = (bundle[PACKET_TYPE_PRIMARY_KEY]["packet"] +
+          b"\xb4\x00\x00\x00" + user_id_packet[1:])
+
+    except Exception as e:
+      log.info(e)
+      continue
```
SantiagoTorres: I wonder if we want to remove this `except` block and just let it blow up instead.

in_toto/gpg/common.py
```diff
@@ -127,62 +134,363 @@ def parse_pubkey_payload(data):
...
+            supported_hash_algorithms={SHA1, SHA256, SHA512},
+            supported_signature_types=SIGNATURE_TYPE_CERTIFICATES,
+            include_info=True)
+        # gpg_verify_signature requires a "keyid" even if it is short.
+        # (see parse_signature_packet for more information about keyids)
+        signature["keyid"] = signature["keyid"] or signature["short_keyid"]
+
+      except Exception as e:
+        log.info(e)
+        continue
```
SantiagoTorres, line 320: same here

in_toto/gpg/dsa.py
```diff
@@ -183,6 +183,8 @@ def gpg_verify_signature(signature_object, pubkey_info, content,
     hash_algorithm_id:
             one of SHA1, SHA256, SHA512 (see in_toto.gpg.constants) used to
             verify the signature
+            NOTE: Overrides any hash algorithm specification in "pubkey_info"'s
+            "hashes" or "method" fields.
```
SantiagoTorres, line 187: I wonder if this note sounds more dramatic than it really is (as far as I understand `hash_algorithm_id` is populated with `pubkey_info["hashes"]` or `"method"` fields when not used directly?

in_toto/gpg/common.py
```diff
@@ -329,14 +672,18 @@ def parse_signature_packet(data):
     if subpacket_type == PARTIAL_KEYID_SUBPACKET:
       short_keyid = binascii.hexlify(subpacket_data).decode("ascii")
 
+    # info["subpackets"].append((
+    #   subpacket_type,
+    #   binascii.hexlify(subpacket_data).decode("ascii")))
```
SantiagoTorres, line 679: This somehow sneaked through. Could you please remove it?

SantiagoTorres (approved):
Whoa, this took me *hours* to review.

On the surface I could only find very minor stylistic stuff, nice job!

I'm approving this, and probably doing the style changes myself. I want to test backwards-compatibilty with old metadata, as I'm sure our integrators would appreciate.

in_toto/gpg/rsa.py
lukpueh: Are you talking about blank lines after imports?

in_toto/gpg/functions.py
lukpueh: `del signature["short_keyid"]` might raise a KeyError. Since `short_keyid` is optional, I'm using pop to be safe.

in_toto/gpg/common.py
lukpueh: sure will change

in_toto/gpg/rsa.py
SantiagoTorres: yes

in_toto/gpg/common.py
lukpueh: I left it for michael, but I can remove it.

in_toto/gpg/functions.py
SantiagoTorres: Ah right. Thanks for clarifying!

in_toto/gpg/common.py
lukpueh: I am doing this in a more recent commit.

in_toto/gpg/common.py
lukpueh: Here we shouldn't because `parse_signature_packet` might fail legitimately, because we want to continue if parse_signature_packet fails because we're giving it e.g. a revocation signature, which might happen. In a more recent commit I added a `TODO: Revise exception taxonomy`

in_toto/gpg/dsa.py
lukpueh: No. Currently `SHA256` is passed explicitly in the interface function `gpg.functions.gpg_verify_signature` to the corresponding dsa or rsa verification function. While it aligns with the possible values for `"hashes"` or `"method"`, assigned in `gpg.common.parse_pubkey_payload`, it is not populated from those fields.

But even if it were populated from those fields (IMO a good idea), we still want to be able to use a public key that we restrict to sha256 for binary signatures (i.e. `key_info["hashes"] = "pgp+SHA2"`), to verify self-signatures with other hash algos ... hence the support for an override and hence the comment. :)

We can of course re-think the fields in the public key dict and how we assign them. But I wouldn't do it as part of this PR.

in_toto/gpg/common.py
lukpueh: I added a TODO comment 60555c7 and will remove or un-comment it with #245

in_toto/gpg/common.py
lukpueh: Done 77d7fb1

in_toto/gpg/common.py
lukpueh: Done a4a7e58

in_toto/gpg/rsa.py
lukpueh: See bac9726

in_toto/gpg/util.py
lukpueh: See bac9726

SantiagoTorres:
LGTM.

Thanks @lukpueh !

PR #258, opened 2019-02-28
Customize CWD for inspections

trishankatdatadog:
**Fixes issue #**:

N/A.

Customize the CWD where inspections are run (before this, "." was assumed, and the integrator could not override this at runtime).

This PR updates:
1. Where commands for inspections can be run.
2. Where products and materials for inspections are recorded.

requirements**:

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -314,7 +314,7 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
 
   return artifacts_dict
 
-def execute_link(link_cmd_args, record_streams):
+def execute_link(link_cmd_args, record_streams, link_cmd_cwd):
   """
   <Purpose>
     Executes the passed command plus arguments in a subprocess and returns
@@ -330,6 +330,8 @@ def execute_link(link_cmd_args, record_streams):
             A bool that specifies whether to redirect standard output and
             and standard error to a temporary file which is returned to the
             caller (True) or not (False).
+    link_cmd_cwd:
+            A path to the directory where the link command must be executed.
 
   <Exceptions>
     TBA (see https://github.com/in-toto/in-toto/issues/6)
@@ -346,13 +348,14 @@ def execute_link(link_cmd_args, record_streams):
   """
   if record_streams:
     process = in_toto.process.run(link_cmd_args, check=False,
-      stdout=in_toto.process.PIPE, stderr=in_toto.process.PIPE,
-      universal_newlines=True)
+      cwd=link_cmd_cwd, stdout=in_toto.process.PIPE,
+      stderr=in_toto.process.PIPE, universal_newlines=True)
     stdout_str, stderr_str = process.stdout, process.stderr
 
   else:
     process = in_toto.process.run(link_cmd_args, check=False,
-      stdout=in_toto.process.DEVNULL, stderr=in_toto.process.DEVNULL)
+      cwd=link_cmd_cwd, stdout=in_toto.process.DEVNULL,
+      stderr=in_toto.process.DEVNULL)
     stdout_str = stderr_str = ""
 
   return {
@@ -413,7 +416,7 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     record_streams=False, signing_key=None, gpg_keyid=None,
     gpg_use_default=False, gpg_home=None, exclude_patterns=None,
     base_path=None, compact_json=False, record_environment=False,
-    normalize_line_endings=False, lstrip_paths=None):
+    normalize_line_endings=False, lstrip_paths=None, link_cmd_cwd=None):
   """
   <Purpose>
     Calls functions in this module to run the command passed as link_cmd_args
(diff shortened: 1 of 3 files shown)


JustinCappos:
Can / will this be reflected in the layout?  It seems like the cwd should be explicitly set in the layout and then captured in the generated link metadata...

trishankatdatadog:
@JustinCappos A few arguments against specifying in root layout:

1. It still wouldn't work anyway, because some parts of the code (recording materials for) hardcoded "." as the CWD.
2. Too onerous. Have to use param substitutions everywhere to specify the "CWD."
3. Too inflexible. If users can specify where link metadata live, why not specify where to perform inspections?

JustinCappos:
Okay, these arguments make sense.  

At some level, there is value in starting to capture broader information about the system where we are running inspections and steps.  We've settled into a pragmatic scenario where we largely just assume this system is going to do the right thing.  Over time we should perhaps consider what added value one gets from verifying more parts of the environment along with the rest of the process.  Perhaps @SantiagoTorres , you, and I can discuss at some point later this month...

trishankatdatadog:
@JustinCappos @SantiagoTorres Sure, happy to, just send me an invite, please, thanks!

trishankatdatadog:
I forgot another argument, just for future reference:
4. Sometimes, you don't have permission to write (e.g., inspection links) to where you are (another user's home dir), and you don't know where to switch to in advance.

lukpueh:
*Re `inspection_dir_path`*:
Adding a way to customize the inspection artifact recording directory definitely makes sense, we provide this option (actually two, one for materials and one products) to functionaries when they generate step links, so we should also provide this to verifiers for inspection links. 

Keep in mind though that `inspection_dir_path` applies to all inspections of the layout and to materials and products alike. I wonder if there is a need for customizing that path on a per-inspection basis and/or a per-artifact type basis. But this would make the `in_toto_verify`-call even trickier. In any case, having a single parameter is surely better than none.

*Re `link_cmd_cwd`*:
What's the benefit of having in-toto change the directory via parameter instead of doing it manually before calling the in-toto command?

*Re "permission to write (e.g., inspection link)"*: 
On a side-note, could it make sense to not write inspection links to disk unless requested by a corresponding option? Currently we only use the in-memory inspection links for further artifact rule processing. The reason why we also dump the inspection links is (a) for auditing purposes and (b) for sublayout inspection propagation, both of which are not always strictly necessary. What do you think?

trishankatdatadog:
@lukpueh 

> What's the benefit of having in-toto change the directory via parameter instead of doing it manually before calling the in-toto command?

Ease-of-use for the integrator. Let me see what this actually does behind the scenes for Popen...

> On a side-note, could it make sense to not write inspection links to disk unless requested by a corresponding option? Currently we only use the in-memory inspection links for further artifact rule processing. The reason why we also dump the inspection links is (a) for auditing purposes and (b) for sublayout inspection propagation, both of which are not always strictly necessary. What do you think?

I think it's a good idea, and this switch should be off by default. Maybe call it the `debug` flag.

lukpueh:
Regarding `Popen`'s `cwd` argument, [here is what it does on posix](https://github.com/python/cpython/blob/master/Modules/_posixsubprocess.c#L474-L475). There doesn't seem to be a `chdir` back to the original path though. Still, if you run `os.getcwd` after running subprocess with a `cwd` argument, you're still at the original path. I suppose the posix call doesn't affect Python's keeping track of `cwd`? Anyway, ease-of-use for the integrator seems like a fair reason. :)

Regarding the flag:
https://github.com/in-toto/in-toto/issues/260

trishankatdatadog:
@lukpueh This PR is not complete, because to do that, we need to update MATCH-es and ALL
OW-s in `verify_all_items_rules()` to work in the context of the cwd. How should we go about this?

SantiagoTorres:
I still think that the fact that the agent is running in a CWD it doesn't have permission to be in is a bad code smell. The fact that changing back to a cwd it doesn't have permission breaks is just a side-effect of this. This is one of the reasons things like `sudo su someuser` are discouraged...

trishankatdatadog:
@SantiagoTorres Disagree. First, we cannot control where users will run the Agent. Second, the Agent has to perform `chdir()` gymnastics just to be able to run in-toto inspection, which indicates that it is the latter that is inflexible right now. As I mentioned earlier, if you can customize where links reside, why not where inspections are run?

SantiagoTorres:
I see where you're coming from, but I still stand by what I suggest. I do advice to review that practice in the agent overall. The agent creates an entry on `/etc/passwd` when it's installed, it should have a directory set as home to which it has access to.

The idea of chdir()-ing is to well keep things isolated and, as far as I understand these chdirs() are a consequence of how the integration is made right now. To me this reads as a "let's move this problem to the in-toto codebase".

I'm not completely against this change (if at all, let's continue making this sensibly, everyone should benefit), I just want to point out that this process should have access to the directory it's running in.

trishankatdatadog:
Let's have a meeting about this sometime soon, I have a feeling this issue is best discussed in person.

SantiagoTorres:
Agreed. Let's discuss in person :+1:

adityasaky:
@trishankatdatadog is there still a requirement for this feature?

trishankatdatadog:
> @trishankatdatadog is there still a requirement for this feature?

No, let's close for now, thanks

PR #259, opened 2019-03-01
Fix illegal modification of artifact queue

lukpueh:
**Fixes issue #**:
None

During artifact rule verification `verify_item_rules` may pass the
`source_artifacts_queue` to `verify_match_rule`, which returns an
updated version of the queue (minus matched artifacts) and
overrides the original `source_artifacts_queue`.

However, the queue is passed to `verify_match_rule` by reference,
and modified directly. While this wouldn't be a problem for
`source_artifacts_queue`, as it is modified and then returned to
override itself, there is another reference pointing to that list,
i.e. (depending on whether `verify_item_rules` operates on material
or product rules) `source_products_queue` or `source_artifacts_queue`,
neither of which should not be modified by `verify_match_rule`.

This commit changes `verify_match_rule` to not modify the passed
reference and also makes an actual copy of `source_materials_queue`
and `source_products_queue` in `verify_item_rules` just to be safe.

that is provided, the better.

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -595,15 +595,6 @@ def verify_match_rule(rule, source_artifacts_queue, source_artifacts, links):
     that were successfully consumed by the rule, i.e. if there was a match with
     a target artifact.
 
-    FIXME:
-    In in-toto/in-toto#204 the behavior of the match rule was changed to NOT
-    FAIL if a required destination artifact could not be found in the
-    corresponding destination link, or if a source and destination artifact
-    pair has no matching hashes. However, the rule verification still fails
-    if a required destination link is not found.
-    As failing the overall rule verification is now left to a subsequent
-    DISALLOW rule, the "fail on missing destination link" should be removed.
-
   <Terms>
     queued source artifacts:
         Artifacts reported by the link for the step/inspection containing passed
@@ -1049,8 +1040,7 @@ def verify_item_rules(source_name, source_type, rules, links):
 
   <Arguments>
     source_name:
-            The name of the item (Step or Inspection) being verified
-            (used for user logging).
+            The name of the item (Step or Inspection) being verified.
 
     source_type:
             "materials" or "products" depending on whether the rules were in the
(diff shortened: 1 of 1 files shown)


in_toto/verifylib.py
```diff
@@ -1137,64 +1105,47 @@ def verify_item_rules(source_name, source_type, rules, links):
...
+          rule, source_materials_queue, source_artifacts_queue,
+          source_materials, source_products)
+
+    # Fail
+    elif rule_type == "disallow":
+      verify_disallow_rule(rule, source_artifacts_queue)
+
+    # Unreachable
+    else: # pragma: no cover
+      pass
```
SantiagoTorres: Would it be useful to log this in case it's really not unreachable? (I know, I know)

in_toto/verifylib.py
```diff
@@ -889,16 +879,13 @@ def verify_delete_rule(rule, source_materials_queue, source_products_queue):
...
-  for matched_material in matched_materials:
-    if matched_material in source_products_queue:
-      raise RuleVerificationError("Rule '{0}' failed, material '{1}' was found"
-          " in products but should have been deleted."
-              .format(" ".join(rule), matched_material))
+  filtered_products = fnmatch.filter(
+      source_products_queue, rule_data["pattern"])
 
-  return list(set(source_materials_queue) - set(matched_materials))
+  return set(filtered_materials) - set(filtered_products)
```
SantiagoTorres, line 879: I'm still surprised everywhere else this is treated like a set and nothing happened...

SantiagoTorres (approved):
This mostly LGTM. I know it's a draft but the changeset looks reasonable enough.

in_toto/verifylib.py
lukpueh: I still don't know you to universally best handle these. I once triggered are related discussion but we never agreed on a best practice (see https://github.com/secure-systems-lab/code-style-guidelines/issues/18).

lukpueh: But yes, pass is a bad idea. Maybe raise a custom `UnreachableError`? :)

in_toto/verifylib.py
lukpueh: Not sure I understand. Before, all `verify_*_rule` functions returned the updated queue(s) as `list` (except for `disallow`). Then I discovered that `match` also modified the queue at the passed reference, which is a bug that I fixed in 4b3028aab6b83823b1c8c78b7a8e4d9244b5e554.

And then I realized that there are other things in the way the queues are updated that need fixing, which I've started in bf0709bb96f3f5f2eeac8723c1146b627bdc84c5, in the course of which I'm trying to simplify the code and align it more with the pseudo code of the spec, e.g. having `verify_*_rule` return the consumed artifacts (as `set`) and update the queue(s) on the caller side.

The reason why we haven't noticed is probably that it requires very specific combinations and sequences of rules and artifact queues to become an issue (and probably because the code is confusing).

I'll follow up with a more precise description of the issue and thorough tests.

lukpueh:
See #261 and #262 for more details and a full fix of the here drafted issue.

Issue #261, opened 2019-03-11
Rule verification false positives


lukpueh:
**Description of issue or feature request**:

The firewall like processing of artifact rules in `verify_item_rules` uses a material queue, a product queue and a generic artifact queue. If one of the sequentially processed rules can be applied successfully, the corresponding artifacts in the corresponding queue are removed.

The queues are updated in different places, e.g. `verify_match_rule` modifies the artifacts queue at the passed reference, which btw. also points to the materials or products queue in the beginning, only to override itself upon return. Other `verify_*_rule` function only return a new updated queue to override the old queue. Depending on the rule type ("match", "allow", ...) and the rule source type ("materials" or "products"), the three queues further update each other.

Besides being hard to understand the rule processing may lead to false positives, i.e. it may fail rule verification although it shouldn't. Here is one symptomatic minimal failing example:

```python
from in_toto.models.metadata import Metablock
from in_toto.models.link import Link
from in_toto import verifylib

links = {
  "item": Metablock(signed=Link(name="item",
       products={
           "foo" : {"sha256": "deadbeef"},
       }
  )),
}

rules = [
  # ALLOW does not consume anything here but cuts reference between
  # `source_artifacts_queue` and `source_products_queue`. NOTE: Without this
  # ALLOW the behavior of the rule sequence would be as expected.
  ["ALLOW", "bogus"],

  # MATCH consumes `foo` and removes it from `source_artifact_queues`
  ["MATCH", "foo", "WITH", "PRODUCTS", "FROM", "item"],

  # CREATE does not consume anything here, and overrides
  # `source_artifacts_queue` with an outdated `source_products_queue`
  # i.e. undoes MATCH from above.
  ["CREATE", "bogus"],

  # DISALLOW fails due to `foo` being in the `source_artifacts_queue` again
  # NOTE: CREATE re-establishes the reference between the two queues, thus
  # repeating the MATCH, CREATE rule sequence before the DISALLOW rule would
  # go back to expected behavior.
  ["DISALLOW", "*"]
]

verifylib.verify_item_rules("item", "products", rules, links)
```

**Current behavior**:
Rule processing implementation is unnecessarily complicated and may lead to false positives under certain circumstances.

**Expected behavior**:
Fix false positives and simplify rule processing.

lukpueh:
[...]

PR #262, opened 2019-03-11
Major rewrite of artifact rule verification

lukpueh:
**Fixes issue #**:
#261, takes over #259

This PR includes several changes to `verify_item_rules` and the from there called `verify_*_rule` functions (plus corresponding tests):

- Harmonization of artifact queue modification

  Before, `verify_item_rules` used a material queue, a product queue and an artifact queue to keep track of the artifacts, consumed in the course of verifying all rules (of a type of an item).  These queues were modified in different places (in different functions passed to by reference) and then re-assigned to each other, sometimes as copy, sometimes as reference, which led to unexpected states of queues.

  This commit removes the materials and products queue (the generic artifacts queue is enough for keeping track of consumed artifacts).

  It further focuses all artifact queue updates in a single place in `verify_item_rules` and changes the `verify_*_rule` functions to return the consumed artifacts instead of the updated queue, which also aligns with the pseudo code in the specification. This change should make the rule processing a lot more transparent to the reader.

- Additional code simplifications
This commit also makes better use of set operations in `verify_*_rule` functions and removes clutter like
  - redundant rule unpacking code
  - unnecessary `if/else`s related to "materials" and "products"
  - unnecessarily long variable names
    the `source_` prefix only makes sense in the scope of the match rule, where there is also a destination.
The commit also updates and simplifies a lot of related code comments and docstrings.

- Full adoption of in-toto/in-toto#204
Remove raised exceptions in `verify_match_rule` and `verify_delete_rule`, where they should just not consume the corresponding artifacts.

- Rethink tests
Some of them were still tailored to pre-#204 logic. This commit makes them more meaningful (and table driven).

- Adopt queue traceback
There is only one queue now.

that is provided, the better.

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -255,7 +255,7 @@ def substitute_parameters(layout, parameter_dictionary):
   """
   <Purpose>
     This function is a transitionary measure for parameter substitution (or
-    any other solution defined by the in-toto team). As of now, it acts as 
+    any other solution defined by the in-toto team). As of now, it acts as
     a very simple replacement layer for python-like parameters
 
   <Arguments>
@@ -304,7 +304,6 @@ def substitute_parameters(layout, parameter_dictionary):
     step.expected_products = new_product_rules
 
   for inspection in layout.inspect:
-    
     new_material_rules = []
     for rule in inspection.expected_materials:
       new_rule = []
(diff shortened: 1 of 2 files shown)


lukpueh:
It might be worth cross-checking with the [corresponding section in the specification](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#4331-rule-processing).

lukpueh:
Re simplification:
I just ported this to our [golang version of in-toto's verifylib](https://github.com/in-toto/in-toto-golang/pull/35) and made a slight modification.

In go I calculate sets of [created, modified and deleted artifacts only once](https://github.com/in-toto/in-toto-golang/blob/c4e428279f428765a8d77dd7531978328d1199de/in_toto/verifylib.go#L190-L201) and before even looking at the rules, which pays off as soon as there is more than one rule of the corresponding type.

Here I re-calculate the [created](https://github.com/in-toto/in-toto/blob/f23618d90d2725094aca23956128e3b35ba9134d/in_toto/verifylib.py#L739), [deleted](https://github.com/in-toto/in-toto/blob/f23618d90d2725094aca23956128e3b35ba9134d/in_toto/verifylib.py#L778) and [modified](https://github.com/in-toto/in-toto/blob/f23618d90d2725094aca23956128e3b35ba9134d/in_toto/verifylib.py#L818) sets each time I encounter the corresponding rule.

The ideal solution would probably be to calculate each set once the first time the corresponding rule is encountered. Any thoughts?

SantiagoTorres:
> The ideal solution would probably be to calculate each set once the first time the corresponding rule is encountered. Any thoughts?

I actually like this "artifact queue compression" idea. Let me walk through it. For now I think this re-write makes things clearer to read and as such is valuable by itself...

in_toto/verifylib.py
```diff
@@ -744,297 +676,205 @@ def verify_match_rule(rule, source_artifacts_queue, source_artifacts, links):
...
-
-  unmatched_materials = set()
+  # Filter queued artifacts using the rule pattern
+  filtered_artifacts = fnmatch.filter(artifacts_queue, rule_pattern)
 
-  for matched_product in matched_products:
-    if matched_product in source_materials_queue:
-      unmatched_materials.add(matched_product)
+  # Consume filtered artifacts that are products but not materials
+  consumed = set(filtered_artifacts) & (products - materials)
```
SantiagoTorres, line 739: I love set operations :+1:

in_toto/verifylib.py
```diff
@@ -1043,70 +883,84 @@ def verify_disallow_rule(rule, source_artifacts_queue):
...
+  for source_type in ["materials", "products"]:
+    traceback_str += "Available {}{}:\n{}\n".format(
+        source_type,
+        [" (used for queue)", ""][RULE_TRACE["source_type"] != source_type],
+        RULE_TRACE[source_type])
+
+  for trace_entry in RULE_TRACE["trace"]:
+    traceback_str += "Queue after '{0}':\n".format(
+        " ".join(trace_entry["rule"]))
+    traceback_str += "{}\n".format(trace_entry["queue"])
```
SantiagoTorres, line 913: Random thought: should we encode `RULE_TRACE` as part of a decorator for function that need it?

in_toto/verifylib.py
```diff
@@ -1115,105 +969,79 @@ def verify_item_rules(source_name, source_type, rules, links):
...
+      consumed = verify_allow_rule(_pattern, artifacts_queue)
+
+    # It's up to the "disallow" rule to raise an error if artifacts were not
+    # consumed as intended
+    elif _type == "disallow":
+      verify_disallow_rule(_pattern, artifacts_queue)
+
+    else: # pragma: no cover (unreachable)
+      raise securesystemslib.exceptions.FormatError(
+          "Invaldid rule type '{}'.".format(_type))
```
SantiagoTorres, line 1036: Now that we've coalesced this interfaces almost there, wouldn't it be worth moving this into something more akin to a dispatch table??

SantiagoTorres (commented):
LGTM, I wanted to leave these three comments yet they don't affect the functionality of this.

SantiagoTorres:
Tested with the some integrator's pipeline and against old metadata. Merging :+1:

in_toto/verifylib.py
lukpueh: Why? To avoid having a global variable? There are only two places that access it -- `verify_item_rules` writes to it, and `_get_artifact_rule_traceback` reads from it. I think a decorator might be a bit of an overkill.

On a side note, we could easily remove the rule queue trace and instead update the logging statements to provide the same information. See https://github.com/in-toto/in-toto/pull/243#issuecomment-444912500 ff.

in_toto/verifylib.py
lukpueh: Sure, why not, as long as it enhances readability. How would you go about different parameters? Use some `*args`/`**kwargs` wizardry? For comparison, [here's how we do it in the go implementation](https://github.com/in-toto/in-toto-golang/blob/ef3bff026632eaf918f5f964dcb894a136a6260c/in_toto/verifylib.go#L257-L285).

Issue #264, opened 2019-03-21
Handling lack of GPG in tests


adityasaky:
**Description of issue or feature request**:
Lack of GPG on the system causes tests to fail.

**Current behavior**:
Currently, there's a check in tests/runtests.py which checks if GPG is present on a system, and if it's not, skips certain tests using a decorator.

- There's a small error in the exception it checks for, specifically `FileNotFound` (also a typo), which may need to be replaced by `subprocess.CalledProcessError`.
- The function also checks for `WindowsError` which is actually the exception returned on Windows systems. Unfortunately, this exception only exists on Windows, so if you have a system without GPG and isn't Windows (ex: Macs don't ship with GPG), this exception isn't going to be defined.
- Even with these things in place, some tests remain in `test_in_toto_run`, `test_in_toto_sign` etc that don't have the decorator, causing tests to fail. Some of these tests may have to pulled out from the main class so as to have the decorator applied to them.

Thanks to @hisenberg08 for bringing this to my attention!

lukpueh:
Thanks for submitting the issue, @adityasaky. FYI #253 already hints at it. 

Regarding your first item, you are right, `FileNotFound` doesn't exist. The exception we want to catch is called `FileNotFoundError`.
`CalledProcessError ` OTOH is raised if the command does exist but exits with non-zero.

lukpueh:
Okay, now I understand why @SantiagoTorres used a blanket exception. I seems that `WindowsError` is only available on windows and `FileNotFoundError` is only available on Python3, and on Python2 an `OSError` is raised. Luckily the latter is also a parent class of the other two. So I'm checking for `OSError` now in a tentative fix (see 23cb4d7e62233b5891e2d7c5d5b6bc07fb1d909d).

Issue #253, opened 2019-01-25
Account for and test testing corner cases such as missing commands


lukpueh:
**Description of issue or feature request**:
I just ran the in-toto test suite on a vanilla [windows vm](https://github.com/in-toto/windows-setup) without `gpg` installed and noticed that we mostly but don't fully follow through with special treatment, i.e. we
- execute some unix shell commands in the tests (`ls`) and
- exclude most but not all gpg tests, if `gpg` is not on the path.
- Also the check if `gpg` is on the path is erroneous (wrong exception)

The reason why we haven't noticed this before is that our Travis and Appveyor builds seem to support those unix commands and both have gpg installed.

Replacing the unix commands is not a big deal. But I wonder if there is a better way to handle a system without gpg, than by skipping individual tests explicitly. At any rate, if we support tests without gpg than we should also run those in our automated builds. 

Maybe we should also exclude the gpg relevant code from test coverage, if gpg is not installed.

A quick fix of the issue can be found in [lukpueh/in-toto@fix-test-corner-cases](https://github.com/lukpueh/in-toto/tree/fix-test-corner-cases).

**Current behavior**:
Partially and erroneously accounts for corner cases.

**Expected behavior**:
Fully, properly and intelligently account for testing corner cases and also test them.

lukpueh:
> the check if gpg is on the path is erroneous (wrong exception)

fixed with https://github.com/in-toto/in-toto/commit/23cb4d7e62233b5891e2d7c5d5b6bc07fb1d909d

lukpueh:
#291 makes securesystemslib's pynacl dependency optional. However, our tests currently require pynacl. Similarly to gpg, we should exclude tests that require pynacl, if pynacl is not available and configure automated tests without pynacl.

PR #265, opened 2019-03-21
Fix check gpg windows (partial)

adityasaky:
**Fixes issue #**: #264 

- Separates tests which use GPG so that decorator can be applied
- Fixes the exceptions expected by `check_usable_gpg`
- Adds decorators to other tests without separating them

At this stage this still uses `WindowsError`, and doesn't handle situations where GPG is missing on non-Windows systems.

### DIFF
--- a/tests/models/test_layout.py
+++ b/tests/models/test_layout.py
@@ -180,7 +180,7 @@ class TestLayoutMethods(unittest.TestCase):
 
     layout._validate_keys()
 
-    self.assertEqual(len(layout.get_functionary_key_id_list()), 4)
+    self.assertEqual(len(layout.get_functionary_key_id_list()), 2)
 
     # Must be a valid key object
     with self.assertRaises(securesystemslib.exceptions.FormatError):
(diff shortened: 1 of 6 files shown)


lukpueh:
@adityasaky, what is your plan for this PR? The exception issue in `tests/runtests.py` has been fixed and merged with #266. However, other related issues, some of which you are partially addressing here, remain (see #253). I'd appreciate your help on this ;)

adityasaky:
@lukpueh well, I'm currently busy with finals week, but I can perhaps pick this up for the week right after. I'll also be clearing up some pending work on TUF and I can probably get this done too, before the summer term begins.

lukpueh:
@adityasaky, `develop` has diverged quite a bit since you opened this. The gpg subpackage is in `securesystemslib` now, and there has been some activity around testing with and without optional dependencies over there. (see e.g. https://github.com/secure-systems-lab/securesystemslib/pull/200, https://github.com/secure-systems-lab/securesystemslib/pull/206, https://github.com/secure-systems-lab/securesystemslib/pull/215).

This might make some things easier, for instance, we don't have to set a `TEST_SKIP_GPG` environment variable `runtests.py` anymore. Instead we can just use `securesystemslib.gpg.constants.HAVE_GPG` in the `@unittest.skipIf` decorator.

I suggest to just close this PR here. But there might still be some work left before we can close the corresponding issues #253 and #264. Do you have capacities to take care of them?

adityasaky:
I agree, closing this PR makes sense.

> Do you have capacities to take care of them?

Unfortunately, I think I can't pick them up right now. Perhaps we can tag them as good-first-issue / help-wanted and see if there's some interest there?

PR #266, opened 2019-03-27
Support gpg key expiration verification

### LINKED ISSUE #236: in-toto should check gpg key expiration

lukpueh:
**Fixes issue #**:
Takes over #245  (Thanks @michizhou, for prototyping this!)
Fixes #236 

This PR adds support for parsing creation date and validity period of gpg primary and sub public keys from successfully verified self-signature subpackets. Two new fields are added to gpg public keys:
- `creation_time`, and
- `validity_period`

If the gpg packets carry ambiguous information we resolve as follows:
- For subpackets we only use the last subpacket in the list and prioritize hashed over unhashed subpackets (see 82368677c8abe98adfa3cb97fcd71c4957e82a95)

- For User IDs and certificates we take the key expiration time from the most recent certificate, i.e. the certificate with the highest signature creation time. Additionally, we prioritize certificates with primary user id flag set True. If the ultimately prioritized certificate does not have a key expiration time subpacket, we don't assign one, even if there were certificates of lower priority carrying that subpacket (see 5f4dfbda6ae2e865f1c8d4b5d8b6eaff6fbea425).

The PR also modifies the gpg API functions `gpg_verify_signature`, to check and raise a custom `KeyExpirationError` if the passed verification key has expired, and `gpg_sign_object` to raise a custom `CommandError` if the executed gpg command exits with non-zero, e.g. because the signing key has expired (see ae047a39fad66400e1e0d705c4e53228226de3cb and 18218239c972428864026a95bef15c9070e6fbb2).

The docstrings of calling functions are adopted accordingly. And in verifylib, where we don't count links with invalid signatures towards the threshold, we add `KeyExpirationError` as additional reason to skip a link (see 19eb3c4b6f0c014287456f046cff95e12577f174).

*OT: 23cb4d7e62233b5891e2d7c5d5b6bc07fb1d909d fixes bugs in the test aggregator script and re-adds some test verbosity that was removed in earlier commits, i.e. it prints which gpg version is available on the test system). We don't have to merge it with this PR (but we might as well).*

See commit messages for more details.

requirements**:


in_toto/gpg/common.py
```diff
@@ -334,11 +337,61 @@ def _assign_certified_key_info(bundle):
...
+      # No key expiration time, go to next certificate
+      if tmp_validity_period == None:
+        continue
+
+      tmp_sig_creation_time = \
+          signature["info"]["subpackets"].get(SIG_CREATION_SUBPACKET)
+
+      # No signature creation time, no way to resolve potential ambiguities,
+      # go to next certificate
+      # TODO: As per the spec this cannot not happen, should we blow up?
```
SantiagoTorres: I do think we'd like to blow up in this case, or log a warning at least.

in_toto/gpg/common.py
```diff
@@ -334,11 +337,61 @@ def _assign_certified_key_info(bundle):
...
+      # responsibility to the implementer.
+
+      # Ambiguity resolution scheme:
+      # We take the key expiration time from the most recent certificate, i.e.
+      # the certificate with the highest signature creation time. Additionally,
+      # we prioritize certificates with primary user id flag set True. Note
+      # that, if the ultimately prioritized certificate does not have a key
+      # expiration time subpacket, we don't assign one, even if there were
+      # certificates of lower priority carrying that subpacket.
+      tmp_validity_period = \
```
SantiagoTorres, line 356: Is there a reason to rename `key_expiration` to `validity_period`? I think it may be confusing. What do you think?

in_toto/gpg/common.py
```diff
@@ -347,12 +400,10 @@ def _get_verified_subkeys(bundle):
...
   <Purpose>
     Helper function to verify the subkey binding signature for all subkeys in
-    the passed bundle. Only valid (i.e. parsable) subkeys that are verifiably
-    bound to the the master key of the bundle are returned. All other subkeys
-    are discarded.
-
-    TODO: Extract additional information about subkeys from the signatures
-    (see inline TODO below).
+    the passed bundle in order to enrich subkeys with additional information
+    (e.g. expiration dates). Only valid (i.e. parsable) subkeys that are
```
SantiagoTorres, line 399: Note to self (and everybody) `parsable` is a word (unlike parse-able, which is an ugly composite word)

in_toto/gpg/common.py
```diff
@@ -660,6 +718,7 @@ def parse_signature_packet(data, supported_signature_types=None,
   # conflict resolution scheme that makes more sense.
   # (see RFC4880 5.2.4.1.)
   # Below we only consider the last and favor hashed over unhashed subpackets
+  # TODO: Should we warn if a we use an unhashed subpacket?
```
SantiagoTorres, line 717: I think this would be a nice-to-have if possible.

SantiagoTorres (approved):
I like this!

one small concern that I have is that this seems to have subsumed #265. Has it?

lukpueh:
Re #265: Yes, partially. It fixes a NameError and a cross-platform bug. See conversation in #264. 

But the fix for those particular bugs as proposed in #265 is actually incorrect, however, #265 provides some other modifications that should not be discarded (I volunteer to review #265 once this PR here is merged).

Btw., to make the mess complete, I also touched the mentioned lines in https://github.com/in-toto/in-toto/pull/279/commits/12d2f23745b39b8dcb3c4203c38355223fe2ce13 and https://github.com/in-toto/in-toto/pull/279/commits/c666aeb0bc4ae1014835979bad5119936792f21c of #279. But don't worry, I'll clean up accordingly, depending on which PR gets merged first. ;)

in_toto/gpg/common.py
lukpueh: As per the [RFC 4880](https://tools.ietf.org/html/rfc4880#section-5.2.3.6):
> **5.2.3.6.  Key Expiration Time**
> The validity period of the key.  This is the number of seconds after the key creation time that the key expires. [...]

So the spec calls it both. But IMO "expiration" or "expiration time" is more ambiguous. It could easily be both, the number of seconds between creation and expiration (which it is), or a set date on which it expires (which it isn't). I think `validity_period` is less ambiguous in that regard. 

That said, I'm not overly passionate about it. :)

in_toto/gpg/common.py
lukpueh: Good point! I slightly refactored the parsing in https://github.com/in-toto/in-toto/pull/266/commits/f071c95f954a9beb7398470f4a4391b11d61b8f8. Please see the commit message for more details.

We still don't blow up, but we log the problem, and with the refactor, blowing up will be easier  in the future  (also see the updated `TODO` comment).

in_toto/gpg/common.py
SantiagoTorres: Hmm, I see what you mean. I just found myself jumping where `tmp_validity_period` was defined and where it was checked, as I would read on the dict the `KEY_EXPIRATION_SUBPACKET`.

I know the RFC is not the clearest piece of text ever written (just wait for v5 keys!) but yeah...

in_toto/gpg/common.py
lukpueh: Yes, it would be. But I'm not sure where to best place the warning. Would it be okay to just keep the `TODO` note for now and revisit at a later time?

PR #280, opened 2019-04-20
in_toto_verifylib: fix require/disallow queue

SantiagoTorres:
**Fixes issue #**: None

This reverts commit 21129a6e52ad848fe34e40dcf942bea089c72060 and
replaces the consumed = artifacts_queue with consumed = set(), the
reason as to why is that a require rule passing overwrites all the queue
with a set() call...

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1038,7 +1038,11 @@ def verify_item_rules(source_name, source_type, rules, links):
     _type = rule_data["rule_type"]
     _pattern = rule_data["pattern"]
 
-    # All rules except "disallow" consume artifacts in the queue
+
+    # Initialize empty consumed set as fallback for rules that do not consume
+    # artifacts. All rules except "disallow" and "require" consume artifacts.
+    consumed = set()
+
     if _type == "match":
       consumed = verify_match_rule(
           rule_data, artifacts_queue, artifacts, links)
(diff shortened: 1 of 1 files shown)


lukpueh (commented):
Thanks for the fix, @SantiagoTorres! I referenced the PR in the list of artifact rule related PRs (https://github.com/in-toto/docs/issues/4).

What do you think about defining the empty set only once above the if/else cascade? Similar to how we do it in our [in-toto-golang](https://github.com/in-toto/in-toto-golang/blob/09155ed93d3472881d134e21145d1d7f8767d056/in_toto/verifylib.go#L257) (where the compiler forced us to do this right  in the first place).

SantiagoTorres:
I was hesitant to do that just to avoid it to allocate a new `set()` for every iteration. I wonder if what we'd like instead is to declare it as an empty set outside the `for`?

lukpueh:
Okay, fair enough. Not sure I understand your alternative proposition. We can't declare `consumed` outside the loop as it gets overridden by match/create/delete/modify/allow rules. Or did you mean declare an `empty_set = set()` and in disallow/require do something like `consumed = empty_set`?

I think my preferred solution would be one of,
- as it is now,
- what I suggested above (despite the unnecessary allocations), 
- in match/create/delete/modify/allow replace
 `consumed = verify_*_rule(...` with `artifacts_queue -= verify_*_rule(...`,
- modify `verify_disallow_rule` and `verify_require_rule` to return an empty set too and also assign it to `consumed`

lukpueh (approved):
Let's merge this soon. Let me know what you think of my suggestions (above). If none if it seems preferable we can merge as is.

SantiagoTorres:
Let's do what you suggested above despite the allocations.

lukpueh:
@SantiagoTorres, I took over your branch. And made the small modification we talked about. I hope that's okay? I also fixed the outdated docstring. Can you take a quick look at it and greenlight?

SantiagoTorres:
LGTM. Thanks!

Issue #270, opened 2019-04-10
Add mechanism for external dependency monitoring


lukpueh:
**Description of issue or feature request**:

[CII silver-level best practices](https://bestpractices.coreinfrastructure.org) require projects to
> monitor or periodically check their external dependencies (including convenience copies) to detect known vulnerabilities, and fix exploitable vulnerabilities or verify them as unexploitable.

- TUF uses pyup.io (see their [.pyup.yml](https://github.com/theupdateframework/tuf/blob/develop/.pyup.yml))
- @SantiagoTorres recommended to take a look at [nvchecker](https://github.com/lilydjwg/nvchecker).

**Current behavior**:
No automatic external dependency monitoring

**Expected behavior**:
Add automatic external dependency monitoring

trishankatdatadog:
I can vouch for pyup, it's very good and easy to use

SantiagoTorres:
I'm strongly against pyup, it just litters the git history, does absolutely no checking on why the version bump is needed and makes it impossible to search for pull requests on GitHub.

trishankatdatadog:
Actually, it does give you the changelog when it's able to find it, but use whatever you guys like, I don't really care

SantiagoTorres:
Exactly, [take a look for yourself on which of these PR's were merged without manual intervention](https://github.com/theupdateframework/tuf/pulls?page=2&q=is%3Apr+is%3Aclosed). If pyup is able to notify without littering the github history or the git commit history (e.g., by squashing all pending updates) then I'm ok with that.

trishankatdatadog:
nvchecker is not automated, is it? you have to remember to do it, right?

anyway, i don't really care what you guys end using, as long as it's reliable 🤷‍♂️

lukpueh:
While working on this issue, we should revise dependency configurations in `requirements*.txt` and `setup.py` and automation configs, i.e. `tox.ini`, `.travis.yml`, `appveyor.yml`.
Also see discussion about optional dependency in #291.

lukpueh:
Thanks for your four cents, @trishankatdatadog and @SantiagoTorres! 

[...]

PR #291, opened 2019-07-02
Make pynacl optional

joshuagl:

install_requires should list "what a project minimally needs to run
correctly"[1](https://packaging.python.org/discussions/install-requires-vs-requirements) and the presence of these dependencies is checked at runtime
when the project is installed via pip.

The securesystemslib pynacl extras allow for faster implementations of
cryptographic algorithms, but aren't required in order to run - therefore
they should not be listed in install_requires.

**Note**: this change allows me to build/run in-toto on a cross-compiled Linux distro without first having to figure out how to cross-compile PyNaCl 😄

requirements**:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,6 +24,8 @@ install:
   - python -m pip install -U pip
   # Install mock, coverage, and pytest, which are needed for unit testing.
   - pip install -r requirements-ci.txt
+  # Ensure securesystemslib[pynacl] is available as it's required for tests
+  - pip install -r requirements.txt
   - pip install -e .
 
 build: false
--- a/setup.py
+++ b/setup.py
@@ -72,9 +72,18 @@ setup(
   python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4",
   packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests",
       "debian"]),
-  install_requires=["six", "securesystemslib[crypto,pynacl]>=0.11.3", "attrs",
+  install_requires=["six", "securesystemslib[crypto]>=0.11.3", "attrs",
                     "python-dateutil", "iso8601", "pathspec",
                     "subprocess32; python_version < '3'"],
+  extras_require={
+    # Install pynacl as optional dependency to use with securesystemslib, as a
+    # workaround for `"ssl-pynacl": ["securesystemslib[pynacl]>=0.11.3"]`,
+    # which currently is not supported in "extra_require" (see pypa/pip#4957).
+    # TODO: Keep track of changes (version, additional requirements) under the
+    # "pynacl" key in securesystemslib's setup.py.
+    # https://github.com/secure-systems-lab/securesystemslib/blob/master/setup.py#L101
+    "pynacl": ["pynacl>1.2.0"]
+  },
   test_suite="tests.runtests",
   tests_require=["mock"],
   entry_points={


lukpueh:
@joshuagl, thanks for your patch and for trying to fix appveyor builds!

I'm not sure if there's a straight-forward way to run each tox build on a different appveyor build, as we do on Travis (see build matrix). IIRC that's why we didn't use tox on appveyor. But I don't think we have to fix this here. 

I suggest you just install securesystemslib with pynacl on appveyor. And I'll leave a note in #253 to, in the future, have tests for both in-toto with minimal and with extra requirements.

joshuagl:
Thanks for commenting @lukpueh 

> I suggest you just install securesystemslib with pynacl on appveyor. And I'll leave a note in #253 to, in the future, have tests for both in-toto with minimal and with extra requirements.

Would you recommend adding that to requirements-ci.txt or shall I just add a `pip install` command to appveyor.yml?

lukpueh:
Do you think it makes sense to have pynacl as extra requirement for in-toto as well? I was just playing around, and I fear something like below does not work:

 ```python
extras_require={
  "sslib-pynacl": ["securesystemslib[pynacl]>=0.11.3"]
}
```
The resolver only sees `securesystemslib` and claims that the requirement is already satisfied from `install_requires`. 

So maybe we'd have to explicitly add pynacl and keep it in sync with the extra requirement in securesystemslib, i.e.:
```python
extras_require={
  "pynacl": ["pynacl>1.2.0"]
}
```

lukpueh:
Regarding you question about *requirements file vs. pip install*, I suggest to just add a `pip install requirements.txt` before `pip install -e .` in appveyor.yml

If we decide to make pynacl available as extra requirement, we could also do something like:
```python
python install -e .[pynacl]
```

On a side note, I plan on revising the `requirements*.txt` files and how and where they are used when working on #270.

joshuagl:
> The resolver only sees `securesystemslib` and claims that the requirement is already satisfied from `install_requires`.

Seems like this is a known issue that has yet to be fixed pypa/pip#4957

> So maybe we'd have to explicitly add pynacl and keep it in sync with the extra requirement in securesystemslib

Won't that ensure pynacl is installed, but still potentially leave us without the securesystemslib extras that use it?

lukpueh:
> Won't that ensure pynacl is installed, but still potentially leave us without the securesystemslib extras that use it?

Not sure I understand. `securesystemslib[pynacl]` just [installs `'pynacl>1.2.0'`](https://github.com/secure-systems-lab/securesystemslib/blob/sslibv0.11.3/setup.py#L101) when installing `securesystemslib`. If pynacl>1.2.0 is installed via e.g `in-toto[pynacl]`, then `securesystemslib` can use it just the same.

We do need to keep the versions in sync and also adapt if `securesystemslib` adds any additional requirements under the `pynacl` key. I suggest to add a comment along the lines of:

```python
  extras_require={
    # Install pynacl as optional dependency to use with securesystemslib, as a
    # workaround for `"ssl-pynacl": ["securesystemslib[pynacl]>=0.11.3"]`,
    # which currently is not supported in "extra_require" (see pypa/pip#4957).
    # TODO: Keep track of changes (version, additional requirements) under the
    # "pynacl" key in securesystemslib's setup.py.
    # https://github.com/secure-systems-lab/securesystemslib/blob/master/setup.py#L101
    "pynacl": ["pynacl>1.2.0"]
  },
```

What do you think?

joshuagl:
> > Won't that ensure pynacl is installed, but still potentially leave us without the securesystemslib extras that use it?
> 
> Not sure I understand. `securesystemslib[pynacl]` just [installs `'pynacl>1.2.0'`](https://github.com/secure-systems-lab/securesystemslib/blob/sslibv0.11.3/setup.py#L101) when installing `securesystemslib`. If pynacl>1.2.0 is installed via e.g `in-toto[pynacl]`, then `securesystemslib` can use it just the same.

My misunderstanding. I was thinking there was extra code in the pynacl extra of securesystemslib, rather than just using it to bring in an additional dependency.

> We do need to keep the versions in sync and also adapt if `securesystemslib` adds any additional requirements under the `pynacl` key. I suggest to add a comment along the lines of:
> 
> ```python
>   extras_require={
>     # Install pynacl as optional dependency to use with securesystemslib, as a
>     # workaround for `"ssl-pynacl": ["securesystemslib[pynacl]>=0.11.3"]`,
>     # which currently is not supported in "extra_require" (see pypa/pip#4957).
>     # TODO: Keep track of changes (version, additional requirements) under the
>     # "pynacl" key in securesystemslib's setup.py.
>     # https://github.com/secure-systems-lab/securesystemslib/blob/master/setup.py#L101
>     "pynacl": ["pynacl>1.2.0"]
>   },
> ```
> 
> What do you think?

Seems like a reasonable change to me.

lukpueh:
Great! May I ask you to make the change? I think we can merge this then.

joshuagl:
> Great! May I ask you to make the change? I think we can merge this then.

Done. I added this to the first commit and force pushed the branch.

lukpueh:
Thanks for your contribution, @joshuagl!

PR #298, opened 2019-07-19
doc: add sphinx boilerplate

SantiagoTorres:
that is provided, the better.

**Fixes issue #**:

requirements**:

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,0 +1,55 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# http://www.sphinx-doc.org/en/master/config
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'in-toto'
+copyright = '2019, NYU Secure Systems Lab'
+author = 'NYU Secure Systems Lab'
+
+# The full version, including alpha/beta/rc tags
+release = '0.3.0'
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+]
+
+# Add any paths that contain templates here, relative to this directory.
+templates_path = ['_templates']
+
+# List of patterns, relative to source directory, that match files and
+# directories to ignore when looking for source files.
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = []
+
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ['_static']
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -10,9 +10,11 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-import os
-import sys
-sys.path.insert(0, os.path.abspath('.'))
+# NOTE: if you want to build the docs without installing in-toto, uncomment the
+# three lines below
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath(os.path.join('..', '..')))
 
 import in_toto
 
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -29,6 +29,10 @@ release = in_toto.__version__
 
 # -- General configuration ---------------------------------------------------
 
+# The master toctree document.
+master_doc = 'index'
+
+
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -10,9 +10,11 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
+
+import in_toto
 
 
 # -- Project information -----------------------------------------------------
@@ -22,7 +24,7 @@ copyright = '2019, NYU Secure Systems Lab'
 author = 'NYU Secure Systems Lab'
 
 # The full version, including alpha/beta/rc tags
-release = '0.3.0'
+release = in_toto.__version__
 
 
 # -- General configuration ---------------------------------------------------
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -12,7 +12,7 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('..'))
+sys.path.insert(0, os.path.abspath('.'))
 
 import in_toto
 
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -9,7 +9,7 @@ the library it provides.
    :maxdepth: 2
    :caption: Contents:
 
-
+   installing
 
 Indices and tables
 ==================
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,0 +1,20 @@
+.. in-toto documentation master file, created by
+   sphinx-quickstart on Fri Jul 19 11:48:34 2019.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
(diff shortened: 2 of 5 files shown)


SantiagoTorres:
I wonder if we can consider this an MVP, and start pushing changes piecemeal instead. Thoughts, @lukpueh ?

lukpueh:
Please fix DCO. I'll give you thorough review later. I like the idea of putting this online and then step-by-step moving our scattered docs over. I'd start by nuking our old installation instructions in the readme and adding a note about moving the docs and a link to the new installation instructions.

SantiagoTorres:
Fixed the DCO! thanks for the upcoming review!

doc/source/conf.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'in-toto'
+copyright = '2019, NYU Secure Systems Lab'
+author = 'NYU Secure Systems Lab'
+
+# The full version, including alpha/beta/rc tags
+release = '0.3.0'
```
lukpueh: Where at 0.4.1 now. 

Would  it be a bad idea to import `__version__` from `in_toto`? It would require whoever builds the docs to have in_toto installed.

doc/source/installing.rst
```diff
@@ -0,0 +1,64 @@
...
+
+If your system doesn't provide in-toto yet, you can also install it from the
+source. To do so, you will need the following dependencies:
+
+- `OpenSSL <https://openssl.org>`_
+- `python-cryptography <https://cryptography.readthedocs.io>`_
+- `python-securesystemslib <https://github.com/secure-systems-lab/securesystemslib/>`_
+- `setuptools <https://pypi.org/project/setuptools/>`_
+
+.. warning:: We strongly suggest you use a `virtual environment
```
lukpueh: I think `.. warning::`-red is a bit too harsh. What about `.. note::`-grey?

doc/source/installing.rst
```diff
@@ -0,0 +1,64 @@
...
+
+To install using pip, simply run:
+ .. code-block:: sh
+
+    pip install in-toto
+
+You may also need to install some system dependencies (depending on your host).
+These are:
+
+- `OpenSSL <https://openssl.org>`_ if you plan on using RSA signatures.
```
lukpueh: This sounds a bit as if it were optional. But OpenSSL is really a requirement. What about:
```suggestion
- `OpenSSL <https://openssl.org>`_ used to generate and verify RSA signatures.
```

doc/source/installing.rst
lukpueh: Also, I think we should move the box to the very top or bottom and say that we suggest the use of virtual environments for installations via pip or from source.

doc/source/index.rst
```diff
@@ -0,0 +1,19 @@
...
+the library it provides.
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
+   installing
+
+Indices and tables
+==================
```
lukpueh: Can we comment out that section until we add code documentation here?

lukpueh (commented):
Very very cool, @SantiagoTorres! I just built it and looks really nice. Please consider addressing my comments and we can merge.

I think we can even put it online right away, maybe after adding a`.. seealso::` box that points to the main README and explains that we are transitioning docs. 

I would then start by replacing the outdated installation section in the README with a link to your new nice installation instructions. :)

What do you think?

h01ger:
On Fri, Oct 18, 2019 at 04:24:23AM -0700, lukpueh wrote:
> Would  it be a bad idea to import `__version__` from `in_toto`? 

on the contrary, I think this is probably the only good idea here. ;)

> It would require whoever builds the docs to have in_toto installed.

I'd say this is a reasonable constraint.


-- 
cheers,
	Holger

               holger@(debian|reproducible-builds|layer-acht).org
       PGP fingerprint: B8BF 5413 7B09 D35C F026 FE9D 091A B856 069A AA1C

h01ger:
On Fri, Oct 18, 2019 at 04:33:41AM -0700, Holger Levsen wrote:
> On Fri, Oct 18, 2019 at 04:24:23AM -0700, lukpueh wrote:
> > Would  it be a bad idea to import `__version__` from `in_toto`? 
> on the contrary, I think this is probably the only good idea here. ;)

or, on a second thought, why include the version at all?


-- 
cheers,
	Holger

               holger@(debian|reproducible-builds|layer-acht).org
       PGP fingerprint: B8BF 5413 7B09 D35C F026 FE9D 091A B856 069A AA1C

doc/source/installing.rst
SantiagoTorres: Hmm, i was under the impression sslib was able not to harddepend on openssl but alas.

lukpueh: You are right, @SantiagoTorres! `securesystemslib` can be installed without openssl, although its ability to cleanly provide a pure-python interface is under revision (see [sslib#179](https://github.com/secure-systems-lab/securesystemslib/issues/179)). However, and IIRC due to sslib's questionable agility in that regard, in-toto currently does [hard-depend on sslib with openssl](https://github.com/in-toto/in-toto/blob/3d6740bc1320b85bf3c2b71a4d46ef5d6bc7643a/setup.py#L94).

Btw. the gpg subpackage also makes heavy use of pyca/cryptography and thus requires openssl, e.g. to export and verify rsa and dsa (and ed25519, see [sslib#188](https://github.com/secure-systems-lab/securesystemslib/pull/188)) signatures.

IMO we can make cryptography/openssl optional for in-toto as soon as [sslib#179](https://github.com/secure-systems-lab/securesystemslib/issues/179) is fixed. 

Also see related https://github.com/in-toto/in-toto/issues/253.

doc/source/installing.rst
lukpueh: ping

lukpueh:
Thanks for addressing my comments, @SantiagoTorres. What do you think about adding a see also box as [I suggested above](https://github.com/in-toto/in-toto/pull/298#pullrequestreview-303805429)?

doc/source/installing.rst
SantiagoTorres: Aha! thanks for clarifying.

doc/source/conf.py
```diff
@@ -0,0 +1,61 @@
...
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+import os
+import sys
+sys.path.insert(0, os.path.abspath('.'))
```
lukpueh: The corresponding commit says *"use path from the root of the repository"*, but wouldn't that be: `os.path.abspath(os.path.join('..', '..'))`?

doc/source/installing.rst
```diff
@@ -0,0 +1,63 @@
+Installation
+============
+
+You can install in-toto using either pip, or your favorite package manager.
+
+.. note:: We strongly suggest you use a `virtual environment 
+    <https://virtualenv.pypa.io/en/stable/>`_ if you are installing in_toto 
+    with pip from source
```
lukpueh:
```suggestion
    with pip or from source.
```

doc/source/installing.rst
```diff
@@ -0,0 +1,63 @@
...
+
+On Debian
+---------
+
+You can install in-toto on Debian using apt/apt-get:
+ .. code-block:: sh
+
+    apt install in-toto
+
+This should provide all dependencies you need to run in-toto
```
lukpueh:
```suggestion
This should provide all dependencies you need to run in-toto.
```

doc/source/installing.rst
```diff
@@ -0,0 +1,63 @@
...
+ .. code-block:: sh
+
+    pip install in-toto
+
+You may also need to install some system dependencies (depending on your host).
+These are:
+
+- `OpenSSL <https://openssl.org>`_ used to generate and verify RSA signatures.
+- `GPG <https://gnupg.org>`_ if you plan on generating gpg signatures (verifying
+  works without gpg)
```
lukpueh:
```suggestion
  works without gpg).
```

doc/source/installing.rst
```diff
@@ -0,0 +1,63 @@
...
+
+To install using pip, simply run:
+ .. code-block:: sh
+
+    pip install in-toto
+
+You may also need to install some system dependencies (depending on your host).
+These are:
+
+- `OpenSSL <https://openssl.org>`_ used to generate and verify RSA signatures.
```
lukpueh:
```suggestion
- `OpenSSL <https://openssl.org>`_ used to generate and verify RSA signatures, and to export and verify signatures created with gpg.
```

doc/source/conf.py
SantiagoTorres: no, that was the previous commit that assumed you were running make from `docs`. We can either change the read the docs build root path or use the current `"."` style. What do you think?

lukpueh: How do you invoke `make`? For me it only works if I call it inside `doc`. If I run it from the repository root with (`make -f doc/Makefile`) it complains about not finding `conf.py`.

If I do run it from `doc`, then `os.path.abspath('.')` in `conf.py` returns `/path/to/in-toto/doc/source` and complains about not finding `in_toto`.

doc/source/conf.py
```diff
@@ -0,0 +1,63 @@
...
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# NOTE: if you want to build the docs without installing in-toto, uncomment the
+# three lines below
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath(os.path.join(('..', '..')))
```
lukpueh:
```suggestion
# sys.path.insert(0, os.path.abspath(os.path.join('..', '..')))
```

lukpueh (approved):
hooray!

PR #300, opened 2019-09-05
Bump release 0.4.0

lukpueh:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## in-toto v0.4.0
+
+* Add REQUIRE artifact rule support (#269, #280)
+* Enhance OpenPGP key export and provide key expiration verification (#266, #288)
+* Make transitive dependency PyNaCl optional for in-toto (#291)
+* Improve automatic testing and code coverage measurement (#295) as well
+  as static analysis with pylint (#279, #296)
+* Update repository metadata
+  - Add initial 1-year roadmap (#268)
+  - Revise dependency handling for monitoring and library compatibility (#294)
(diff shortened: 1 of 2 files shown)


SantiagoTorres:
I think we've neglected this for a while, but shouldn't we update `debian/changelog` with our overarching CHANGELOG information?

SantiagoTorres:
Other than that, I think this is pretty good to go :+1:

lukpueh:
Regarding debian files, please consider reading some thoughts I described in https://github.com/secure-systems-lab/securesystemslib/pull/177#issue-313976219, which also apply here.

The TL;DR is: we probably have to update the `debian` files (see vcswatch), but we can IMHO do it after this release, because they are not part of upstream source releases anyway. Does that work for you?

lukpueh:
Btw. the up-to-date (with upstream release 0.3.0, i.e. downstream release 0.3.0-1) debian files live on the [`in-toto/in-toto@debian`](https://github.com/in-toto/in-toto/tree/debian/debian) branch.

h01ger:
On Mon, Sep 09, 2019 at 09:13:31AM -0700, Santiago Torres wrote:
> I think we've neglected this for a while, but shouldn't we update `debian/changelog` with our overarching CHANGELOG information?

not really needed (for the release), this is job of the Debian maintainers 
(which partly happen to be the same people.)

just updating CHANGELOG seems enough.


-- 
cheers,
	Holger

               holger@(debian|reproducible-builds|layer-acht).org
       PGP fingerprint: B8BF 5413 7B09 D35C F026 FE9D 091A B856 069A AA1C

PR #285, opened 2019-05-09
Add logging to help debug #282

adityasaky:
#282 has a test failing seemingly randomly, and we're as yet unable to ascertain the cause. This PR logs the contents of STDOUT and STDERR, which we can then use to see why the test is failing sporadically.

Issue #282, opened 2019-04-30
test_in_toto_run_with_byproduct test fails sporadically


lukpueh:
**Description of issue or feature request**:
I've see recently seen these appear sporadically on different Python versions. 
```
======================================================================
FAIL: test_in_toto_run_with_byproduct (tests.test_runlib.TestInTotoRun)
Successfully run, verify recorded byproduct.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/lukpueh/in-toto/tests/test_runlib.py", line 495, in test_in_toto_run_with_byproduct
    "Python" in link.signed.byproducts.get("stderr"))
AssertionError: False is not true
Stderr:
in_toto.runlib:501:INFO:Running 'test_step'...
in_toto.runlib:524:INFO:Running command 'python --version'...
in_toto.runlib:538:INFO:Creating link metadata...
```

**Current behavior**:
- Meaningless test output (i.e. `AssertionError: False is not true`)
- Test fails sporadically

**Expected behavior**:
- Enhance test output (i.e. add message to `self.assertTrue(<expr>, <message>)`)
- Fix test

lukpueh:
Unsurprisingly this bug is due to a race condition in the `while proc.poll() ...`- loop in  `run_duplicate_streams`:

https://github.com/in-toto/in-toto/blob/a481e474d39aae6b77ec8b9ca1597f919b4ada1d/in_toto/process.py#L194-L212

The subprocess might write something to its stdout/stderr files after the last read from the parent, but exit before the next read.

The original author of that feature had an additional [post-loop read](https://github.com/in-toto/in-toto/pull/160/files#diff-a6df82494c4bf0be4600800b771dd778R298-R302) probably for that reason, which didn't survive the port, when the PR was taken over. (*IIRC the `... or stdout_part or stderr_part` in the loop condition were meant to serve a similar purpose. But looking at them now, they don't make much sense...)*

So, assuming that the last write to a stdout/stderr file happens before the process exits, than a post-loop read (or maybe a do-until loop) should fix this issue. Otherwise, if the process can exit before the data becomes available, solving will be more complex.

Apart from above, we should probably also add back the [comments from the original author regarding buffered outputs](https://github.com/in-toto/in-toto/pull/160/files#diff-a6df82494c4bf0be4600800b771dd778R267-R271).
The function does not duplicate in realtime if the child process buffers its outputs, as e.g. python's print does unless you tell it not to. Note the difference between...

```python
[...]

PR #309, opened 2019-09-25
Add downstream patches to fix testing in debian 

lukpueh:
**Fixes issue #**:
None.

A few tests call the python interpreter in a subprocess, e.g. to test our custom `process` module or to run simple in-toto step or inspection commands.

In debian, where we only build for python3, "python" is not available during build, and thus makes those tests fail.

This commit adds quilt downstream patches to change the used interpreter to "python3" in the relevant tests (this approach is backed by our Debian-man @h01ger).

Here's how I applied the patch and built in-toto 0.4.0 with tests:

```
 # Assuming chroot is set up
 # See https://wiki.debian.org/sbuild

 # Download upstream tarball
wget https://github.com/in-toto/in-toto/releases/download/v0.4.0/in-toto-0.4.0.tar.gz \
  -O python-in-toto_0.4.0.orig.tar.gz

 # Extract upstream tarball
tar xf python-in-toto_0.4.0.orig.tar.gz

 # Clone upstream repo for debian files and tests
git clone https://github.com/in-toto/in-toto.git in-toto-upstream

 # Copy debian files
cp -r in-toto-upstream/debian in-toto-0.4.0/

 # Change into unpacked tarball
cd in-toto-0.4.0

 # Apply patch in debian/patches/
quilt push

 # Build with tests
sbuild -A -s -d unstable
```

*An alternative for most tests might be, to use `sys.executable` instead of the hardcoded interpreter name. But this does not work for the inspection ("python ./scripts/tar ...") defined as string in "demo.layout.template".*

requirements**:


lukpueh:
So I've fixed all these problems (lintian warnings, missing gnupg dependency) and now tests keep failing at `test_in_toto_run_with_byproduct (tests.test_runlib.TestInTotoRun)`. This issues has been described in https://github.com/in-toto/in-toto/issues/282, but there's no fix for it yet. I'll look at it with a fresh pair of eyes early next week.

adityasaky:
Edit: Ah I get it now. Would you expect this to fail for in-toto regularly even if you're not specifically building it? I spun up a debian instance to see if tests would fail there when just cloning the repository and `test_in_toto_run_with_byproduct` did fail there with no STDOUT and STDERR entries that #285 should have got us. Is this the same thing you see? Or is this unrelated?

The link file I see is:

```
{
 "signatures": [],
 "signed": {
  "_type": "link",
  "byproducts": {
   "return-value": 0,
   "stderr": "",
   "stdout": ""
  },
  "command": [
   "python",
   "--version"
  ],
  "environment": {},
  "materials": {},
  "name": "test_step",
  "products": {}
 }
}
```

In `run_duplicate_streams`, the issue seems to be that the `stdout_reader` isn't reading the data from the tmpfile. I inspected the file itself - the output is there.

lukpueh:
Tests are running through when building downstream with the debian/* files in this branch now. :)
 See https://mentors.debian.net/package/in-toto for latest successful upload to mentors.

The commits in this PR do not only address @h01ger's review comments but also hot-fix some bugs from upstream (see #282 and https://github.com/secure-systems-lab/securesystemslib/pull/186).

debian/upstream/signing-key.asc
```diff
@@ -29,48 +24,29 @@ Vtvua2wFGj9DrxO+WQIPZrX/jJu7ntqxL2I4zLKRrvnMjAGtjT3I7zLsMvLaZN02
...
-G3C4CogmNe13vjpzW5e6/jpcq9fvjMd7iW+kV8+IrEYcw8xYo4YB7MiZYw9EGkiN
-cg22w2wkqJUDQQSIFI6oUJE8iL2e1twZXfjcoA3uFsaSP0823DTFKwkjNAZVTfWT
-tZzLzsmMVjRjJd72VvN9ChZYlQZCkkBMNQuisIAG9/kLp7KQxELClI7gB1/yJGdJ
-dSZBzF0b2Na+3co13ZJCKpnvEg7AOH+Dys1cV3FO45/kSmkBD+cUAdJeD7zMpTRM
-jhk8lyMNB4JtWXMfW8V39M2IkG+YH0d6ogaqAjNv5/IyR/whpkzcRlgqRUdvEtcB
-2Oemsf7tTE2CD/bGRp2Byezdyl7q2JaUXY8WXUQ05ko=
-=2tbg
------END PGP PUBLIC KEY BLOCK-----
\ No newline at end of file
+7L3JGDlPqKmG4qStrgzSxb/8Bpa5Ag0EVv7N3AEQALxccl4AppwwA4tCcxKdC0Ho
```
SantiagoTorres, line 27: Just curious, why is your pubkey block shorter now?

lukpueh: I exported with the `export-minimal` option ([as suggested by Lintian](https://lintian.debian.org/tags/public-upstream-key-not-minimal.html)). 

From the GPG manpages:

```
export-minimal
  Export the smallest key possible. This removes all signatures except the most
  recent self-signature on each user ID. This option is the  same as running
  the --edit-key command "minimize" before export except that the local copy of
  the key is not modified. Defaults to no.
```

lukpueh:
Updates were reviewed by @h01ger and used to put in-toto 0.4.0-1 into debian NEW queue.

Also see https://mentors.debian.net/package/in-toto

PR #332, opened 2019-11-26
doc: add command line tools

SantiagoTorres:
**Fixes issue #**:
None

**Description of the changes being introduced by the pull request**:

This PR adds a small series of documents to describe the current CLI tools. These are not meant to be exhaustive (i.e., to include every single command line argument), but rather to provide a little insight on how these tools can be used.

**Please verify and check that the pull request fulfills the following
requirements**:

- [x] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [N/A] Tests have been added for the bug fix or new feature
- [x] Docs have been added for the bug fix or new feature

Issue #239, opened 2018-10-26
Rethink command hierarchy and add usage examples


michizhou:
Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.

**Description of issue or feature request**:
Most of the issues mentioned by @vladimir-v-diaz that concerned the README's command line tool descriptions in his gpg support review (#150) have been resolved (#234). However, the rethinking of the command hierarchy and the inclusion of command usage examples have yet to be implemented.

**Current behavior/Expected behavior**:
From Vlad's comments:

- Rethink the command hierarchy, e.g. have in-toto as single base command with subcommands (`in-toto <subcommand>`), instead of in-toto-run, in-toto-verify, in-toto-record, ... https://github.com/in-toto/in-toto/pull/150#discussion_r155337421
- Add additional usage examples for all commands to reduce emphasis on supported command-line options for better user experience https://github.com/in-toto/in-toto/pull/150#discussion_r155332638

lukpueh:
This is really a nice-to-have feature. Removing 1.0 label and adding low-prio.

PR #341, opened 2020-01-16
Major CLI documentation overhaul and automation

### LINKED ISSUE #277: Define stable API

As per our [roadmap](https://github.com/in-toto/in-toto/blob/develop/ROADMAP.md#stable-api), the
> current API for interacting with the in-toto CLI and its library hasn't changed much, yet there are no guarantees of this happening. During this year we'll revisit the API definition and settle for a well-defined user-facing interface for metadata operations, key creation/loading/usage and verification.

This is issue page is meant for discussion and should primarily entail updates in the documentation (see related #276) to clearly define a stable API and CLI for a 1.0.0 release. It should also assess whether and what changes to the implementation are required and initiate them.

Also see other related [API](https://github.com/in-toto/in-toto/labels/API%20specification) and [CLI](https://github.com/in-toto/in-toto/labels/CLI%20specification) specification issues

**Current behavior**:
No clearly defined API 

**Expected behavior**:
Clearly defined API and CLI, primarily via documentation.

lukpueh:
**Fixes issue #**:
Partially fixes #277 and #239 
Supersedes #332 

This PR performs a major revision of all in-toto cli tools so that the description-, help and usage texts can be used as single source of truth, both in the command line (e.g. when using the `--help` flag on any of the `in-toto-*` commands, and when auto-generating sphinx documentation (e.g. for readthedocs). 

Changes include:
- refactor cli modules and enhance contents and format of help texts (8580862, cd30f84, eccedc8, 881ca1e, ef48a7e, 6ef8f95)
- add custom sphinx argparse extension and docs generation configuration (ed2dec8, ea5cf92)
- streamline cli module document headers and update docs links in main in-toto README.md (3b5c04f, 5578d3e)

Apologies for the size of the diff! I Strongly suggest to review commit by commit, reading the commit messages. I also suggest to go through the rendered documentation on readtheodcs:
https://in-toto-lukpueh.readthedocs.io/en/latest/command-line-tools/index.html

Big kudos to @SantiagoTorres and his work in #332, which I have integrated in this PR.

requirements**:


tests/test_common_args.py
```diff
@@ -0,0 +1,78 @@
...
+    """Create an empty parser and perform some basic assertions prior to
+    testing parser action group (i.e. argument group) helper functions. """
+    # Create empty argument parser
+    self.parser = argparse.ArgumentParser()
+
+    # Assert parser has the protected member "_action_groups" and it is a list
+    # NOTE: argparse could remove this at any time without notice
+    self.assertTrue(
+        isinstance(getattr(self.parser, "_action_groups", None), list))
+
```
SantiagoTorres, line 35: I like we are testing against this, but I've always been wary of `type` vs `isinstance`. I personally don't know if we want to ensure it's a list-ish type, or must be a list and nothing else. What's your take?

in_toto/util.py
```diff
@@ -186,8 +186,8 @@ def import_public_keys_from_files_as_dict(filepaths, key_types=None):
 
 def import_gpg_public_keys_from_keyring_as_dict(keyids, gpg_home=False):
   """Creates a dictionary of gpg public keys retrieving gpg public keys
-  identified by the list of passed `keyids` from the gpg keyring at `gpg_home`.
-  If `gpg_home` is False the default keyring is used. """
+  identified by the list of passed `keyids` from a gpg keyring in `gpg_home`.
+  If `gpg_home` is False the default gpg home directory is used. """
```
SantiagoTorres, line 190: Thanks for disambiguating this. It makes me wonder whether we want to use `--keyring` instead later down the line though. What do you think?

README.md
```diff
@@ -133,7 +136,8 @@ formats or signing routines change.
 
 For a detailed list of all command line arguments and their usage, run
 `in-toto-sign --help` or look at the
-[code documentation](https://github.com/in-toto/in-toto/blob/develop/in_toto/in_toto_sign.py).
+[online
+documentation](https://in-toto.readthedocs.io/en/latest/command-line-tools/in-toto-sign.html).
```
SantiagoTorres, line 140: Is it me, or the line wrapping here also causes the links to be split? (or it could be the richdiff viewer being wonky as well..)

SantiagoTorres:
Mostly minor nits, although I'd like to bring some of these to discussion. Great work!!

README.md
lukpueh: AFAICS the markdown parser can deal line breaks inside brackets. The rendered README also suggests that that's the case: https://github.com/in-toto/in-toto/blob/5578d3e871171cc1f9bcce9f4228a58c2ab7ef27/README.md

in_toto/util.py
lukpueh: I wouldn't replace it because there is merit to overriding the homedir and we do this in tests a lot. But it would definitely be nice to add support for the`--keyring` option. Maybe worth a feature request in securesystemslib?

tests/test_common_args.py
lukpueh: You are absolutely right. In this case we should do a `type` and not an `instance` check, because in `common_args.sort_action_groups` we replace what `_action_groups` references with a list.  So if `argparse` were to change the type to something list-ish, it could lead to unexpected behavior and we might not detect it in the test here. Thanks for pointing this out. I'll push a change.

doc/source/installing.rst
```diff
@@ -38,22 +38,22 @@ To install using pip, simply run:
 You may also need to install some system dependencies (depending on your host).
 These are:
 
-- `OpenSSL <https://openssl.org>`_ used to generate and verify RSA signatures, and to export and verify signatures created with gpg.
-- `GPG <https://gnupg.org>`_ if you plan on generating gpg signatures (verifying
-  works without gpg).
+- `OpenSSL <https://openssl.org>`_ used to generate and verify RSA signatures, and to export and verify signatures created with GPG.
+- `GPG <https://gnupg.org>`_ if you plan on generating GPG signatures (verifying
```
SantiagoTorres: I wonder if we'd like to say `if you plan on generating PGP signatures` instead.

doc/source/installing.rst
```diff
@@ -38,22 +38,22 @@ To install using pip, simply run:
...
+  works without GPG).
 
 Installing from Source
 ----------------------
 
-If your system doesn't provide in-toto yet, you can also install it from the
+If your system doesn't yet provide in-toto, you can also install it from the
```
SantiagoTorres: I'm not completely convinced moving this yet is necessary? I'm also wondering if we could simplify this:

```
if your system doesn't provide in-toto, you can install it from the source
```

to be a little bit more succint.

SantiagoTorres (commented):
I found two minor nits (that I'm not completely sure are worth fixing). I can merge afterwards. Thanks for this!

doc/source/installing.rst
lukpueh: Or we say `[PG]{3}` and let the reader decide, additionally we should add a reference to [RFC 4880 -- 1.1. Terms](https://tools.ietf.org/html/rfc4880#section-1.1) :)

lukpueh:
Thanks for the feedback, @SantiagoTorres. I just pushed a commit that incorporates both of your suggestions.

SantiagoTorres:
Thank you!

PR #345, opened 2020-01-27
Add tmp dir and gpg key test mixins.

### LINKED ISSUE #87: Clean up unit-test code.

lukpueh:
**Fixes issue #**:
Supersedes #216
Partially fixes #87 

- Adds two test mixins `TmpDirMixin` and `GPGKeysMixin` with methods and change to a temporary directory, change back to the original cwd and delete the temporary directory, and to copy gpg rsa keys to the cwd. 

- Removes repetitive code from test setup and teardown methods and uses mixins instead

NOTE: the mixin has a class variable `gnupg_home` which contains the path to the keyring relative to cwd. This fixes an issue with gpg on windows that has troubles migrating old keyrings if the homedir
is passed as absolute path with drive prefix (e.g. `C:/path/to/keyring`).
 

requirements**:

### DIFF
--- a/tests/test_in_toto_mock.py
+++ b/tests/test_in_toto_mock.py
@@ -19,18 +19,17 @@
 import os
 import unittest
 import shutil
-import tempfile
 import logging
 
 from in_toto.in_toto_mock import main as in_toto_mock_main
 
-import tests.common
+from tests.common import CliTestCase, TmpDirMixin
 
 # Required to cache and restore default log level
 logger = logging.getLogger("in_toto")
 
 
-class TestInTotoMockTool(tests.common.CliTestCase):
+class TestInTotoMockTool(CliTestCase, TmpDirMixin):
   """Test in_toto_mock's main() - requires sys.argv patching; and
   in_toto_mock- calls runlib and error logs/exits on Exception. """
   cli_main_func = staticmethod(in_toto_mock_main)
@@ -39,16 +38,12 @@ class TestInTotoMockTool(tests.common.CliTestCase):
   def setUpClass(self):
     """Create and change into temporary directory,
     dummy artifact and base arguments. """
+    self.set_up_test_dir()
 
     # Below tests override the base logger ('in_toto') log level to
     # `logging.INFO`. We cache the original log level before running the tests
     # to restore it afterwards.
     self._base_log_level = logger.level
-
-    self.working_dir = os.getcwd()
-
-    self.test_dir = tempfile.mkdtemp()
-    os.chdir(self.test_dir)
 
     self.test_step = "test_step"
     self.test_link = self.test_step + ".link"
@@ -58,12 +53,11 @@ class TestInTotoMockTool(tests.common.CliTestCase):
 
   @classmethod
   def tearDownClass(self):
-    """Change back to initial working dir and remove temp test directory. """
-    os.chdir(self.working_dir)
-    shutil.rmtree(self.test_dir)
+    self.tear_down_test_dir()
 
     # Restore log level to what it was before running in-toto-mock
     logger.setLevel(self._base_log_level)
+
 
   def tearDown(self):
     try:
--- a/tests/test_in_toto_mock.py
+++ b/tests/test_in_toto_mock.py
@@ -18,7 +18,6 @@
 """
 import os
 import unittest
-import shutil
 import logging
 
 from in_toto.in_toto_mock import main as in_toto_mock_main
--- a/tests/common.py
+++ b/tests/common.py
@@ -68,10 +68,14 @@ class TmpDirMixin():
 
 class GPGKeysMixin():
   """Mixin with classmethod to copy GPG rsa test keyring to a subdir 'rsa' in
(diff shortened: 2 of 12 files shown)


lukpueh:
For documentation purposes, I'd like to share debug output from an [appveyor builder](https://www.appveyor.com/docs/how-to/rdp-to-build-worker/) that shows how it struggles with absolute paths to non-default homedirs if the gpg keyring needs to be migrated.



This directory contains a gpg keyring in an old format.
```bash
C:\projects\in-toto\tests\gpg_keyrings>ls rsa
7B3ABB26B97B655AB9296BD15B0BD02E1C768C43.ssh  pubring.gpg  trustdb.gpg
8288EF560ED3795F9DF2C0DB56193089B285DA58.ssh  random_seed
8465A1E2E0FB2B40ADB2478E18FB3F537E0C8A17.ssh  secring.gpg
```

When e.g. trying to sign with a newer version of gpg, it will try to migrate the format of the keys, which fails.
```
C:\projects\in-toto\tests\gpg_keyrings>echo "TFW" | gpg --detach-sign --homedir C:/projects/in-toto/tests/gpg_keyrings/rsa
gpg: starting migration from earlier GnuPG versions
gpg: can't connect to the agent: IPC connect call failed
gpg: error: GnuPG agent unusable. Please check that a GnuPG agent can be started.
gpg: migration aborted
gpg: can't connect to the agent: IPC connect call failed
gpg: can't connect to the agent: IPC connect call failed
gpg: keydb_search failed: No agent running
gpg: no default secret key: No agent running
gpg: signing failed: No agent running
```

Doing the same thing, but passing a relative instead of an absolute path to the keyring, the migration 
 and signature creation succeeds.
```
C:\projects\in-toto\tests\gpg_keyrings>echo "TFW" | gpg --detach-sign --homedir rsa
gpg: starting migration from earlier GnuPG versions
gpg: porting secret keys from '/c/projects/in-toto/tests/gpg_keyrings/rsa/secring.gpg' to gpg-agent
gpg: key 18FB3F537E0C8A17: secret key imported
gpg: key 5B0BD02E1C768C43: secret key imported
gpg: key 56193089B285DA58: secret key imported
gpg: key D2C9FE930766998D: secret key imported
gpg: key 9EA70BD13D883381: secret key imported
gpg: key 7CB07D6D2C199C7C: secret key imported
gpg: migration succeeded
[TRUNCATING SIGNATURE OUTPUT (bytes)]
```

Once migration has happened gpg accepts absolute paths too and can successfully create signatures.
```
C:\projects\in-toto\tests\gpg_keyrings>echo "TFW" | gpg --detach-sign --homedir C:/projects/in-toto/tests/gpg_keyrings/rsa
[TRUNCATING SIGNATURE OUTPUT (bytes)]
```

It seems that [gpg starts the gpg-agent](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob;f=common/asshelp.c;h=83c3787868463cd826e97953d65bbed29a3340f2;hb=HEAD#l531) but is then [not able to connect to it](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob;f=common/asshelp.c;h=83c3787868463cd826e97953d65bbed29a3340f2;hb=HEAD#l547). Note though that this only happens if migration is necessary (i.e. coming from [migrate.c](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob;f=g10/migrate.c;h=c52c67a77c54584eaebf3c11ac4753a07e016429;hb=HEAD#l77)). If migration has already happened, absolute paths are accepted, even if the agent has to be restarted...

This PR should sufficiently (for our purposes) work around this issue.

tests/test_in_toto_mock.py
```diff
@@ -58,13 +52,12 @@ def setUpClass(self):
...
   def tearDownClass(self):
-    """Change back to initial working dir and remove temp test directory. """
-    os.chdir(self.working_dir)
-    shutil.rmtree(self.test_dir)
+    self.tear_down_test_dir()
 
     # Restore log level to what it was before running in-toto-mock
     logger.setLevel(self._base_log_level)
 
+
```
adityasaky, line 60: I think there's been some inconsistency here. As I understand it, methods within classes have one blank line before and after. I think we have a mix of single lines and double lines throughout the codebase, and probably something a linter will catch?

adityasaky (approved):

tests/test_in_toto_mock.py
lukpueh: You're right. Seems like pylint does not support that though. Before we investigate other linters, we should overhaul our code style guidelines. IMO the current recommendations for blank lines (5 lines between top-level functions) is overkill. In in-toto we usually use 2 or 3. I think a slight inconsistency is not a very big deal.

lukpueh:
Thanks for the review, @adityasaky!

PR #357, opened 2020-04-09
WIP Command Standard Streams

cindykimxp:
Please fill in the fields below to submit a pull request. The more information
that is provided, the better.

**Fixes issue #:**
in-toto#351, the issues presented in runlib.in-toto-run

**Description of the changes being introduced by the pull request:**
- To correctly print streams to console or record streams in link metadata at appropriate times
- Add quiet keyword argument (default True) to enable (False) printing streams to console
- Use record_streams keyword (default False) to enable (True) recording streams in link metadata

Follows the logic:
When record_streams is true:
- if quiet is true --> link metadata
- if quiet is false --> both link metadata and terminal

When record_streams is false:
- if quiet is true --> nowhere
- if quiet is false --> terminal

**Note:** changes to in-toto-run have not yet been made; there may be a need to add an argument under runlib.in_toto_run(...quiet=args.quiet)...
**Additional review requests:** "process.stdout" and "process.stderr"... a bit unsure if these are implemented correctly (info on these: https://nodejs.org/api/process.html#process_process_stderr) (https://nodejs.org/api/process.html#process_process_stdout)

Please verify and check that the pull request fulfills the following
requirements:
 - The code follows the Code Style Guidelines
 - Tests have been added for the bug fix or new feature
 - Docs have been added for the bug fix or new feature


in_toto/runlib.py
```diff
@@ -319,7 +319,7 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
 
   return artifacts_dict
 
-def execute_link(link_cmd_args, record_streams):
+def execute_link(link_cmd_args, record_streams, quiet):
```
adityasaky, line 322: Is there a default for `quiet`?

in_toto/runlib.py
```diff
@@ -350,15 +350,29 @@ def execute_link(link_cmd_args, record_streams):
     - The return value of the executed command.
   """
   if record_streams:
-    return_code, stdout_str, stderr_str = \
-        securesystemslib.process.run_duplicate_streams(link_cmd_args)
+    if (quiet == False): #record_streams true, quiet false
```
adityasaky:
```suggestion
    if quiet is False: #record_streams true, quiet false
```
or
`if not quiet`
Alternatively, `if quiet:` with the body that you currently have under `else` should also work.

in_toto/runlib.py
```diff
@@ -350,15 +350,29 @@ def execute_link(link_cmd_args, record_streams):
...
+     process = securesystemslib.process.run(link_cmd_args, check=False,
+     stdout=securesystemslib.process.PIPE,
+     stderr=securesystemslib.process.PIPE)
+     stdout_str = process.stdout
+     stderr_str = process.stderr
+     return_code = process.returncode
 
   else:
-    process = securesystemslib.process.run(link_cmd_args, check=False,
+    if (quiet == False): #record_streams false, quiet false
```
adityasaky: Same as above

in_toto/runlib.py
```diff
@@ -350,15 +350,29 @@ def execute_link(link_cmd_args, record_streams):
...
   """
   if record_streams:
-    return_code, stdout_str, stderr_str = \
-        securesystemslib.process.run_duplicate_streams(link_cmd_args)
+    if (quiet == False): #record_streams true, quiet false
+      return_code, stdout_str, stderr_str = \
+      securesystemslib.process.run_duplicate_streams(link_cmd_args)
+    else: #record_streams true, quiet true
+     process = securesystemslib.process.run(link_cmd_args, check=False,
+     stdout=securesystemslib.process.PIPE,
```
adityasaky:
```suggestion
       stdout=securesystemslib.process.PIPE,
```

in_toto/runlib.py
```diff
@@ -350,15 +350,29 @@ def execute_link(link_cmd_args, record_streams):
...
+     stderr=securesystemslib.process.PIPE)
```
adityasaky:
```suggestion
       stderr=securesystemslib.process.PIPE)
```

in_toto/runlib.py
```diff
@@ -350,15 +350,29 @@ def execute_link(link_cmd_args, record_streams):
...
   else:
-    process = securesystemslib.process.run(link_cmd_args, check=False,
+    if (quiet == False): #record_streams false, quiet false
+      process = securesystemslib.process.run(link_cmd_args, check=False, 
+      stdout=None, stderr=None)
+      stdout_str = stderr_str = ""
+      return_code = process.returncode
+    else: #record_streams false, quiet true
+      process = securesystemslib.process.run(link_cmd_args, check=False,
       stdout=securesystemslib.process.DEVNULL,
```
adityasaky:
```suggestion
        stdout=securesystemslib.process.DEVNULL,
```

in_toto/runlib.py
```diff
@@ -350,15 +350,29 @@ def execute_link(link_cmd_args, record_streams):
...
       stderr=securesystemslib.process.DEVNULL)
```
adityasaky:
```suggestion
        stderr=securesystemslib.process.DEVNULL)
```

in_toto/runlib.py
adityasaky:
```suggestion
def execute_link(link_cmd_args, record_streams, quiet=True):
```

going by the call below

in_toto/runlib.py
```diff
@@ -528,7 +542,7 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
 
   if link_cmd_args:
     LOG.info("Running command '{}'...".format(" ".join(link_cmd_args)))
-    byproducts = execute_link(link_cmd_args, record_streams)
+    byproducts = execute_link(link_cmd_args, record_streams, quiet=True)
```
adityasaky:
```suggestion
    byproducts = execute_link(link_cmd_args, record_streams, quiet=quiet)
```

adityasaky: By using `quiet=True`, you're not actually using the `quiet` that was passed to `in_toto_run`, but rather setting a keyword arg for `execute_link`. :)

adityasaky (changes requested):
Thanks for the PR, @cindykimxp. Just some quick changes, which I think will also help with the CI errors.

adityasaky:
~Also, [`securesystemslib.process.run`](https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/process.py#L55) calls [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run). I suspect you'd have to pass `capture_output=True` with `stdout` and `stderr` to capture those streams. @lukpueh?~

Ah I think I misunderstood the condition initially, nvm.

in_toto/runlib.py
```diff
@@ -350,15 +350,29 @@ def execute_link(link_cmd_args, record_streams):
     - The return value of the executed command.
   """
   if record_streams:
-    return_code, stdout_str, stderr_str = \
-        securesystemslib.process.run_duplicate_streams(link_cmd_args)
+    if (quiet == False): #record_streams true, quiet false
+      return_code, stdout_str, stderr_str = \
+      securesystemslib.process.run_duplicate_streams(link_cmd_args)
+    else: #record_streams true, quiet true
+     process = securesystemslib.process.run(link_cmd_args, check=False,
```
adityasaky: Also, these have been indented with a single space instead of two.

in_toto/runlib.py
cindykimxp: Oh, yes! I should've included this! Thanks for catching that

in_toto/runlib.py
cindykimxp: Yeah, I can definitely clean it up a bit. I'll move it around without the parentheses and the == (old school convention habits!) Thanks!

in_toto/runlib.py
cindykimxp: Oh, YES! I was looking for the correct way to do this one. I wasn't quite sure about this -- thank you!

in_toto/runlib.py
cindykimxp: I think I made the changes you suggested! Let me know if it hasn't correctly been updated -- and thanks so much!

lukpueh (commented):
Many thanks for working on this, @cindykimxp! And thanks for your review, @adityasaky!

This already looks good. Here is a bit of feedback from my side:

- We need to make sure that the tests pass.
The reason why they only fail on Python 3 is that a Python 3 subprocess per default returns stdout and stderr with the wrong data type for our purposes.
I suggest you check out the `securesystemslib.process.run(...)` function and the documentation of the underlying Python function (hint: Use  node.js documentation won't help you much there).

- We should follow our code style guide
Please use double indentation (4 spaces) for line-continuation, this makes it easier to distinguish it from single indentation (2 spaces) that indicate a new block of code.

- (bonus) These are some high-level style considerations
You have four distinct blocks of code now in `execute_link`, that are executed depending on the combination of the values of `record_streams` and `quiet`. The latter three look fairly similar. Maybe you can do common things in those blocks only once? ... just like cancellation in math. :)

Thanks again for adding features to in-toto and please let me know if you have any questions.

in_toto/runlib.py
```diff
@@ -356,13 +356,25 @@ def execute_link(link_cmd_args, record_streams):
     - The return value of the executed command.
   """
   if record_streams:
-    return_code, stdout_str, stderr_str = \
-        securesystemslib.process.run_duplicate_streams(link_cmd_args)
+    if not quiet: #record_streams true, quiet false
+      return_code, stdout_str, stderr_str = \
+      securesystemslib.process.run_duplicate_streams(link_cmd_args)
```
lukpueh: Please use double indentation here.

in_toto/runlib.py
```diff
@@ -319,7 +319,7 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
 
   return artifacts_dict
 
-def execute_link(link_cmd_args, record_streams):
+def execute_link(link_cmd_args, record_streams, quiet=True):
```
lukpueh: I think `quiet` could be mandatory here. `execute_link` is more or less internal API and only called from `in_toto_run`. For `in_toto_run` OTOH `quiet` should indeed be optional.

lukpueh (commented):
💯  for figuring out the encoding issue! :) Here are some suggestions for next steps:

- Double indent line continuation (see inline comment below)
- Fix the linter issue flagged by Travis
- Add a description of the argument to the docstrings of `execute_link` and `in_toto_run`.
  NOTE: `execute_link` uses the old secure-systems-lab docstring style and `in_toto_run` uses the new google docstring style, please make it consistent with the existing styles.

SantiagoTorres:
Closing this one in favor of #393

Issue #304, opened 2019-09-11
Add a flag to accept a target directory to put generated metadata in


adityasaky:
Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.

**Description of issue or feature request**: Currently, the metadata is generated in the directory `in-toto-run` is run in. By adding an appropriate flag, users can specify where to write the generated metadata to.

This can be handy when there are many steps being run from the command line and all link files can be sent to a consistent location for future verification.

**Current behavior**: Doesn't exist

**Expected behavior**:
```
$ in-toto-run <...> --metadata-directory <path> -- <command>
```

adityasaky:
Closed by #364. Great work, @SolidifiedRay!

PR #364, opened 2020-05-27
Add a new flag to specify where to write the generated metadata

SolidifiedRay:
Fix issue #304 

Description of the changes being introduced by the pull request:
A new flag, --metadata-directory (shortcut: -d), is added to specify where to write the generated metadata. 
Expected behavior:
$ in-toto-run <...> --metadata-directory <path> -- <command>

### DIFF
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -122,6 +122,15 @@ QUIET_KWARGS = {
   "help": "suppress all output"
 }
 
+METADATA_DIRECTORY_ARGS = ["-d", "--metadata-directory"]
+METADATA_DIRECTORY_KWARGS = {
+  "required": False,
+  "type": str,
+  "metavar": "<directory>",
+  "help": ("path to a directory to dump metadata. If '--metadata-directory'"
+           " is not passed, the current working direcotry is used.")
+}
+
 
 def title_case_action_groups(parser):
   """Capitalize the first character of all words in the title of each action
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -34,7 +34,8 @@ from in_toto.common_args import (EXCLUDE_ARGS, EXCLUDE_KWARGS, BASE_PATH_ARGS,
     BASE_PATH_KWARGS, LSTRIP_PATHS_ARGS, LSTRIP_PATHS_KWARGS, KEY_ARGS,
     KEY_KWARGS, KEY_TYPE_KWARGS, KEY_TYPE_ARGS, GPG_ARGS, GPG_KWARGS,
     GPG_HOME_ARGS, GPG_HOME_KWARGS, VERBOSE_ARGS, VERBOSE_KWARGS, QUIET_ARGS,
-    QUIET_KWARGS, sort_action_groups, title_case_action_groups)
+    QUIET_KWARGS, METADATA_DIRECTORY_ARGS, METADATA_DIRECTORY_KWARGS,
+    sort_action_groups, title_case_action_groups)
 
 
 # Command line interfaces should use in_toto base logger (c.f. in_toto.log)
@@ -67,6 +68,14 @@ GPG home directory and record a file named 'foo' as material and product.
 
   {prog} start -n edit-foo --gpg -m path/to/foo
   {prog} stop -n edit-foo --gpg -p path/to/foo
+
+
+Create link metadata file signed with the private key loaded from 'key_file',
+record all files in the CWD as material and product, and dump finished link
+file to the target directory (on stop).
+
+  {prog} start -n edit-files -k path/to/key_file -m .
+  {prog} stop -d path/to/target/dir -n edit-files -k path/to/key_file -p .
 
 """.format(prog=parser.prog)
 
@@ -128,6 +137,9 @@ GPG home directory and record a file named 'foo' as material and product.
       " resulting link metadata's product section when running the 'stop'"
       " subcommand. Symlinks to files are followed."))
 
+  subparser_stop.add_argument(*METADATA_DIRECTORY_ARGS,
+      **METADATA_DIRECTORY_KWARGS)
+
   parser.add_argument('--version', action='version',
                       version='{} {}'.format(parser.prog, __version__))
 
@@ -187,7 +199,8 @@ def main():
           signing_key=key, gpg_keyid=gpg_keyid,
           gpg_use_default=gpg_use_default, gpg_home=args.gpg_home,
           exclude_patterns=args.exclude_patterns, base_path=args.base_path,
-          lstrip_paths=args.lstrip_paths)
+          lstrip_paths=args.lstrip_paths,
+          metadata_directory=args.metadata_directory)
 
   except Exception as e:
     LOG.error("(in-toto-record {0}) {1}: {2}"
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -482,6 +482,9 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     lstrip_paths (optional): A list of path prefixes used to left-strip
         artifact paths before storing them in the resulting link metadata.
 
+    metadata_directory (optional): A directory path to write the resulting link
+        metadata file to. Default destination is the current working directory.
+
   Raises:
     securesystemslib.exceptions.FormatError: Passed arguments are malformed.
 
@@ -493,7 +496,8 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
 
     securesystemslib.process.subprocess.TimeoutExpired: Link command times out.
 
-    IOError: Cannot write link metadata.
+    IOError, FileNotFoundError, NotADirectoryError, PermissionError:
+        Cannot write link metadata.
 
     securesystemslib.exceptions.CryptoError, \
             securesystemslib.exceptions.UnsupportedAlgorithmError:
@@ -654,7 +658,8 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
 
     securesystemslib.process.subprocess.TimeoutExpired: Link command times out.
 
-    IOError: Cannot write link metadata.
+    IOError, PermissionError:
+        Cannot write link metadata.
 
     securesystemslib.exceptions.CryptoError, \
             securesystemslib.exceptions.UnsupportedAlgorithmError:
@@ -781,6 +786,9 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
     lstrip_paths (optional): A list of path prefixes used to left-strip
         artifact paths before storing them in the resulting link metadata.
 
+    metadata_directory (optional): A directory path to write the resulting link
+        metadata file to. Default destination is the current working directory.
+
   Raises:
     securesystemslib.exceptions.FormatError: Passed arguments are malformed.
 
@@ -795,7 +803,8 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
 
     securesystemslib.process.subprocess.TimeoutExpired: Link command times out.
 
-    IOError: Cannot write link metadata.
+    IOError, FileNotFoundError, NotADirectoryError, PermissionError:
+        Cannot write link metadata.
 
     securesystemslib.exceptions.CryptoError, \
             securesystemslib.exceptions.UnsupportedAlgorithmError:
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -424,7 +424,7 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     record_streams=False, signing_key=None, gpg_keyid=None,
     gpg_use_default=False, gpg_home=None, exclude_patterns=None,
     base_path=None, compact_json=False, record_environment=False,
-    normalize_line_endings=False, lstrip_paths=None):
+    normalize_line_endings=False, lstrip_paths=None, metadata_directory=None):
   """Performs a supply chain step or inspection generating link metadata.
 
   Executes link_cmd_args, recording paths and hashes of files before and after
@@ -527,6 +527,9 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
   if base_path:
     securesystemslib.formats.PATH_SCHEMA.check_match(base_path)
 
+  if metadata_directory:
+    securesystemslib.formats.PATH_SCHEMA.check_match(metadata_directory)
+
   if material_list:
     LOG.info("Recording materials '{}'...".format(", ".join(material_list)))
 
@@ -578,6 +581,10 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
   if signature:
     signing_keyid = signature["keyid"]
     filename = FILENAME_FORMAT.format(step_name=name, keyid=signing_keyid)
+
+    if metadata_directory is not None:
+      filename = os.path.join(metadata_directory, filename)
+
     LOG.info("Storing link metadata to '{}'...".format(filename))
     link_metadata.dump(filename)
 
@@ -728,7 +735,7 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
 def in_toto_record_stop(step_name, product_list, signing_key=None,
     gpg_keyid=None, gpg_use_default=False, gpg_home=None,
     exclude_patterns=None, base_path=None, normalize_line_endings=False,
-    lstrip_paths=None):
+    lstrip_paths=None, metadata_directory=None):
   """Finalizes preliminary link metadata generated with in_toto_record_start.
 
   Loads preliminary link metadata file, verifies its signature, and records
@@ -824,6 +831,9 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
   if base_path:
     securesystemslib.formats.PATH_SCHEMA.check_match(base_path)
 
+  if metadata_directory:
+    securesystemslib.formats.PATH_SCHEMA.check_match(metadata_directory)
+
   # Load preliminary link file
   # If we have a signing key we can use the keyid to construct the name
   if signing_key:
@@ -904,6 +914,10 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
     link_metadata.sign_gpg(keyid, gpg_home)
 
   fn = FILENAME_FORMAT.format(step_name=step_name, keyid=keyid)
+
+  if metadata_directory is not None:
+    fn = os.path.join(metadata_directory, fn)
+
   LOG.info("Storing link metadata to '{}'...".format(fn))
   link_metadata.dump(fn)
 
--- a/tests/test_runlib.py
+++ b/tests/test_runlib.py
@@ -22,6 +22,8 @@ import os
 import unittest
 import shutil
 import tempfile
+import sys
+import stat
 
 import in_toto.settings
 import in_toto.exceptions
@@ -520,6 +522,35 @@ class TestInTotoRun(unittest.TestCase, TmpDirMixin):
         FILENAME_FORMAT.format(step_name=self.step_name, keyid=self.key["keyid"]))
     self.assertEqual(repr(link), repr(link_dump))
 
+  def test_in_toto_run_with_metadata_directory(self):
+    """Successfully run with metadata directory,
+     compare dumped link is equal to returned link"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp(dir=os.getcwd()))
+    link = in_toto_run(self.step_name, [self.test_artifact],
+        [self.test_artifact], ["python", "--version"], True, self.key,
+        metadata_directory=tmp_dir)
+    file_path = os.path.join(tmp_dir, FILENAME_FORMAT.format(step_name=self.step_name,
+        keyid=self.key["keyid"]))
+    link_dump = Metablock.load(file_path)
+    self.assertEqual(repr(link), repr(link_dump))
+
+  def test_in_toto_run_compare_with_and_without_metadata_directory(self):
+    """Successfully run with and without metadata directory,
+     compare the signed is equal"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp(dir=os.getcwd()))
+    in_toto_run(self.step_name, [self.test_artifact], [self.test_artifact],
+        ["python", "--version"], True, self.key, metadata_directory=tmp_dir)
+    file_path = os.path.join(tmp_dir, FILENAME_FORMAT.format(step_name=self.step_name,
+        keyid=self.key["keyid"]))
+    link_dump_with_md = Metablock.load(file_path)
+
+    in_toto_run(self.step_name, [self.test_artifact], [self.test_artifact],
+        ["python", "--version"], True, self.key)
+    link_dump_without_md = Metablock.load(
+        FILENAME_FORMAT.format(step_name=self.step_name, keyid=self.key["keyid"]))
+    self.assertEqual(repr(link_dump_with_md.signed),
+        repr(link_dump_without_md.signed))
+
   def test_in_toto_run_verify_recorded_artifacts(self):
     """Successfully run, verify properly recorded artifacts. """
     link = in_toto_run(self.step_name, [self.test_artifact],
@@ -572,6 +603,40 @@ class TestInTotoRun(unittest.TestCase, TmpDirMixin):
     with self.assertRaises(securesystemslib.exceptions.FormatError):
       in_toto_run(self.step_name, None, None,
           ["python", "--version"], True, self.key_pub)
+
+  def test_nonexistent_directory(self):
+    """Fail run, passed metadata_directory not exist. """
+    expected_error = IOError if sys.version_info < (3, 0) \
+        else FileNotFoundError
+    with self.assertRaises(expected_error):
+      in_toto_run(self.step_name, None, None, ["python", "--version"],
+          True, self.key, metadata_directory='nonexistentDir')
+
+  def test_not_a_directory(self):
+    """Fail run, passed metadata_directory is not a directory. """
+    fd, path = tempfile.mkstemp()
+    os.write(fd, b"hello in-toto")
+    os.close(fd)
+    # Windows will raise FileNotFoundError instead of NotADirectoryError
+    expected_error = IOError if sys.version_info < (3, 0) \
+        else (NotADirectoryError, FileNotFoundError)
+    with self.assertRaises(expected_error):
+      in_toto_run(self.step_name, None, None, ["python", "--version"],
+          True, self.key, metadata_directory=path)
+    os.remove(path)
+
+  @unittest.skipIf(os.name == 'nt', "chmod doesn't work properly on Windows")
+  def test_in_toto_read_only_metadata_directory(self):
+    """Fail run, passed metadata directory is read only"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp())
+    # make the directory read only
+    os.chmod(tmp_dir, stat.S_IREAD)
+    expected_error = IOError if sys.version_info < (3, 0) \
+        else PermissionError
+    with self.assertRaises(expected_error):
+      in_toto_run(self.step_name, None, None, ["python", "--version"],
+          True, self.key, metadata_directory=tmp_dir)
+    os.rmdir(tmp_dir)
 
 
 class TestInTotoRecordStart(unittest.TestCase, TmpDirMixin):
@@ -661,6 +726,23 @@ class TestInTotoRecordStop(unittest.TestCase, TmpDirMixin):
     self.assertEqual(list(link.signed.products.keys()), [self.test_product])
     os.remove(self.link_name)
 
+  def test_compare_metadata_with_and_without_metadata_directory(self):
+    """Test record stop with and without metadata directory,
+     compare the expected product"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp(dir=os.getcwd()))
+    in_toto_record_start(self.step_name, [], self.key)
+    in_toto_record_stop(self.step_name, [self.test_product], self.key,
+        metadata_directory=tmp_dir)
+    link_path = os.path.join(tmp_dir, self.link_name)
+    link_with_md = Metablock.load(link_path)
+
+    in_toto_record_start(self.step_name, [], self.key)
+    in_toto_record_stop(self.step_name, [self.test_product], self.key)
+    link_without_md = Metablock.load(self.link_name)
+    self.assertEqual(link_with_md.signed, link_without_md.signed)
+    os.remove(link_path)
+    os.remove(self.link_name)
+
   def test_create_metadata_with_expected_cwd(self):
     """Test record start/stop run, verify cwd. """
     in_toto_record_start(self.step_name, [], self.key, record_environment=True)
@@ -744,6 +826,43 @@ class TestInTotoRecordStop(unittest.TestCase, TmpDirMixin):
     finally:
       for path in paths:
         os.remove(path)
+
+  def test_nonexistent_directory(self):
+    """Test record stop with nonexistent metadata directory"""
+    expected_error = IOError if sys.version_info < (3, 0) \
+        else FileNotFoundError
+    with self.assertRaises(expected_error):
+      in_toto_record_start(self.step_name, [], self.key)
+      in_toto_record_stop(self.step_name, [], self.key,
+          metadata_directory='nonexistentDir')
+
+  def test_not_a_directory(self):
+    """Test record stop, passed metadata directory is not a dir"""
+    fd, path = tempfile.mkstemp()
+    os.write(fd, b"hello in-toto")
+    os.close(fd)
+    # Windows will raise FileNotFoundError instead of NotADirectoryError
+    expected_error = IOError if sys.version_info < (3, 0) \
+        else (NotADirectoryError, FileNotFoundError)
+    with self.assertRaises(expected_error):
+      in_toto_record_start(self.step_name, [], self.key)
+      in_toto_record_stop(self.step_name, [], self.key,
+          metadata_directory=path)
+    os.remove(path)
+
+  @unittest.skipIf(os.name == 'nt', "chmod doesn't work properly on Windows")
+  def test_read_only_metadata_directory(self):
+    """Test record stop with read only metadata directory"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp())
+    # make the directory read only
+    os.chmod(tmp_dir, stat.S_IREAD)
+    expected_error = IOError if sys.version_info < (3, 0) \
+       else PermissionError
+    with self.assertRaises(expected_error):
+      in_toto_record_start(self.step_name, [], self.key)
+      in_toto_record_stop(self.step_name, [], self.key,
+          metadata_directory=tmp_dir)
+    os.rmdir(tmp_dir)
 
 
 if __name__ == "__main__":
--- a/in_toto/in_toto_run.py
+++ b/in_toto/in_toto_run.py
@@ -32,7 +32,8 @@ from in_toto.common_args import (EXCLUDE_ARGS, EXCLUDE_KWARGS, BASE_PATH_ARGS,
     BASE_PATH_KWARGS, LSTRIP_PATHS_ARGS, LSTRIP_PATHS_KWARGS, KEY_ARGS,
     KEY_KWARGS, KEY_TYPE_KWARGS, KEY_TYPE_ARGS, GPG_ARGS, GPG_KWARGS,
     GPG_HOME_ARGS, GPG_HOME_KWARGS, VERBOSE_ARGS, VERBOSE_KWARGS, QUIET_ARGS,
-    QUIET_KWARGS, sort_action_groups, title_case_action_groups)
+    QUIET_KWARGS, METADATA_DIRECTORY_ARGS, METADATA_DIRECTORY_KWARGS,
+    sort_action_groups, title_case_action_groups)
 
 # Command line interfaces should use in_toto base logger (c.f. in_toto.log)
 LOG = logging.getLogger("in_toto")
@@ -141,6 +142,7 @@ e.g. 'document.pdf'.
   parser.add_argument(*EXCLUDE_ARGS, **EXCLUDE_KWARGS)
   parser.add_argument(*BASE_PATH_ARGS, **BASE_PATH_KWARGS)
   parser.add_argument(*LSTRIP_PATHS_ARGS, **LSTRIP_PATHS_KWARGS)
+  parser.add_argument(*METADATA_DIRECTORY_ARGS, **METADATA_DIRECTORY_KWARGS)
 
   verbosity_args = parser.add_mutually_exclusive_group(required=False)
   verbosity_args.add_argument(*VERBOSE_ARGS, **VERBOSE_KWARGS)
@@ -209,7 +211,8 @@ def main():
         record_streams=args.record_streams, signing_key=key,
         gpg_keyid=gpg_keyid, gpg_use_default=gpg_use_default,
         gpg_home=args.gpg_home, exclude_patterns=args.exclude_patterns,
-        base_path=args.base_path, lstrip_paths=args.lstrip_paths)
+        base_path=args.base_path, lstrip_paths=args.lstrip_paths,
+        metadata_directory=args.metadata_directory)
 
   except Exception as e:
     LOG.error("(in-toto-run) {0}: {1}".format(type(e).__name__, e))
--- a/tests/test_in_toto_record.py
+++ b/tests/test_in_toto_record.py
@@ -20,6 +20,8 @@
 
 import sys
 import unittest
+import tempfile
+import os
 
 if sys.version_info >= (3, 3):
   import unittest.mock as mock # pylint: disable=no-name-in-module,import-error
@@ -145,6 +147,12 @@ class TestInTotoRecordTool(CliTestCase, TmpDirMixin, GPGKeysMixin):
       self.assert_cli_sys_exit(["start"] + args, 0)
       self.assert_cli_sys_exit(["stop"] + args, 0)
 
+      # Start/stop sign with metadata directory
+      args = ["--step-name", "test9", "--key", self.rsa_key_path]
+      tmp_dir = os.path.realpath(tempfile.mkdtemp(dir=os.getcwd()))
+      metadata_directory_arg = ["--metadata-directory", tmp_dir]
+      self.assert_cli_sys_exit(["start"] + args, 0)
+      self.assert_cli_sys_exit(["stop"] + metadata_directory_arg + args, 0)
 
 
   def test_glob_no_unfinished_files(self):
--- a/tests/test_in_toto_run.py
+++ b/tests/test_in_toto_run.py
@@ -22,6 +22,7 @@ import os
 import sys
 import unittest
 import glob
+import tempfile
 
 # Use external backport 'mock' on versions under 3.3
 if sys.version_info >= (3, 3):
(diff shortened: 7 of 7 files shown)


in_toto/models/metadata.py
```diff
@@ -65,7 +65,7 @@ def __repr__(self):
       )
 
 
-  def dump(self, filename):
+  def dump(self, filename, path = ""):
```
adityasaky: A couple of things:

1. By convention, there should be no spaces around the `=` for keyword arguments. You'll also see mention of it in the lint report on Travis: https://travis-ci.org/github/in-toto/in-toto/jobs/691965668#L252
2. I don't think we need to take this approach. While the argument is called `filename`, it can accept a full path. While we can consider changing the name of the variable, we can definitely just pass in a full path here, meaning we may not need to make any changes in this file.

in_toto/models/metadata.py
```diff
@@ -82,6 +82,7 @@ def dump(self, filename):
       None.
 
     """
+    filename = path + filename
```
adityasaky: I'm leaving this comment here, but it may apply elsewhere if we decide to perform this operation elsewhere and pass in a full path to `dump`. I suggest using `os.path.join` to generate the full path, as this also uses the correct separators, as well as adding a trailing separator that may have been missed.

We should also check if the value passed for `metadata_directory` is a directory, and throw an error if not. I also recommend catching any permission based errors - what if the user doesn't have the permission to write to the specified directory?

in_toto/runlib.py
```diff
@@ -571,7 +571,9 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     signing_keyid = signature["keyid"]
     filename = FILENAME_FORMAT.format(step_name=name, keyid=signing_keyid)
     LOG.info("Storing link metadata to '{}'...".format(filename))
-    link_metadata.dump(filename)
+    if metadata_directory == None:
```
adityasaky: In Python, we use `is` for this comparison (also caught by the linter FYI). `if metadata_directory is None:` basically. However, the `./` doesn't actually provide us with anything, so I recommend making this condition `if metadata_directory is not None:` and performing the directory check and join within to `filename`. If it indeed is `None`, we don't need to bother, and things will continue as they do now.

adityasaky (changes requested):
Thanks for the PR, @SolidifiedRay. This is a solid start! I have some comments on the changes.

For the next steps, you'd need to write unit tests for this feature addition. Also, `in-toto-record` should get the same flag.

in_toto/in_toto_record.py
```diff
@@ -128,6 +128,14 @@ def create_parser():
       " resulting link metadata's product section when running the 'stop'"
       " subcommand. Symlinks to files are followed."))
 
+  subparser_start.add_argument("-d", "--metadata-directory", type=str, required=False, 
+    metavar="<directory>", help=(
+      "specify where to write the generated metadata to."))
```
lukpueh: FYI, there's an `in_toto.common_args` module that we use to define argparse arguments that are used multiple times only once. I think `--metadata-directory` qualifies for it. :)

in_toto/runlib.py
```diff
@@ -570,6 +580,14 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
   if signature:
     signing_keyid = signature["keyid"]
     filename = FILENAME_FORMAT.format(step_name=name, keyid=signing_keyid)
+    
+    if metadata_directory is not None:
+      if not os.path.isdir(metadata_directory):
+        raise ValueError("--metadata-directory value is not a directory")
```
lukpueh: Two general notes:
- a caller of this function might not know about the cli argument `--metadata-directory`. it's better to use the name used in this scope that is `metadata_directory`.
- it's more helpful user feedback if you include the value of the variable

in_toto/runlib.py
```diff
@@ -570,6 +580,14 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
...
+      if not os.access(metadata_directory, os.W_OK):
+        raise PermissionError("Cannot write to {}".format(metadata_directory))
```
lukpueh: I suggest to take an [EAFP](https://docs.python.org/3/glossary.html#term-eafp)-approach here, it's less code and more reliable. You can still enhance the error message by catching and re-raising the exception with a better message. If you do so I suggest to do it once in the dump function.

SolidifiedRay: I notice that both _in_toto_run.py_ and _in_toto_record.py_ have "try" and "except", which can catch the directory value error and permission error. So I decided to simply remove these two checks without changing the dump function.

adityasaky:
Thanks for taking a look, @lukpueh :)

in_toto/runlib.py
```diff
@@ -818,11 +842,21 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
...
+
   # FIXME: Currently there is no way to know the default GPG key's keyid and
   # so we glob for preliminary link files
+  # FIXME: UNFINISHED_FILENAME_FORMAT_GLOB_DIR has a problem with window system
   else:
     unfinished_fn_glob = UNFINISHED_FILENAME_FORMAT_GLOB.format(
         step_name=step_name, pattern="*")
+
+    if metadata_directory is not None:
+      unfinished_fn_glob = UNFINISHED_FILENAME_FORMAT_GLOB_DIR.format(
```
adityasaky: I think this needs to be imported. https://travis-ci.org/github/in-toto/in-toto/jobs/693631357#L259

in_toto/models/link.py
```diff
@@ -27,6 +27,8 @@
 FILENAME_FORMAT_SHORT = "{step_name}.link"
 UNFINISHED_FILENAME_FORMAT = ".{step_name}.{keyid:.8}.link-unfinished"
 UNFINISHED_FILENAME_FORMAT_GLOB = ".{step_name}.{pattern}.link-unfinished"
+UNFINISHED_FILENAME_FORMAT_GLOB_DIR = \
+  ".{metadata_directory}/.{step_name}.{pattern}.link-unfinished"
```
adityasaky: I think the first `.` needs to be dropped?

Also, as we discussed, I think we need a better way to do it. For one thing, we're using `/` as the separator and we need to check how that'd work on a Windows machine. @lukpueh do you have any thoughts? `glob.glob` does accept the separator.

adityasaky (changes requested):

in_toto/models/link.py
SolidifiedRay: For the window machine, my first thought is to add another format and check both formats when signing with gpg, but there should be a better method.

in_toto/common_args.py
```diff
@@ -122,6 +122,14 @@
   "help": "suppress all output"
 }
 
+METADATA_DIRECTORY_ARGS = ["-d", "--metadata-directory"]
+METADATA_DIRECTORY_KWARGS = {
+  "required": False,
+  "type":str,
```
adityasaky: nit: space before `str`

in_toto/in_toto_record.py
```diff
@@ -128,6 +129,12 @@ def create_parser():
       " resulting link metadata's product section when running the 'stop'"
       " subcommand. Symlinks to files are followed."))
 
+  subparser_start.add_argument(*METADATA_DIRECTORY_ARGS,
+    **METADATA_DIRECTORY_KWARGS)
```
adityasaky: style change: double indentation for line continuation

> double-indentation (4 spaces) is prefered for line continuation, to distinguish it from functional indentation

https://github.com/secure-systems-lab/code-style-guidelines#maximum-line-length

in_toto/in_toto_record.py
```diff
@@ -128,6 +129,12 @@ def create_parser():
...
+
+  subparser_stop.add_argument(*METADATA_DIRECTORY_ARGS,
+    **METADATA_DIRECTORY_KWARGS)
```
adityasaky: double indentation again

in_toto/runlib.py
```diff
@@ -485,6 +487,9 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     lstrip_paths: (optional)
             If a prefix path is passed, the prefix is left stripped from
             the path of every artifact that contains the prefix.
+    metadata_directory: (optional)
+            If a metadata directory is passed, write the generated metadata
```
adityasaky: nit: grammar (?)

How about:

"If a directory path is passed, link metadata is written to that location." or something similar?

in_toto/runlib.py
```diff
@@ -497,6 +502,12 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     securesystemslib.gpg.exceptions.CommandError:
         If gpg is used for signing and the command exits with a non-zero code.
 
+
+    ValueError if a metadata_directory is passed and is not a directory
```
adityasaky: Can we double check if this is a `ValueError` or a `FileNotFoundError`? I see `dump` results in the latter if the path doesn't exist, but I'm wondering if something else is causing a `ValueError`. I may be wrong here though.

```
>>> from in_toto.models.link import Link
>>> from in_toto.models.metadata import Metablock
>>> m = Metablock(signed=Link())
>>> m.dump('dne/ayylmao.link')
  File "/home/saky/Work/secure-systems-lab/in-toto/in_toto/models/metadata.py", line 85, in dump
    with open(filename, "wb") as fp:
FileNotFoundError: [Errno 2] No such file or directory: 'dne/ayylmao.link'
```

in_toto/runlib.py
```diff
@@ -625,6 +641,9 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
     lstrip_paths: (optional)
             If a prefix path is passed, the prefix is left stripped from
             the path of every artifact that contains the prefix.
+    metadata_directory: (optional)
+            If a metadata directory is passed, write the generated metadata
```
adityasaky: same as above, though we don't necessarily have to use the snippet I included.

in_toto/runlib.py
```diff
@@ -818,11 +843,21 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
...
         keyid=signing_key["keyid"])
 
+    if metadata_directory is not None:
+      unfinished_fn = os.path.join(metadata_directory, unfinished_fn)
+
   # FIXME: Currently there is no way to know the default GPG key's keyid and
   # so we glob for preliminary link files
+  # FIXME: UNFINISHED_FILENAME_FORMAT_GLOB_DIR has a problem with window system
   else:
     unfinished_fn_glob = UNFINISHED_FILENAME_FORMAT_GLOB.format(
```
adityasaky: I think it may be nicer to wrap this as part of the `if metadata_directory` condition below, instead of executing both in case a value is passed. It also slightly improves readability IMO.

This is assuming we stick with `UNFINISHED_FILENAME_FORMAT_GLOB_DIR` of course.

in_toto/models/link.py
```diff
@@ -27,6 +27,8 @@
 FILENAME_FORMAT_SHORT = "{step_name}.link"
 UNFINISHED_FILENAME_FORMAT = ".{step_name}.{keyid:.8}.link-unfinished"
 UNFINISHED_FILENAME_FORMAT_GLOB = ".{step_name}.{pattern}.link-unfinished"
+UNFINISHED_FILENAME_FORMAT_GLOB_DIR = \
+  "{metadata_directory}/.{step_name}.{pattern}.link-unfinished"
```
adityasaky: I'm still not convinced by this.

- I think `glob` is smart about `//` in the overall pattern (this applies if the `metadata_directory` value includes a trailing `/`) but it's probably best avoided?
- We'll probably not catch permission issues.

Imagine a scenario where the metadata directory can be written to, but not read from, for whatever reason. `glob.glob` will search for the pattern with the `metadata_directory` included, but won't detect a lack of permission (we're not going to see `PermissionError`). It'll just return an empty list, and the user will get a message saying they didn't run `start`.

This all makes me wonder if we should even use `metadata_directory` for `in-toto-record start`. We continue the current policy of writing the unfinished link to the current working directory, and write only final metadata to the specified target in `in-toto-record stop`. Thoughts @lukpueh?

in_toto/runlib.py
```diff
@@ -818,11 +843,21 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
     unfinished_fn = UNFINISHED_FILENAME_FORMAT.format(step_name=step_name,
         keyid=signing_key["keyid"])
 
+    if metadata_directory is not None:
+      unfinished_fn = os.path.join(metadata_directory, unfinished_fn)
+
   # FIXME: Currently there is no way to know the default GPG key's keyid and
   # so we glob for preliminary link files
+  # FIXME: UNFINISHED_FILENAME_FORMAT_GLOB_DIR has a problem with window system
```
adityasaky: Did you get a chance to confirm this? :)

adityasaky (changes requested):
This is mostly looking quite good, in my opinion, @SolidifiedRay! I think, some nits apart, this PR needs some tests and a proper fix for the `in-toto-record` pattern, and we should be good to go. :smile:

in_toto/runlib.py
adityasaky: I see I misunderstood you slightly here. If I pass in a file that exists but isn't a directory, I receive a `NotADirectoryError`, which I don't see in the Python 2 documentation. I think there are a few different possibilities here.

adityasaky: On Python 2, I'm receiving an `IOError`. I think it's worth being explicit here. :slightly_smiling_face: 

```
>>> from in_toto.models.link import Link
>>> from in_toto.models.metadata import Metablock
>>> import os
>>> os.listdir('.')
['env', 'hello', 'final_products']
>>> Metablock(signed=Link()).dump(os.path.join('hello', 'world.link'))
  File "/home/saky/tmp/in-toto/test/env/lib/python2.7/site-packages/in_toto/models/metadata.py", line 85, in dump
    with open(filename, "wb") as fp:
IOError: [Errno 20] Not a directory: 'hello/world.link'
```

SolidifiedRay: Oh, this comment is from the previous commit when we have the check for os.isdir() and I forget to change it. 
Do you suggest listing all the possible Exceptions here?

in_toto/runlib.py
SolidifiedRay: I found an IBM [QA page](https://www.ibm.com/support/pages/windows-forward-slash-and-backslash-characters-separators) which say that "windows can use the windows directory separator (\) and the Unix directory separator(/)" However, I didn't find the Microsoft documentation that confirms this.

in_toto/models/link.py
SolidifiedRay: I also agree with you about the idea that we may not need to use `metadata_director` for the `in-toto-record start`. Since we generate the final link when we run `in-toto-record stop`, maybe it will be more reasonable to decide the path to store the link at that time?

> I'm still not convinced by this.
> 
> * I think `glob` is smart about `//` in the overall pattern (this applies if the `metadata_directory` value includes a trailing `/`) but it's probably best avoided?
> * We'll probably not catch permission issues.
> 
> Imagine a scenario where the metadata directory can be written to, but not read from, for whatever reason. `glob.glob` will search for the pattern with the `metadata_directory` included, but won't detect a lack of permission (we're not going to see `PermissionError`). It'll just return an empty list, and the user will get a message saying they didn't run `start`.
> 
> This all makes me wonder if we should even use `metadata_directory` for `in-toto-record start`. We continue the current policy of writing the unfinished link to the current working directory, and write only final metadata to the specified target in `in-toto-record stop`. Thoughts @lukpueh?

in_toto/runlib.py
SolidifiedRay: > Did you get a chance to confirm this? :)

I found an IBM [QA page](https://www.ibm.com/support/pages/windows-forward-slash-and-backslash-characters-separators) which says that "windows can use the windows directory separator(\) and the Unix directory separator(/)" However, I didn't find the Microsoft documentation that confirms this.

in_toto/models/link.py
SolidifiedRay: I also agree with you about the idea that we may not need to use `metadata_director `for the `in-toto-record start`. Since we generate the final link when we run `in-toto-record stop`, maybe it will be more reasonable to decide the path to store the link at that time?

in_toto/runlib.py
adityasaky: To confirm, we'd be replacing the current `ValueError` listing with `IOError` and `NotADirectoryError` right? I think that's okay. We should specify Python 2 / 3 distinction of course. And a `FileNotFoundError` in Python 3 for when the path doesn't exist. Python 2 raises `IOError` again for that.

To summarize:
- `IOError` on Python 2 if the path doesn't exist or isn't a directory
- `FileNotFoundError` on Python 3 if path doesn't exist
- `NotADirectoryError` on Python 3 if path isn't a directory

We should of course verify these with the unit tests. :slightly_smiling_face:

in_toto/models/link.py
adityasaky: Yep, I'm leaning towards that. Let's get some opinions. :smile:

lukpueh:
Thanks for the great work, @SolidifiedRay and for your help, @adityasaky! 

Regarding your [questions the CNCF TUF slack channel](https://cloud-native.slack.com/archives/CM46K2VT2/p1591307176223100):
I think either one of storing the temporary link metadata from `in-toto-record` in `cwd` or in `metadata_directory` is okay, as long as it is documented in the function docstring. In most cases the user won't care where the temporary file is stored. The only scenario I can think of, is that a user doesn't have the required permissions in cwd (but does in `metadata_directory`). Although that seems unlikely. So I'm really fine either way.

If you do store the preliminary link file in `metadata_directory`, why don't you also just use join like you do without the glob pattern? I.e.:
```python
unfinished_fn_glob = os.path.join(metadata_directory, unfinished_fn_glob) 
```

That should solve the trailing slash and directory separator issue.

Also thanks a lot for pointing out the permission error masking, I didn't know that glob had that behavior. At any rate this issue exists with our without `metadata_directory`. We could include that bit of information in the error message for convenience. Looking at the error message I see that it talks about cwd, so if you glob in `metadata_directory`, don't forget to update that.

tests/test_runlib.py
```diff
@@ -447,6 +447,7 @@ def setUpClass(self):
     generate_and_write_rsa_keypair(self.key_path)
     self.key = prompt_import_rsa_key_from_file(self.key_path)
     self.key_pub = prompt_import_rsa_key_from_file(self.key_path + ".pub")
+    self.metadata_directory = './'
```
adityasaky: Let's use `tempfile` to create a directory for the tests and write to it. We can then also theoretically play around with its permission bits.

tests/test_runlib.py
```diff
@@ -627,6 +645,8 @@ def setUpClass(self):
     self.link_name_unfinished = UNFINISHED_FILENAME_FORMAT.format(
         step_name=self.step_name, keyid=self.key["keyid"])
 
+    self.metadata_directory = "./"
```
adityasaky: same here

tests/test_runlib.py
```diff
@@ -642,6 +662,16 @@ def test_create_metadata_with_expected_product(self):
...
     os.remove(self.link_name)
 
+  def test_create_metadata_with_expected_product_in_target_directory(self):
+    """Test record stop records expected product in target directory """
+    in_toto_record_start(self.step_name, [], self.key)
+    in_toto_record_stop(self.step_name, [self.test_product], self.key,
+        metadata_directory=self.metadata_directory)
+    link_path = os.path.join(self.metadata_directory, self.link_name)
+    link = Metablock.load(link_path)
+    self.assertEqual(list(link.signed.products.keys()), [self.test_product])
```
adityasaky: I'm wondering if we want to store the link with and without the `metadata_directory` flag and compare them in some way.

tests/test_runlib.py
```diff
@@ -642,6 +662,16 @@ def test_create_metadata_with_expected_product(self):
...
+    os.remove(self.link_name)
```
adityasaky, line 744: +1 on clean up! I think this will be easier with a teardown method perhaps, and if we're using a newly created temporary directory.

tests/test_runlib.py
```diff
@@ -726,6 +756,13 @@ def test_normalize_line_endings(self):
       for path in paths:
         os.remove(path)
 
+  def test_wrong_metadata_directory(self):
+    """Test record stop with wrong metadata directory"""
+    with self.assertRaises(IOError):
```
adityasaky: Remind me, this is dependent on the python version, right? If so, we should add one with the "other" exception as well.

adityasaky (changes requested):
Great starting point for the tests, @SolidifiedRay! I think we can move forward from this point and add some more test cases for different scenarios. :smile:

tests/test_runlib.py
SolidifiedRay: I am not sure about using a teardown method. Other test functions in this class all use `os.remove()` and I just follow this pattern.

adityasaky: That makes sense, really this only applies if we're creating a class for `metadata_directory` specific tests. Depends on how many we have, and how many can use the same directory that we create (so that we can do it once for the class and reuse). I think we can work out the rest of it and reorganize?

lukpueh: IMO you don't need to wrap this in a try/finally block, it makes it a tiny bit harder to read and the whole test case is run in a temporary directory, which gets removed on tearDownClass anyway (see TmpDirMixin).

in_toto/runlib.py
```diff
@@ -497,6 +505,15 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     securesystemslib.gpg.exceptions.CommandError:
         If gpg is used for signing and the command exits with a non-zero code.
 
+    IOError on Python 2 if the path doesn't exist or isn't a directory
+
+    FileNotFoundError on Python 3 if path doesn't exist
+
+    NotADirectoryError on Python 3 if path isn't a directory
+
+    PermissionError if a metadata_directory is passed and we don't have the
```
adityasaky: This should clarify it's Python 3 only. I see you're already testing for an `IOError` in the tests. :slightly_smiling_face:

tests/test_runlib.py
```diff
@@ -501,6 +503,47 @@ def test_in_toto_run_compare_dumped_with_returned_link(self):
...
+      os.rmdir(tmp_dir)
+
+  def test_in_toto_run_compare_with_and_without_metadata_directory(self):
+    """Successfully run with and without metadata directory,
+     compare the signed is equal"""
+    try:
+      tmp_dir = tempfile.mkdtemp()
+      in_toto_run(self.step_name, [self.test_artifact], [self.test_artifact],
+          ["python", "--version"], True, self.key, metadata_directory=tmp_dir)
+      file_path = os.path.join(tmp_dir, self.step_name)
```
adityasaky: This works, no doubt about that, but let's avoid confusion by keeping `FILENAME_FORMAT.format()` with just the step name, and passing that as a whole to `os.path.join`.

```
file_path = os.path.join(tmp_dir, FILENAME_FORMAT.format(step_name=self.step_name, keyid=self.key["keyid"]))
link_dump_with_md = Metablock.load(file_path)
```

What do you think?

tests/test_runlib.py
```diff
@@ -642,6 +720,45 @@ def test_create_metadata_with_expected_product(self):
...
+  def test_create_metadata_with_expected_product_in_target_directory(self):
+    """Test record stop records expected product in target directory """
+    try:
+      tmp_dir = tempfile.mkdtemp()
+      in_toto_record_start(self.step_name, [], self.key)
+      in_toto_record_stop(self.step_name, [self.test_product], self.key,
+          metadata_directory=tmp_dir)
+      link_path = os.path.join(tmp_dir, self.link_name)
+      link = Metablock.load(link_path)
+      self.assertEqual(list(link.signed.products.keys()), [self.test_product])
```
adityasaky: I'm not sure we're very concerned about the products here.

tests/test_runlib.py
```diff
@@ -642,6 +720,45 @@ def test_create_metadata_with_expected_product(self):
...
+      in_toto_record_start(self.step_name, [], self.key)
+      in_toto_record_stop(self.step_name, [self.test_product], self.key,
+          metadata_directory=tmp_dir)
+      link_path = os.path.join(tmp_dir, self.link_name)
+      link_with_md = Metablock.load(link_path)
+
+      in_toto_record_start(self.step_name, [], self.key)
+      in_toto_record_stop(self.step_name, [self.test_product], self.key)
+      link_without_md = Metablock.load(self.link_name)
+      self.assertEqual(list(link_with_md.signed.products.keys()),
```
adityasaky: Why not just `link_with_md.signed` and `link_without_md.signed`?

in_toto/runlib.py
```diff
@@ -785,6 +813,15 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
...
```
adityasaky: Python 3 mention here as well

in_toto/runlib.py
```diff
@@ -823,6 +860,7 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
   else:
     unfinished_fn_glob = UNFINISHED_FILENAME_FORMAT_GLOB.format(
         step_name=step_name, pattern="*")
+
```
adityasaky: Stray line that probably snuck in during one of the prior discussions about where to put the unfinished link.

tests/test_runlib.py
```diff
@@ -22,6 +22,8 @@
 import unittest
 import shutil
 import tempfile
+import sys
+from stat import S_IREAD
```
adityasaky: Let's `import stat` here and use `stat.S_IREAD` in the `chmod` call.

tests/test_runlib.py
```diff
@@ -501,6 +503,47 @@ def test_in_toto_run_compare_dumped_with_returned_link(self):
...
+      link = in_toto_run(self.step_name, [self.test_artifact],
+          [self.test_artifact], ["python", "--version"], True, self.key,
+          metadata_directory=tmp_dir)
+      file_path = os.path.join(tmp_dir, self.step_name)
+      link_dump = Metablock.load(
+          FILENAME_FORMAT.format(step_name=file_path, keyid=self.key["keyid"]))
+      self.assertEqual(repr(link), repr(link_dump))
+
+    #clean up
+    finally:
```
adityasaky: I still think we should reorganize these, perhaps all under a class of tests for metadata directory, and use `tearDown` / `tearDownClass` for clean up. We shouldn't need `try` here.

adityasaky (changes requested):
This mostly looks quite good, and I think you've covered almost all the cases, @SolidifiedRay! Some minor nits, and reorganization, and we should be good to go.

Oh, and please take a look at why AppVeyor is failing. :)

tests/test_runlib.py
adityasaky: Though, I think it'd be a first to have a class that's not for a specific method defined in `runlib`, so we may not want to go that direction either.

tests/test_runlib.py
SolidifiedRay: I am thinking that:
If we already compare the dumped link with the metadata directory is equal to the dumped link without the metadata directory, we don't need this function here to check for the products.
Or maybe we should check something other than products?
What do you think?

tests/test_runlib.py
```diff
@@ -726,6 +814,40 @@ def test_normalize_line_endings(self):
...
+    if sys.version_info < (3, 0):
+      with self.assertRaises(IOError):
+        in_toto_record_start(self.step_name, [], self.key)
+        in_toto_record_stop(self.step_name, [], self.key,
+            metadata_directory='this-is-not-a-dir')
+    else:
+      with self.assertRaises(FileNotFoundError):
+        in_toto_record_start(self.step_name, [], self.key)
+        in_toto_record_stop(self.step_name, [], self.key,
+            metadata_directory='this-is-not-a-dir')
```
lukpueh: I suggest to just put the things that differ in the if/else clause, e.g.:
```python
expected_error = IOError if  sys.version_info < (3, 0) else FileNotFoundError
```
Same applies to `test_read_only_metadata_directory`.

tests/test_runlib.py
```diff
@@ -726,6 +814,40 @@ def test_normalize_line_endings(self):
...
+      with self.assertRaises(NotADirectoryError):
+        path = tempfile.mkstemp(dir=os.getcwd())[1]
+        in_toto_record_start(self.step_name, [], self.key)
+        in_toto_record_stop(self.step_name, [], self.key,
+            metadata_directory=path)
```
lukpueh: Since these lines are only run on py3, you could put them in separate test function and add a decorator to skip on py2.

tests/test_runlib.py
```diff
@@ -642,6 +707,34 @@ def test_create_metadata_with_expected_product(self):
...
 
+  def test_create_metadata_with_expected_product_in_target_directory(self):
+    """Test record stop records expected product in target directory """
+    tmp_dir = os.path.realpath(tempfile.mkdtemp(dir=os.getcwd()))
+    in_toto_record_start(self.step_name, [], self.key)
+    in_toto_record_stop(self.step_name, [self.test_product], self.key,
+        metadata_directory=tmp_dir)
+    link_path = os.path.join(tmp_dir, self.link_name)
+    link = Metablock.load(link_path)
+    self.assertEqual(list(link.signed.products.keys()), [self.test_product])
```
adityasaky: I'm still not sure I understand this test in the context of `metadata_directory`. Could you go over the reasoning, @SolidifiedRay? :slightly_smiling_face:

adityasaky: I'd have expected it to be similar to `test_in_toto_run_with_metadata_directory`.

tests/test_runlib.py
```diff
@@ -726,6 +819,43 @@ def test_normalize_line_endings(self):
...
+          metadata_directory='nonexistentDir')
+
+  @unittest.skipIf(sys.version_info < (3, 0),
+      "NotADirectoryError is python3 only")
+  def test_not_a_directory(self):
+    """Test record stop, passed metadata directory is not a dir"""
+    fd, path = tempfile.mkstemp()
+    os.write(fd, b"hello in-toto")
+    os.close(fd)
+    with self.assertRaises((NotADirectoryError, FileNotFoundError)):
```
adityasaky: Just to confirm and keep things straight in my head; this is because on Windows we're seeing `FileNotFoundError` instead of `NotADirectoryError` correct? Can we follow that down and see where the difference is? Is it `os.path.join`?

SolidifiedRay: Yes, this is because on Windows's we're seeing `FileNotFoundError` instead of `NotADirectoryError`. But I don't understand what do you mean by `os.path.join`. Could you explain a bit more? @adityasaky

adityasaky: I'm curious exactly which operation is raising this error (either way). As I recall, the value of `metadata_directory` is used in `os.path.join` so is that a difference in implementation there, between Windows and the rest :tm:?

tests/test_runlib.py
SolidifiedRay: For `test_in_toto_run_with_metadata_directory`, `in_toto_run` itself will return a link, so I can compare that link with the `link_dump`. But I think `in_toto_record_stop` will not return a link, so I am not sure what to compare. 
For example, I can do `link = in_toto_run()` and compare this `link` with `link_dump`.
However, I am not able to do something like `link = in_toto_record_stop()`
Do you have any suggestions for this? @adityasaky

tests/test_runlib.py
SolidifiedRay: Yes, you are right. @adityasaky  
I change ` filename = os.path.join(metadata_directory , filename)` to `filename = metadata_directory + filename`, and this will raise `NotADirectoryError`.

adityasaky: This is very interesting. I'd have more expected it to be an issue with `open()` rather than `join`, because I didn't think `join` cared about what contents it was stringing together. Can you dump the difference in the `filename` strings when you obtain it through `os.path.join` and `metadata_directory + filename`?

SolidifiedRay: Do you mean dump the `filename` strings to file and compare them? If so, the result `filename` strings look the same to me.

tests/test_runlib.py
adityasaky: Ah I see. I think the test which runs `in-toto-record` twice, with and without `metadata_directory`, suffices. Maybe we can dump this one.

tests/test_runlib.py
adityasaky: I see. For now, let's do a `skipIf` and move on. Perhaps you can open an issue for this particular case? :)

adityasaky: Correction: I had a brianfart, let's `skipIf` the `PermissionError` test, and retain both exceptions here (with an explanation).

tests/test_runlib.py
```diff
@@ -554,6 +585,44 @@ def test_in_toto_wrong_key(self):
...
+
+  @unittest.skipIf(sys.version_info < (3, 0),
+      "NotADirectoryError is python3 only")
+  def test_not_a_directory(self):
+    """Fail run, passed metadata_directory is not a directory. """
+    fd, path = tempfile.mkstemp()
+    os.write(fd, b"hello in-toto")
+    os.close(fd)
+    # Windows will raise FileNotFoundError instead of NotADirectoryError
+    # The reason may be os.path.join is implemented differetnly in
```
```suggestion
```

tests/test_runlib.py
```diff
@@ -554,6 +585,44 @@ def test_in_toto_wrong_key(self):
...
+    # in different os
```
```suggestion
```

tests/test_runlib.py
```diff
@@ -554,6 +585,44 @@ def test_in_toto_wrong_key(self):
...
+      in_toto_run(self.step_name, None, None, ["python", "--version"],
+          True, self.key, metadata_directory=path)
+    os.remove(path)
+
+  @unittest.skipIf(os.name == 'nt', "chmod doesn't work properly on Windows")
+  def test_in_toto_read_only_metadata_directory(self):
+    """Fail run, passed metadata directory is read only"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp())
+    # make the directory read only
+    # FIXME: this method can't make dir read-only in the
```
```suggestion
```

Let's open an issue instead.

tests/test_runlib.py
```diff
@@ -554,6 +585,44 @@ def test_in_toto_wrong_key(self):
...
+    # Windows machine
```
```suggestion
```

tests/test_runlib.py
```diff
@@ -554,6 +585,44 @@ def test_in_toto_wrong_key(self):
...
 
+  def test_nonexistent_directory(self):
+    """Fail run, passed metadata_directory not exist. """
+    expected_error = IOError if sys.version_info < (3, 0) \
+        else FileNotFoundError
+    with self.assertRaises(expected_error):
+      in_toto_run(self.step_name, None, None, ["python", "--version"],
+          True, self.key, metadata_directory='nonexistentDir')
+
+  @unittest.skipIf(sys.version_info < (3, 0),
```
adityasaky: This makes some sense to me, but are we never checking for `IOError` on Python 2?

tests/test_runlib.py
```diff
@@ -726,6 +812,47 @@ def test_normalize_line_endings(self):
...
+      in_toto_record_stop(self.step_name, [], self.key,
+          metadata_directory=path)
+    os.remove(path)
+
+  @unittest.skipIf(os.name == 'nt', "chmod doesn't work properly on Windows")
+  def test_read_only_metadata_directory(self):
+    """Test record stop with read only metadata directory"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp())
+    # make the directory read only
+    # FIXME: this method can't make dir read-only in the
```
```suggestion
```

can be a part of the issue from before

tests/test_runlib.py
```diff
@@ -726,6 +812,47 @@ def test_normalize_line_endings(self):
...
+    # Windows machine
```
```suggestion
```

tests/test_runlib.py
```diff
@@ -554,6 +585,44 @@ def test_in_toto_wrong_key(self):
...
+
+  @unittest.skipIf(os.name == 'nt', "chmod doesn't work properly on Windows")
+  def test_in_toto_read_only_metadata_directory(self):
+    """Fail run, passed metadata directory is read only"""
+    tmp_dir = os.path.realpath(tempfile.mkdtemp())
+    # make the directory read only
+    # FIXME: this method can't make dir read-only in the
+    # Windows machine
+    os.chmod(tmp_dir, stat.S_IREAD)
+    expected_error = IOError if  sys.version_info < (3, 0) \
```
```suggestion
    expected_error = IOError if sys.version_info < (3, 0) \
```

in_toto/in_toto_record.py
```diff
@@ -68,6 +69,14 @@ def create_parser():
   {prog} start -n edit-foo --gpg -m path/to/foo
   {prog} stop -n edit-foo --gpg -p path/to/foo
 
+
+Create link metadata file signed with the private key loaded from 'key_file,
```
adityasaky:
```suggestion
Create link metadata file signed with the private key loaded from 'key_file',
```

in_toto/in_toto_record.py
```diff
@@ -68,6 +69,14 @@ def create_parser():
...
+recording, record all files in the CWD as material and product, and dump
```
adityasaky:
```suggestion
record all files in the CWD as material and product, and dump the
```

in_toto/in_toto_record.py
```diff
@@ -68,6 +69,14 @@ def create_parser():
...
+the finished link file to the target directory (on stop).
```
adityasaky:
```suggestion
finished link file to the target directory (on stop).
```

in_toto/runlib.py
```diff
@@ -435,6 +436,10 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     The link file is dumped to `link.FILENAME_FORMAT` using the signing key's
     keyid.
 
+    The link file is dumped to the target file if metadata_directory
```
adityasaky:
```suggestion
    The link file is dumped to the target directory if metadata_directory
```

in_toto/runlib.py
```diff
@@ -435,6 +436,10 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
...
+    parameter is passed. Otherwise, it is dumped to the current working
+    directory
```
adityasaky:
```suggestion
    directory.
```

in_toto/runlib.py
```diff
@@ -713,14 +736,18 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
...
-    lstrip_paths=None):
+    lstrip_paths=None, metadata_directory=None):
   """
   <Purpose>
     Finishes creating link metadata for a multi-part in-toto step.
     Loads unfinished link metadata file from disk, verifies
     that the file was signed with either the passed signing key, a gpg key
     identified by the passed gpg_keyid or the default gpg key.
 
+    The link file is dumped to the target file if metadata_directory
```
adityasaky:
```suggestion
    The link file is dumped to the target directory if metadata_directory
```

in_toto/runlib.py
```diff
@@ -713,14 +736,18 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
...
+    parameter is passed. Otherwise, it is dumped to the current working
+    directory
```
adityasaky:
```suggestion
    directory.
```

in_toto/runlib.py
```diff
@@ -760,6 +787,9 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
     lstrip_paths: (optional)
             If a prefix path is passed, the prefix is left stripped from
             the path of every artifact that contains the prefix.
+    metadata_directory: (optional)
+            If a metadata directory is passed, write the generated metadata
```
adityasaky:
```suggestion
            If a metadata directory is passed, the generated metadata
```

in_toto/runlib.py
```diff
@@ -760,6 +787,9 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
...
+            to this specific directory
```
adityasaky:
```suggestion
            is written to this specific directory.
```

adityasaky (approved):
LGTM, thanks @SolidifiedRay!

cc @lukpueh @SantiagoTorres for any thoughts and potentially merging this in.

- Also, perhaps worth worth cleaning up commits?
- And worth exploring something more granular than `os.name` for the `PermissionError` test?

in_toto/common_args.py
```diff
@@ -122,6 +122,14 @@
   "help": "suppress all output"
 }
 
+METADATA_DIRECTORY_ARGS = ["-d", "--metadata-directory"]
+METADATA_DIRECTORY_KWARGS = {
+  "required": False,
+  "type": str,
+  "metavar": "<directory>",
+  "help": "specify where to write the generated metadata to."
```
lukpueh: Nit: Can you please start the help text with a noun so that it is consistent with the existing help texts? E.g.:
"path to directory ... "

in_toto/runlib.py
```diff
@@ -785,6 +815,17 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
     securesystemslib.gpg.exceptions.CommandError:
         If gpg is used for signing and the command exits with a non-zero code.
 
+    IOError on Python 2 if the path doesn't exist or isn't a directory,
+        or we don't have the access to write the generated metadta to
```
lukpueh:
```suggestion
        or we don't have the access to write the generated metadata to
```

in_toto/runlib.py
```diff
@@ -497,6 +505,17 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
...
```
lukpueh:
```suggestion
        or we don't have the access to write the generated metadata to
```

in_toto/runlib.py
```diff
@@ -570,6 +589,10 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
   if signature:
     signing_keyid = signature["keyid"]
     filename = FILENAME_FORMAT.format(step_name=name, keyid=signing_keyid)
+
+    if metadata_directory is not None:
```
lukpueh, line 589: Could you add a:
```python
  if metadata_directory:
    securesystemslib.formats.PATH_SCHEMA.check_match(metadata_directory)
```
on top of the function body to fail early?

lukpueh: Same goes for `in_toto_record_stop`

tests/test_runlib.py
```diff
@@ -554,6 +585,40 @@ def test_in_toto_wrong_key(self):
...
+          True, self.key, metadata_directory='nonexistentDir')
+
+  def test_not_a_directory(self):
+    """Fail run, passed metadata_directory is not a directory. """
+    fd, path = tempfile.mkstemp()
+    os.write(fd, b"hello in-toto")
+    os.close(fd)
+    # Windows will raise FileNotFoundError instead of NotADirectoryError
+    expected_error = IOError if sys.version_info < (3, 0) \
+       else (NotADirectoryError, FileNotFoundError)
```
lukpueh:
```suggestion
        else (NotADirectoryError, FileNotFoundError)
```
Same indentation mistake exists in some of the other ternary expressions below. Please change there too.

lukpueh (changes requested):
Solid work, @SolidifiedRay! Thanks a ton!! :) I have left a few comments inline and here are some general observations:

Kudos for adding comprehensive runlib testing!! Can you please also test the new command line options? The corresponding test modules are `tests/test_in_toto_run.py` and `tests/test_in_toto_record.py`. Just look at the existing tests and how they make use of the `CliTestCase` class. IMO it's enough to just add one test (to each of the two modules) that checks that the cli argument actually writes link metadata to the specified dir.

> Also, perhaps worth worth cleaning up commits?

Yes! I'm okay with squashing all commits into one if you want. At any rate, it would make resolving the conflicts caused by the recently overhauled docstrings a lot easier if all changes to a given docstring in runlib would be in one commit only.

I'm happy to assist or take over the conflict resolving. I made a lot of changes to the docstring format and also consolidated the wording and I would like your changes to fit in.

> And worth exploring something more granular than os.name for the PermissionError test?

I think that's okay. Does "chmod doesn't work properly on Windows" mean that the builtin exists but doesn't have the desired effect? If so then yes, let's just skip on `os.name == 'nt'`. :)

lukpueh:
Thanks for addressing my comments @SolidifiedRay! I rebased onto develop and resolved the conflicts. I also took the liberty to re-word and shorten your code doc a bit (see 1673579). I hope that works for you.

May I ask you to write a nice commit message for 9f2ca1a? I think we are good to merge then.

SolidifiedRay:
Thanks a lot for reviewing my PR @lukpueh! I edited my commit message, may you check it?

lukpueh (approved):
🎉  Thanks for updating the commit message! I just did another quick pass over the PR and it looks great to me. :) Merging ...

PR #367, opened 2020-06-04
Add max timeout setting

kristelfung:
**Fixes issue #**: N/A

Allow the user to specify a timeout for the `in-toto-run` command in settings.py. 

requirements**:

### DIFF
--- a/doc/source/command-line-tools/settings.rst
+++ b/doc/source/command-line-tools/settings.rst
@@ -21,6 +21,8 @@ Setting Types
   ``in-toto-run`` are searched relative to the base path. The base path itself
   is not included in the link metadata. Default is the current working
   directory.
+- ``LINK_CMD_EXEC_TIMEOUT`` -- maximum timeout setting for the in-toto-run
+  command.
 
 
 Example Usage
@@ -31,6 +33,7 @@ Example Usage
   # Configure settings via bash-style environment variable export
   export IN_TOTO_ARTIFACT_BASE_PATH='/home/user/project'
   export IN_TOTO_ARTIFACT_EXCLUDE_PATTERNS='*.link:.gitignore'
+  export LINK_CMD_EXEC_TIMEOUT='10'
 
 .. code-block:: sh
 
@@ -38,3 +41,4 @@ Example Usage
   [in-toto settings]
   ARTIFACT_BASE_PATH=/home/user/project
   ARTIFACT_EXCLUDE_PATTERNS=*.link:.gitignore
+  LINK_CMD_EXEC_TIMEOUT=10
--- a/doc/source/command-line-tools/settings.rst
+++ b/doc/source/command-line-tools/settings.rst
@@ -33,7 +33,7 @@ Example Usage
   # Configure settings via bash-style environment variable export
   export IN_TOTO_ARTIFACT_BASE_PATH='/home/user/project'
   export IN_TOTO_ARTIFACT_EXCLUDE_PATTERNS='*.link:.gitignore'
-  export LINK_CMD_EXEC_TIMEOUT='10'
+  export IN_TOTO_LINK_CMD_EXEC_TIMEOUT='10'
 
 .. code-block:: sh
 
--- a/in_toto/settings.py
+++ b/in_toto/settings.py
@@ -40,3 +40,6 @@ ARTIFACT_EXCLUDE_PATTERNS = ["*.link*", ".git", "*.pyc", "*~"]
 # If not set the current working directory is used as base path
 # FIXME: Do we want different base paths for materials and products?
 ARTIFACT_BASE_PATH = None
+
+# Max timeout for the in-toto-run command
+LINK_CMD_EXEC_TIMEOUT = 10
--- a/tests/test_runlib.py
+++ b/tests/test_runlib.py
@@ -453,17 +453,17 @@ class TestLinkCmdExecTimeoutSetting(unittest.TestCase):
     timeout_old = in_toto.settings.LINK_CMD_EXEC_TIMEOUT
 
     # Modify timeout
-    in_toto.settings.LINK_CMD_EXEC_TIMEOUT = -1
+    in_toto.settings.LINK_CMD_EXEC_TIMEOUT = 0.1
 
     # check if exception is raised
     with self.assertRaises(securesystemslib.process.subprocess.TimeoutExpired):
       # Call execute_link to see if new timeout is respected
-      in_toto.runlib.execute_link(['python', '--version'], True)
+      in_toto.runlib.execute_link([sys.executable, '-c', 'while True: pass'], True)
 
     # check if exception is raised
     with self.assertRaises(securesystemslib.process.subprocess.TimeoutExpired):
       # Call execute_link to see if new timeout is respected
-      in_toto.runlib.execute_link(['python', '--version'], False)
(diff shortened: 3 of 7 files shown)


in_toto/settings.py
```diff
@@ -40,3 +40,6 @@
 # If not set the current working directory is used as base path
 # FIXME: Do we want different base paths for materials and products?
 ARTIFACT_BASE_PATH = None
+
+# Setting for max amount of time before command times out
+TIMEOUT = 10
```
trishankatdatadog: Needs a better name. Something like `LINK_CMD_EXEC_TIMEOUT`?

lukpueh (changes requested):
Thanks for the PR, @kristelfung! Can you please add tests?

Please also try to get rid of the merge commit (see our [git history guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/git-history.md) for details) and sign off your commits so that the DCO check passes.

kristelfung:
> Thanks for the PR, @kristelfung! Can you please add tests?
> 
> Please also try to get rid of the merge commit (see our [git history guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/git-history.md) for details) and sign off your commits so that the DCO check passes.

Yes, working on it! Perhaps you could clarify on what you mean by tests?

lukpueh:
> clarify on what you mean by tests

I'm thinking of a unittest for the function you changed that checks if a timeout configured via `LINK_CMD_EXEC_TIMEOUT` is adhered to. You can take a look at `securesystemslib`'s tests of its process module for inspiration.

nielstanis:
I was at the point filing an issue for being able to _configure_ the timeout of a run, got it when doing demo with in-toto. I can see that if you do a run that takes longer it might not be the right approach to do it; but the current 3MS is just too short for the things I'm doing. 
Any idea when this will be merged back to develop and released?

kristelfung:
> I was at the point filing an issue for being able to _configure_ the timeout of a run, got it when doing demo with in-toto. I can see that if you do a run that takes longer it might not be the right approach to do it; but the current 3MS is just too short for the things I'm doing.
> Any idea when this will be merged back to develop and released?

Hey @nielstanis, sorry for the delay! I will try to get it out by this weekend!

nielstanis:
Appreciate that @kristelfung !

tests/test_runlib.py
```diff
@@ -424,6 +424,23 @@ def test_hash_artifact_passing_algorithm(self):
...
+    timeout_old = in_toto.settings.LINK_CMD_EXEC_TIMEOUT
+    self.assertEqual(in_toto.settings.LINK_CMD_EXEC_TIMEOUT, timeout_old)
+
+    # Modify timeout and check that it returns the same value
+    timeout_new = timeout_old + 1
+    in_toto.settings.LINK_CMD_EXEC_TIMEOUT = timeout_new
+    self.assertEqual(in_toto.settings.LINK_CMD_EXEC_TIMEOUT, timeout_new)
+
+    # Restore original timeout
+    in_toto.settings.LINK_CMD_EXEC_TIMEOUT = timeout_old
```
lukpueh, line 469: Hooray, for adding a test and making it work. 🎉  

Unfortunately, this currently only test if Python's name assignment works, which is not very meaningful for you newly added feature. Instead it would be great to test if the test new `LINK_CMD_EXEC_TIMEOUT` actually affects the timeout of an executed process. An easy way to do this, and this is how securesystemslib and Python's standard library subprocess module do it, is to use -1 as timeout parameter and see if `TimeoutExpired` is raised immediately.

adityasaky: > is to use -1 as timeout parameter and see if TimeoutExpired is raised immediately.

@itsbenwu and @kristelfung arrived at that conclusion yesterday! :smile:

requirements-pinned.txt
```diff
@@ -31,7 +31,7 @@ cryptography==2.9         # via securesystemslib
 enum34==1.1.10             # via cryptography
 ipaddress==1.0.23         # via cryptography
 iso8601==0.1.12
-pathspec==0.7.0
+pathspec==0.8.0
```
lukpueh: Would you kindly rebase your branch on the most recent `develop`? This should make this line go away.

lukpueh (commented):
Great work! Could you try figure out, why the test fails on windows?

lukpueh (changes requested):
Thanks for fixing the tests, @kristelfung! Can you document the new setting in `doc/source/command-line-tools/settings.rst`?

doc/source/command-line-tools/settings.rst
```diff
@@ -31,10 +33,12 @@ Example Usage
   # Configure settings via bash-style environment variable export
   export IN_TOTO_ARTIFACT_BASE_PATH='/home/user/project'
   export IN_TOTO_ARTIFACT_EXCLUDE_PATTERNS='*.link:.gitignore'
+  export LINK_CMD_EXEC_TIMEOUT='10'
```
lukpueh: Have you tried if this works? IIRC it needs a prefix. Would you mind adding a test case for LINK_CMD_EXEC_TIMEOUT to test_user_settings.py?

tests/test_runlib.py
```diff
@@ -445,6 +445,30 @@ def test_hash_artifact_passing_algorithm(self):
...
+
+    # check if exception is raised
+    with self.assertRaises(securesystemslib.process.subprocess.TimeoutExpired):
+      # Call execute_link to see if new timeout is respected
+      in_toto.runlib.execute_link([sys.executable, '-c', 'while True: pass'], True)
+
+    # check if exception is raised
+    with self.assertRaises(securesystemslib.process.subprocess.TimeoutExpired):
+      # Call execute_link to see if new timeout is respected
+      in_toto.runlib.execute_link([sys.executable, '-c', 'while True: pass'], False)
```
lukpueh, line 466: Why test twice?

tests/test_runlib.py
```diff
@@ -445,6 +445,30 @@ def test_hash_artifact_passing_algorithm(self):
...
 
+class TestLinkCmdExecTimeoutSetting(unittest.TestCase):
+  """Tests LINK_CMD_EXEC_TIMEOUT setting in settings.py file. """
+
+  def test_timeout_setting(self):
+    # Save the old timeout and make sure it's saved properly
+    timeout_old = in_toto.settings.LINK_CMD_EXEC_TIMEOUT
+
+    # Modify timeout
+    in_toto.settings.LINK_CMD_EXEC_TIMEOUT = 0.1
```
lukpueh, line 456: The `int()` in `execute_link` will chop off the digits after the decimal point. Maybe we should use `float()` instead?

lukpueh (changes requested):

tests/test_runlib.py
kristelfung: When passing in `True`, `execute_link` calls `run_duplicate_streams`, and when passing in `False`, we call `run`. I think I had to test them both

tests/test_runlib.py
lukpueh: Why did you resolve this comment? Do you not think that we should cast to `float` rather than to `int`?

kristelfung: Oops sorry, I had corrected it locally and didn't push changes

lukpueh (approved):
LGTM!! 🎉 

Thanks, @kristelfung!

PR #368, opened 2020-06-08
Update test_runlib.py

IshaVDave:
Fixes issue in-toto#365:

Now checks the platform of the user's system. If it's windows, it raises an error that asks the user if the developer mode is activated.

### DIFF
--- a/tests/test_runlib.py
+++ b/tests/test_runlib.py
@@ -277,6 +277,14 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
     self.assertListEqual(sorted(list(artifacts_dict.keys())),
       sorted(self.full_file_path_list))
 
+  @staticmethod
+  def _raise_win_dev_mode_error():
+    """If the platform is Windows, raises an error that asks the user if
+    developer mode is activated."""
+    if os.name == "nt":
+      raise IOError("Developer mode is required to work with symlinks on "
+          "Windows. Is it enabled?")
+
 
   @unittest.skipIf("symlink" not in os.__dict__, "symlink is not supported in this platform")
   def test_record_symlinked_files(self):
@@ -292,7 +300,11 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
     for pair in link_pairs:
       # We only use the basename of the file (source) as it is on the same
       # level as the link (target)
-      os.symlink(os.path.basename(pair[0]), pair[1])
+      try:
+        os.symlink(os.path.basename(pair[0]), pair[1])
+      except IOError:
+        TestRecordArtifactsAsDict._raise_win_dev_mode_error()
+        raise
 
     # Record files and linked files
     # follow_symlink_dirs does not make a difference as it only concerns linked dirs
@@ -325,7 +337,11 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
 
     # Create dead links
     for link in links:
-      os.symlink("does/not/exist", link)
+      try:
+        os.symlink("does/not/exist", link)
+      except IOError:
+        TestRecordArtifactsAsDict._raise_win_dev_mode_error()
+        raise
 
     # Record files without dead links
     # follow_symlink_dirs does not make a difference as it only concerns linked dirs
@@ -345,8 +361,12 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
   def test_record_follow_symlinked_directories(self):
     """Record files in symlinked dirs if follow_symlink_dirs is True. """
 
-    # Link to subdir
-    os.symlink("subdir", "subdir_link")
+    try:
+      # Link to subdir
+      os.symlink("subdir", "subdir_link")
+    except IOError:
+      TestRecordArtifactsAsDict._raise_win_dev_mode_error()
+      raise
 
     link_pairs = [
       ("subdir/foosub1", "subdir_link/foosub1"),
@@ -363,7 +383,6 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
     # ... and the hashes of each link/file pair match
     for pair in link_pairs:
       self.assertDictEqual(artifacts_dict[pair[0]], artifacts_dict[pair[1]])
-
 
     # Record with follow_symlink_dirs FALSE (default)
     artifacts_dict = record_artifacts_as_dict(["."])


tests/test_runlib.py
```diff
@@ -1,22 +1,18 @@
+
```
lukpueh: Please don't add or remove blank lines that are not related to the lines you are working on.

tests/test_runlib.py
```diff
@@ -278,99 +274,111 @@ def test_record_dot_check_files_hash_dict_schema(self):
       sorted(self.full_file_path_list))
 
 
+  def raiseException():
+    if os.name == "nt":
+      raise IOError("Is developer mode enabled?")
```
lukpueh: Some notes on [code style](https://github.com/secure-systems-lab/code-style-guidelines) and programming style: 
- We usually use snake_case and not camelCase for function names.
- We often start small helper functions like this with an underscore.
- Please try to choose mnemonic function names. I would maybe go for something like `_raise_win_dev_mode_error`.
- Please always add a docstring, to your functions, for short functions like this one it can be a oneliner.
- I think the error message could need some additional context. If I saw this in my error log, I would like to know why this IOError asks me about developer mode. :)
- It might be helpful (also for context) to include the error message from the original exception.

lukpueh (changes requested):
Congrats on your first PR, @IshaDave! 🎉 This looks quite good to me. I have left a couple of suggestions inline. Please address them and I'll give it another look.

On a more general note, it would be great if you could write [more meaningful commit messages](https://github.com/secure-systems-lab/lab-guidelines/blob/master/commits.md#how-to-write-good-commit-messages).

IshaVDave:
Thank you Lukas! I was also surprised about the deleted lines because I
hadn't done it intentionally. I'll fix that and implement all the code
style and context-related changes ASAP :)

lukpueh:
Thanks, @IshaDave!

> I was also surprised about the deleted lines because I hadn't done it intentionally

I suspect it was your sneaky IDE. ;)  As a general rule, please always make sure that the [diff](https://github.com/in-toto/in-toto/pull/368/files) of your PRs only contain the changes you intended.

IshaVDave:
Hi @lukpueh. I implemented all the edits, let me know if you feel something else needs to be changed too!

lukpueh:
Thank you, @IshaDave! Did you take a look at the changes your PR contains, as I asked above? It seems to me that there are still unintended edits. :P

Also, we perform automatic checks on PRs and require them to pass before we merge. As you can see in the table below "Some checks were not successful". Can you please try to find out what went wrong and fix it?

IshaVDave:
Okay, sure. On it!

tests/test_runlib.py
```diff
@@ -278,99 +278,111 @@ def test_record_dot_check_files_hash_dict_schema(self):
       sorted(self.full_file_path_list))
 
 
+  def _raise_win_dev_mode_error(self):
+    # If the platform is Windows, raises an error that asks the user if developer mode is activated.
```
lukpueh: Minor nit: This is a comment (`# ....`) but should be a docstring (`"""...."""`).

tests/test_runlib.py
```diff
@@ -278,99 +278,111 @@ def test_record_dot_check_files_hash_dict_schema(self):
...
+            sorted(self.full_file_path_list + [pair[1] for pair in link_pairs]))
+
+        # ... and the hashes of each link/file pair match
+        for pair in link_pairs:
+          self.assertDictEqual(artifacts_dict[pair[0]], artifacts_dict[pair[1]])
+
+      for pair in link_pairs:
+        os.unlink(pair[1])
+    except IOError:
+      TestRecordArtifactsAsDict._raise_win_dev_mode_error(self)
```
lukpueh: Minor nit: I suggest this syntax `self. _raise_win_dev_mode_error()` when calling instance methods. It just feels more idiomatic. I tend to only use the `ClassName.method_name`-syntax when calling class- or static-methods.

By the way, since you don't need the method nor the class instance in `_raise_win_dev_mode_error` you should consider making it a `staticmethod`. :)

@SantiagoTorres, @adityasaky, what do you think?

lukpueh (changes requested):
Thanks for fixing the git history, @IshaDave! I have left to minor comments inline and I also have one general concern about the design here. Correct me if I'm wrong, but I am pretty sure that if any of the code in your try block(s) raises an `IOError` and the platform is not Windows, then the test would still pass, which is not what we want.

The issue is amplified by the fact that there is a lot of code in the try block, so there are a lot of places where an unexpected error could be raised. It generally is good programming style to only wrap the part in try/else that is expected to raise an exception.

I think you should do the following:
1. Wrap only the line that you expect to raise an IOError on Windows in a `try` block. (It should be enough to only wrap the first such line, because if there is a problem, the test should not get past it anyways).
2. In the corresponding `except` block you check
    - If you're indeed on Windows, and raise the exception with the developer mode hint (as you already do),
    - Otherwise, if you are not on Windows, the exception comes fully unexpected and we sure want to know about it. This means the exception must be re-`raise`ed. :)

What do you think?

tests/test_runlib.py
IshaVDave: I agree. I initially had planned to make it static, but decided instead to make it uniform with all the other methods. Now that you voice my opinion too, I think I'll just go with static :)

lukpueh: Regarding uniformity, the difference is that `_raise_win_dev_mode_error` is a pure utility function with no ties to the class or its instances, whereas the other methods do need to be instance methods to be executed as `unittest`.

IshaVDave:
> I think you should do the following:
> 
> 1. Wrap only the line that you expect to raise an IOError on Windows in a `try` block. (It should be enough to only wrap the first such line, because if there is a problem, the test should not get past it anyways).
> 2. In the corresponding `except` block you check
>    
>    * If you're indeed on Windows, and raise the exception with the developer mode hint (as you already do),
>    * Otherwise, if you are not on Windows, the exception comes fully unexpected and we sure want to know about it. This means the exception must be re-`raise`ed. :)
> 
> What do you think?

Yes, I realize what you pointed out. I had this conversation with @SantiagoTorres too, but I missed on incorporating the 're-raising'. I'll do that, thanks a lot! :)

tests/test_runlib.py
```diff
@@ -277,6 +277,12 @@ def test_record_dot_check_files_hash_dict_schema(self):
     self.assertListEqual(sorted(list(artifacts_dict.keys())),
       sorted(self.full_file_path_list))
 
+  @staticmethod
+  def _raise_win_dev_mode_error():
+    """If the platform is Windows, raises an error that asks the user if developer mode is activated."""
+    if os.name == "nt":
+      raise IOError("Developer mode is required to work with symlinks on Windows. Is it enabled?")
```
lukpueh: Looks like we [don't flag line length in the tests](https://github.com/in-toto/in-toto/blob/41878e286eafceba41dac8d1e352eff1c7c44713/tox.ini#L29-L31). Could you still please wrap the docstring and the error message at 80 chars? Thanks. :)

lukpueh (approved):
Excellent work, @IshaDave!

Issue #378, opened 2020-07-27
duplicate module imports


shibumi:
Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.

**Description of issue or feature request**:
@trishankatdatadog was so nice and reported, that we have a few duplicate imports. We should get rid of those, if possible.

https://lgtm.com/projects/g/in-toto/in-toto/?mode=list

**Current behavior**:
duplicate imports, that are not necessary.

**Expected behavior**:
import necessary modules just once.

lukpueh:
FYI *lgtm* has the following [recommendation](https://lgtm.com/projects/g/in-toto/in-toto/?mode=list) to fix this:

*"Remove the from xxx import yyy statement. Add yyy = xxx.yyy if required."*

lukpueh:
(Also, and on an unrelated side-note, in-toto's lgtm report looks pretty good to me.) 💯  

(ping https://github.com/secure-systems-lab/securesystemslib/issues/243)

shibumi:
@lukpueh definitely. It's really nitpicking here. I will quickly prepare a PR for this. I don't think it will take much time.

shibumi:
Fixed in: https://github.com/in-toto/in-toto/pull/379

PR #379, opened 2020-07-28
Fix issue 378: duplicate imports

shibumi:
We had a few unnecessary duplicate imports.
This commit removes the additional "from-import" lines,
and just uses the full path for the import.

that is provided, the better.

**Fixes issue #**:

Fixes issue #378 

Use imports and not `from .. import ..`-lines. I also realized that the latter is against our Code Style Guidelines.. ^^

requirements**:

### DIFF
--- a/in_toto/util.py
+++ b/in_toto/util.py
@@ -12,8 +12,6 @@ import securesystemslib.exceptions
 import securesystemslib.gpg.functions
 
 from in_toto.exceptions import UnsupportedKeyTypeError
-from securesystemslib.interface import (import_ed25519_privatekey_from_file,
-    import_ed25519_publickey_from_file)
 
 DEFAULT_RSA_KEY_BITS = 3072
 
@@ -170,7 +168,8 @@ def import_public_keys_from_files_as_dict(filepaths, key_types=None):
   for idx, filepath in enumerate(filepaths):
 
     if key_types[idx] == KEY_TYPE_ED25519:
-      key = import_ed25519_publickey_from_file(filepath)
+      key = securesystemslib.interface.import_ed25519_publickey_from_file(
+          filepath)
     elif key_types[idx] == KEY_TYPE_RSA:
       key = import_rsa_key_from_file(filepath)
     else:  # pragma: no cover
@@ -241,10 +240,12 @@ def prompt_import_ed25519_privatekey_from_file(filepath):
   """
   password = None
   try:
-    import_ed25519_privatekey_from_file(filepath, password)
+    securesystemslib.interface.import_ed25519_privatekey_from_file(
+        filepath, password)
   except securesystemslib.exceptions.CryptoError:
     password = prompt_password()
-  return import_ed25519_privatekey_from_file(filepath, password)
+  return securesystemslib.interface.import_ed25519_privatekey_from_file(
+      filepath, password)
 
 
 def prompt_import_rsa_key_from_file(filepath):
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -39,8 +39,6 @@ import in_toto.models.layout
 import in_toto.models.link
 import in_toto.formats
 from in_toto.models.metadata import Metablock
-from in_toto.models.link import (FILENAME_FORMAT, FILENAME_FORMAT_SHORT)
-from in_toto.models.layout import SUBLAYOUT_LINK_DIR_FORMAT
 from in_toto.exceptions import (RuleVerificationError, LayoutExpiredError,
     ThresholdVerificationError, BadReturnValueError,
     SignatureVerificationError)
@@ -144,7 +142,8 @@ def load_links_for_layout(layout, link_dir_path):
       for keyid in [authorized_keyid] + list(layout.keys.get(authorized_keyid,
           {}).get("subkeys", {}).keys()):
 
-        filename = FILENAME_FORMAT.format(step_name=step.name, keyid=keyid)
+        filename = in_toto.models.link.FILENAME_FORMAT.format(
+            step_name=step.name, keyid=keyid)
         filepath = os.path.join(link_dir_path, filename)
 
         try:
@@ -222,7 +221,8 @@ def run_all_inspections(layout):
 
     # Dump the inspection link file for auditing
     # Keep in mind that this pollutes the verifier's (client's) filesystem.
-    filename = FILENAME_FORMAT_SHORT.format(step_name=inspection.name)
+    filename = in_toto.models.link.FILENAME_FORMAT_SHORT.format(
+        step_name=inspection.name)
     link.dump(filename)
 
     in_toto.settings.ARTIFACT_BASE_PATH = base_path_backup
@@ -1311,11 +1311,11 @@ def verify_sublayouts(layout, chain_link_dict, superlayout_link_dir_path):
         # name relative the the current link directory path, i.e. if there
         # are multiple levels of sublayout nesting, the links are expected to
         # be nested accordingly
-        sublayout_link_dir = SUBLAYOUT_LINK_DIR_FORMAT.format(
+        sub_link_dir = in_toto.models.layout.SUBLAYOUT_LINK_DIR_FORMAT.format(
             name=step_name, keyid=keyid)
 
         sublayout_link_dir_path = os.path.join(
-            superlayout_link_dir_path, sublayout_link_dir)
+            superlayout_link_dir_path, sub_link_dir)
 
         # Make a recursive call to in_toto_verify with the
         # layout and the extracted key object


shibumi:
I have tested the PR with calling `python tests/runtests.py`. I hope this is enough.

trishankatdatadog (approved):

shibumi:
ah okay, the linting fails, because the line is too long.

shibumi:
Mh is not that easy to satisfy the linter.. it's complaining about the variable name `m` or `ml`, because it has to be snakecase..

shibumi:
Mh I have added a pylint exception now for the few lines. Otherwise I couldn't make this PR work, I have no idea if it's worth it to fix a few double imports.

lukpueh (commented):
Thanks for fixing the *lgtm* warning, @shibumi!

This might be a matter of taste, but a local scope module shortcut does not make a lot of sense to me, especially if then used only once.

I suggest do either make the shortcut module scope, which is what we had before this PR, or use the full module path in the call statement and wrap the line appropriately, which makes it look more like java but is what we usually do in in-toto/tuf/securesystemslib code.

I also like what [google's style guide suggests](https://google.github.io/styleguide/pyguide.html#22-imports), i.e. to:
```python
from full.path.to.package import module
...
module.function()
```

The caveat I see there, is that the module name by itself should be meaningful, which is not really the case for `interface`, and must not conflict with any locally defined variables, which is not the case for `link`. But we could work around this. :)

What do you think?

shibumi:
Hi @lukpueh 
> This might be a matter of taste, but a local scope module shortcut does not make a lot of sense to me, especially if then used only once.

Sounds good, I think I will try the Java-like style first to stick with our current styleguide.

shibumi:
@lukpueh I have modified the lines, we are using linewraps now. However, my IDE pycharm states, that we are violating PEP8 now.

IIRC, PEP8 compliance has been discussed in the newest code style issue, too. Right?

in_toto/verifylib.py
```diff
@@ -222,7 +221,10 @@ def run_all_inspections(layout):
 
     # Dump the inspection link file for auditing
     # Keep in mind that this pollutes the verifier's (client's) filesystem.
-    filename = FILENAME_FORMAT_SHORT.format(step_name=inspection.name)
+    # ignore snakecase for pylint to satisfy line limit
+    # pylint: disable=C0103
+    m = in_toto.models.link
+    filename = m.FILENAME_FORMAT_SHORT.format(step_name=inspection.name)
```
lukpueh: [ping](https://github.com/in-toto/in-toto/pull/379#pullrequestreview-459873487)

lukpueh (changes requested):
Thanks for the quick patch, @shibumi. But this is not exactly what I meant. I think escaping newlines is really ugly and should only be done if the other alternatives don't work for some reasons.

The preferred (and [PEP8 compliant](https://www.python.org/dev/peps/pep-0008/#indentation)) way is to do hanging indent with an extra indentation level, or aligned indent.

I think in in-toto we almost exclusively use hanging over aligned indent, albeit with arguments on the first line, which is disallowed by PEP8 but IMO doesn't impede readability.

in_toto/verifylib.py
shibumi: oops, looks like I missed one

shibumi:
Ok, I tried using a hanging indent for most of them, however one line is still too long. I tried renaming the variable, but I think making `sublayout_link_dir` to `sub_link_dir` would destroying the readability of the variable name. It will be more difficult to tell, what the variable is doing, although we use the variable only in the next line. What do you think?

in_toto/util.py
```diff
@@ -170,7 +168,8 @@ def import_public_keys_from_files_as_dict(filepaths, key_types=None):
   for idx, filepath in enumerate(filepaths):
 
     if key_types[idx] == KEY_TYPE_ED25519:
-      key = import_ed25519_publickey_from_file(filepath)
+      key = securesystemslib.interface.import_ed25519_publickey_from_file(
+        filepath)
```
lukpueh:
```suggestion
          filepath)
```

in_toto/verifylib.py
```diff
@@ -1311,11 +1311,11 @@ def verify_sublayouts(layout, chain_link_dict, superlayout_link_dir_path):
         # name relative the the current link directory path, i.e. if there
         # are multiple levels of sublayout nesting, the links are expected to
         # be nested accordingly
-        sublayout_link_dir = SUBLAYOUT_LINK_DIR_FORMAT.format(
-            name=step_name, keyid=keyid)
+        sub_link_dir = in_toto.models.layout.SUBLAYOUT_LINK_DIR_FORMAT.format(
```
shibumi, line 1314: This is what i mean.. I am not sure if `sub_link_dir` is okay as variable name... However we only use it in the next line. What do you think?

in_toto/util.py
```diff
@@ -241,10 +240,12 @@ def prompt_import_ed25519_privatekey_from_file(filepath):
   """
   password = None
   try:
-    import_ed25519_privatekey_from_file(filepath, password)
+    securesystemslib.interface.import_ed25519_privatekey_from_file(
+      filepath, password)
```
lukpueh:
```suggestion
        filepath, password)
```

in_toto/util.py
```diff
@@ -241,10 +240,12 @@ def prompt_import_ed25519_privatekey_from_file(filepath):
...
   except securesystemslib.exceptions.CryptoError:
     password = prompt_password()
-  return import_ed25519_privatekey_from_file(filepath, password)
+  return securesystemslib.interface.import_ed25519_privatekey_from_file(
+    filepath, password)
```
lukpueh:
```suggestion
      filepath, password)
```

in_toto/verifylib.py
```diff
@@ -144,7 +142,8 @@ def load_links_for_layout(layout, link_dir_path):
       for keyid in [authorized_keyid] + list(layout.keys.get(authorized_keyid,
           {}).get("subkeys", {}).keys()):
 
-        filename = FILENAME_FORMAT.format(step_name=step.name, keyid=keyid)
+        filename = in_toto.models.link.FILENAME_FORMAT.format\
+          (step_name=step.name, keyid=keyid)
```
lukpueh:
```suggestion
        filename = in_toto.models.link.FILENAME_FORMAT.format(
            step_name=step.name, keyid=keyid)
```

in_toto/verifylib.py
```diff
@@ -222,7 +221,8 @@ def run_all_inspections(layout):
 
     # Dump the inspection link file for auditing
     # Keep in mind that this pollutes the verifier's (client's) filesystem.
-    filename = FILENAME_FORMAT_SHORT.format(step_name=inspection.name)
+    filename = in_toto.models.link.FILENAME_FORMAT_SHORT.format(
+      step_name=inspection.name)
```
lukpueh:
```suggestion
        step_name=inspection.name)
```

in_toto/verifylib.py
```diff
@@ -1311,11 +1311,11 @@ def verify_sublayouts(layout, chain_link_dict, superlayout_link_dir_path):
         # name relative the the current link directory path, i.e. if there
         # are multiple levels of sublayout nesting, the links are expected to
         # be nested accordingly
-        sublayout_link_dir = SUBLAYOUT_LINK_DIR_FORMAT.format(
-            name=step_name, keyid=keyid)
+        sub_link_dir = in_toto.models.layout.SUBLAYOUT_LINK_DIR_FORMAT.format(
+          name=step_name, keyid=keyid)
```
lukpueh:
```suggestion
            name=step_name, keyid=keyid)
```

in_toto/verifylib.py
lukpueh, line 1314: 👍  works for me.

lukpueh (changes requested):

lukpueh:
Added **double** indentation for the hanging indent and replaced one escaped newline. Feel free to batch apply my suggestions and squash into your commit. I think we can merge then. :)

shibumi:
@lukpueh please verify if this has worked correctly. I am not sure about my git skills here. What I did is:

1. applied all changes via github GUI
2. checked out this PR directly via one of my git alias'es
3. rebased the PR branch on my fork branch
4. force-pushed to this PR again.

lukpueh (approved):
LGTM

trishankatdatadog:
Great job, guys! Sorry this was not as simple as expected...

lukpueh:
No worries, @trishankatdatadog. Maybe we should put our impeccable lgtm rating to the README.md? I reckon there is a badge for it?

trishankatdatadog:
I think let's try enabling CodeQL instead (appears to be the next evolution of LGTM now that it has been acquired by GitHub). Could you see if we can turn on the [public beta](https://github.blog/changelog/2020-05-06-github-advanced-security-code-scanning-now-available-in-limited-public-beta/)?

shibumi:
In case you want a badge for LGTM.. here is one:

https://img.shields.io/lgtm/grade/python/github/secure-systems-lab/securesystemslib

(built via https:/shields.io )

Issue #260, opened 2019-03-04
Parametrize writing of inspection links


lukpueh:
**Description of issue or feature request**:
During final product verification, inspections defined in the layout are *run* in a similar manner as functionaries *run* steps of the layout, generating link metadata (see [`verifylib.run_all_inspections`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/verifylib.py#L160)).

The corresponding link metadata is required for inspection artifact rule verification (see [`verifylib.in_toto_verify#L1551-L1558`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/verifylib.py#L1551-L1558)).

In addition, each link metadata object is written to disk in the `cwd` of where `in_toto_verify` is called (see [`verifylib.run_all_inspections#L211-L214`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/verifylib.py#L211-L214)). This has two reasons:
- propagation of sublayout inspections to the superlayout (see [4.5.3 Inspections on sublayouts](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#453-inspections-on-sublayouts) of the specification).
- auditing

The former reason is only relevant if the inspection links need to be propagated to a superlayout final product for further verification, which is not the case if verification takes place at the end-user.

Also see related discussion in https://github.com/in-toto/in-toto/pull/258.

**Current behavior**:
Link metadata files for inspection's are always written to `cwd` where `in_toto_verify` is called.

**Expected behavior**:
Provide parameter to `in_toto_verify` to control whether link metadata files for inspections should be written to `cwd`.  Several approaches are conceivable:
- a boolean `persist_inspection_links` (only write if True)
- a path string `inspection_link_destination` (only write if supplied)
- a combination of above (only write if boolean is True, write to path if supplied, to cwd otherwise)

trishankatdatadog:
Good idea, but let's not have two parameters, please :)

adityasaky:
Do we want to keep this open for the possibility of supporting passing in a target directory?

PR #380, opened 2020-08-02
Issue: 260 - Paramaterize inspection links

ghost:
**Fixes issue #260**:

Added an optional boolean parameter to run_all_inspections in verifylib.py which determines whether or not a metadata link file is written. When true, the file is written to the client's cwd. When false, no file is written. Set to true by default, since that was the behavior before this change.

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -39,8 +39,6 @@ import in_toto.models.layout
 import in_toto.models.link
 import in_toto.formats
 from in_toto.models.metadata import Metablock
-from in_toto.models.link import (FILENAME_FORMAT, FILENAME_FORMAT_SHORT)
-from in_toto.models.layout import SUBLAYOUT_LINK_DIR_FORMAT
 from in_toto.exceptions import (RuleVerificationError, LayoutExpiredError,
     ThresholdVerificationError, BadReturnValueError,
     SignatureVerificationError)
@@ -144,7 +142,8 @@ def load_links_for_layout(layout, link_dir_path):
       for keyid in [authorized_keyid] + list(layout.keys.get(authorized_keyid,
           {}).get("subkeys", {}).keys()):
 
-        filename = FILENAME_FORMAT.format(step_name=step.name, keyid=keyid)
+        filename = in_toto.models.link.FILENAME_FORMAT.format(
+            step_name=step.name, keyid=keyid)
         filepath = os.path.join(link_dir_path, filename)
 
         try:
@@ -1317,11 +1316,11 @@ def verify_sublayouts(layout, chain_link_dict, superlayout_link_dir_path):
         # name relative the the current link directory path, i.e. if there
         # are multiple levels of sublayout nesting, the links are expected to
         # be nested accordingly
-        sublayout_link_dir = SUBLAYOUT_LINK_DIR_FORMAT.format(
+        sub_link_dir = in_toto.models.layout.SUBLAYOUT_LINK_DIR_FORMAT.format(
             name=step_name, keyid=keyid)
 
         sublayout_link_dir_path = os.path.join(
-            superlayout_link_dir_path, sublayout_link_dir)
+            superlayout_link_dir_path, sub_link_dir)
 
         # Make a recursive call to in_toto_verify with the
         # layout and the extracted key object
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -167,7 +167,7 @@ def load_links_for_layout(layout, link_dir_path):
   return steps_metadata
 
 
-def run_all_inspections(layout):
+def run_all_inspections(layout, persist_inspection_links):
   """
   <Purpose>
     Extracts all inspections from a passed Layout's inspect field and
@@ -180,6 +180,10 @@ def run_all_inspections(layout):
   <Arguments>
     layout:
             A Layout object which is used to extract the Inspections.
+
+    persist_inspection_links:
+            A boolean that determines whether or not link metadata files for
+            inspection are written to cwd.
 
   <Exceptions>
     Calls function that raises BadReturnValueError if an inspection returned
@@ -220,10 +224,12 @@ def run_all_inspections(layout):
 
     inspection_links_dict[inspection.name] = link
 
+    # If the client requests persistent inspection links,
     # Dump the inspection link file for auditing
     # Keep in mind that this pollutes the verifier's (client's) filesystem.
-    filename = FILENAME_FORMAT_SHORT.format(step_name=inspection.name)
-    link.dump(filename)
+    if persist_inspection_links:
+      filename = FILENAME_FORMAT_SHORT.format(step_name=inspection.name)
+      link.dump(filename)
 
     in_toto.settings.ARTIFACT_BASE_PATH = base_path_backup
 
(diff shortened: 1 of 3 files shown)


adityasaky:
Thanks for taking this issue up, @alexsmith15! One of us will be along shortly to take a look.

tests/test_verifylib.py
```diff
@@ -146,6 +146,38 @@ def test_inspection_fail_with_non_zero_retval(self):
     with self.assertRaises(BadReturnValueError):
       run_all_inspections(layout)
 
+  def test_inspection_persistence_false(self):
+    """Test metadata link file persistence"""
+    scripts_directory = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "scripts")
+
+    layout = Layout.read({
```
lukpueh: You could just use `self.layout` instead of redefining one here.

tests/test_verifylib.py
```diff
@@ -146,6 +146,38 @@ def test_inspection_fail_with_non_zero_retval(self):
...
+    })
+    run_all_inspections(layout)
+    self.assertTrue(os.path.exists("touch-foobar.link"))
+
+  def test_inspection_persistence_true(self):
+    """Test metadata link file non-persistence"""
+    scripts_directory = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "scripts")
+
+    layout = Layout.read({
```
lukpueh: See comment above about `self.layout`

in_toto/verifylib.py
```diff
@@ -1423,6 +1432,12 @@ def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
 
     step_name (optional): A name assigned to the returned link. This is mostly
         useful during recursive sublayout verification.
+    
+    persist_inspection_links: (optional)
+            A boolean that determines whether or not link metadata files for
+            inspection are written to cwd.
+            When true, the link inspection file is written to cwd.
+            When false, the file is not written.
```
lukpueh: Minor nit: Can you un-indent one level to be consistent with the other arguments? Also, I think the the first two lines of the description should be enough to understand how to use the argument.

in_toto/verifylib.py
```diff
@@ -167,7 +167,7 @@ def load_links_for_layout(layout, link_dir_path):
   return steps_metadata
 
 
-def run_all_inspections(layout):
+def run_all_inspections(layout, persist_inspection_links=True):
```
lukpueh: Minor nit: I think it's okay to make this a non-optional argument here. 

Although not identified as such by leading underscore(s), `run_all_inspections`  is really more of an internal function that's only called by `in_toto_verify`, which passes `persist_inspection_links` no matter what.

lukpueh (commented):
Great work, @alexsmith15, and thanks for taking a stab at #260! 

@trishankatdatadog, @radu-matei, since you signaled interest in the original issue, would you mind giving feedback whether an on/off solution fits our needs? Or if we also/instead want to allow a custom path for inspection links?

tests/test_verifylib.py
```diff
@@ -146,6 +146,38 @@ def test_inspection_fail_with_non_zero_retval(self):
     with self.assertRaises(BadReturnValueError):
       run_all_inspections(layout)
 
+  def test_inspection_persistence_false(self):
```
trishankatdatadog, line 149:
```suggestion
  def test_inspection_persistence_true(self):
```

tests/test_verifylib.py
```diff
@@ -146,6 +146,38 @@ def test_inspection_fail_with_non_zero_retval(self):
...
+        "steps": [],
+        "inspect": [{
+          "name": "touch-foobar",
+          "run": ["python", os.path.join(scripts_directory, "touch"), "bar"],
+        }]
+    })
+    run_all_inspections(layout)
+    self.assertTrue(os.path.exists("touch-foobar.link"))
+
+  def test_inspection_persistence_true(self):
```
trishankatdatadog:
```suggestion
  def test_inspection_persistence_false(self):
```

trishankatdatadog (changes requested):
> @trishankatdatadog, @radu-matei, since you signaled interest in the original issue, would you mind giving feedback whether an on/off solution fits our needs? Or if we also/instead want to allow a custom path for inspection links?

Thanks for your work, @alexsmith15, and for pinging us, @lukpueh! I think this is good enough for now, aside for some minor changes. What do you think, @radumatei?

trishankatdatadog:
> Thanks for your work, @alexsmith15, and for pinging us, @lukpueh! I think this is good enough for now, aside for some minor changes. What do you think, @RaduMatei?

Especially since the Python reference implementation still implicitly assumes storing the layout, links, and pubkeys all in the same directory anyway, right? (Which is why we need to `ALLOW` for these files as materials to the inspection.)

lukpueh:
> Especially since the Python reference implementation still implicitly assumes storing the layout, links, and pubkeys all in the same directory anyway, right? (Which is why we need to ALLOW for these files as materials to the inspection.)

in-toto-run/verify cli/api allows or requires you to pass paths for all these.

trishankatdatadog:
> in-toto-run/verify cli/api allows or requires you to pass paths for all these.

Then we're set, I think.

SantiagoTorres:
Awesome stuff @alexsmith15 ! could you please make sure to sign-off your commits and re-push? :)

ghost:
Hey thanks for the feedback and sorry for the delay! Yes I will get that fixed by the end of the day today!

ghost:
Re-did commits with sign-off and addressed the changes that were suggested. Let me know if I need to make other changes or if I missed someting!

tests/test_verifylib.py
```diff
@@ -144,7 +144,18 @@ def test_inspection_fail_with_non_zero_retval(self):
         }]
     })
     with self.assertRaises(BadReturnValueError):
-      run_all_inspections(layout)
+      run_all_inspections(layout, True)
+
+  def test_inspection_persistence_false(self):
+    """Test metadata link file non-persistence"""
+    os.remove("touch-bar.link")
```
lukpueh, line 151: It's a good idea to make sure that there is no `touch-bar.link` before running the inspections, but blindly removing it if it indeed does not exist fails the tests. I suggest to just assert that it does not exist.

in_toto/verifylib.py
```diff
@@ -219,11 +223,12 @@ def run_all_inspections(layout):
...
     inspection_links_dict[inspection.name] = link
 
+    # If the client requests persistent inspection links,
     # Dump the inspection link file for auditing
     # Keep in mind that this pollutes the verifier's (client's) filesystem.
-    filename = in_toto.models.link.FILENAME_FORMAT_SHORT.format(
-        step_name=inspection.name)
-    link.dump(filename)
+    if persist_inspection_links:
+      filename = FILENAME_FORMAT_SHORT.format(step_name=inspection.name)
```
lukpueh, line 230: The direct import of `FILENAME_FORMAT_SHORT` was removed in a recent PR (#379) please use the full path.

lukpueh (changes requested):
Many thanks for resolving the conflicts and addressing the review comments! 🎉 

Please always make sure that the test pass before you request a review  (see [review guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/code-review.md)) and please also don't include merge commits in your PRs (see [git history guidelines](https://github.com/secure-systems-lab/lab-guidelines/blob/master/git-history.md)). :)

adityasaky:
@kristelfung's takeover PR #392 has now been merged.

PR #382, opened 2020-08-14
Prepare Debian 0.5.0-1 release

lukpueh:
- Replace never released 0.4.1-1 changelog entry, including note-
  worthy upstream changes since 0.4.0
- Update securesystemslib dependency to >= 0.16.0


SantiagoTorres:
I don't see any issues here. And consider it was accepted on debian I'm merging. Thank you!

PR #395, opened 2020-09-18
fix dangling link : layout-creation.md

didier-durand:
Hello,

Link //github.com/in-toto/in-toto/blob/develop/layout-creation.md gets a 404 in README.md

So, the correct one is probably now : https://in-toto.readthedocs.io/en/latest/layout-creation-example.html

Didier

Signed-off-by: didier <durand.didier@gmail.com>


adityasaky (approved):
LGTM, thanks for the fix @didier-durand!

PR #408, opened 2020-11-05
Adopt recent sslib key interface changes

lukpueh:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #**: #80
(Follows up on #402 to address recent changes in pending secure-systems-lab/securesystemslib#288)

**Description of the changes being introduced by the pull request**:
PR #402 adopted key interface changes from the pending secure-systems-lab/securesystemslib#288 PR and was merged prematurely. The sslib PR now has further evolved, in order to follow the principle of secure defaults in regards to private key encryption, which requires the following adoptions in in-toto:

- The not secure by default `generate_and_write_*_keypair` function is now protected (`_generate_and_write_*_keypair`), and only used for the keygen cli utility, where it is really convenient.

- In other cases we use either `generate_and_write_*_keypair` (for encrypted keys only) or `generate_and_write_unencrypted_*_keypair`.

Furthermore, the newly added sslib key generation interface functions are added to the in-toto API docs, including additional notes and cross-references relevant for key handling.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #405, opened 2020-10-28
Misc docs fixes

lukpueh:
**Fixes issue #**: -

**Description of the changes being introduced by the pull request**:
- Create new docs section for tool configuration-related documentation
- Move settings docs and artifact exclude pattern docs to that section
- Fix/enhance wording and formatting

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #402, opened 2020-10-21
Remove util module and use securesystemslib key interfaces instead

lukpueh:
NOTE: Requires secure-systems-lab/securesystemslib#288 to be merged and released. Currently the required securesystemslib are pulled in via f1f02f4, which should be removed before merging here.

**Fixes issue #80**:

**Description of the changes being introduced by the pull request**:

With https://github.com/secure-systems-lab/securesystemslib/pull/288 in place we no longer need to wrap or implement key handling (generation + import) functions in an 'in_toto.util' module. This PR:
- replaces calls to 'in_toto.util' with calls to 'securesystemslib.interface' and 'securesystemslib.gpg.functions'
- adds sphinx docs for securesystemslib functions that might be useful for in-toto users
  (see demo on [in-toto-lukpueh.readthedocs.io](https://in-toto-lukpueh.readthedocs.io/en/latest/api.html#key-utilities))
- removes 'in_toto.util' and tests
- adds command line flags to 'in-toto-run', 'in-toto-record' and 'in-toto-sign' to take and/or prompt for a signing key decryption password
- cleans up tests (adds a test key generation mixin for DRY test code, removes obsolete test code)

As usual, details can be found in the commit messages.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #396, opened 2020-09-18
roadmap.md: fixing a couple of typos

didier-durand:
Hello,

fixing a couple of typos in roadmap.md

Didier

Signed-off-by: didier <durand.didier@gmail.com>

PR #413, opened 2020-11-10
Prepare v1.0.0 release

lukpueh:
### Added
* '-P/--password' (prompt) cli argument for in-toto-run/in-toto-record (#402)
* in-toto-run link command timeout setting (#367)
* API and usage documentation for cryptographic key handling with securesystemslib (#402, #408)
* Artifact recording exclude pattern documentation (#373, #405)
* Test key generation mixin (#402)
* 2021 roadmap document (#381)

### Changed
* Move 'settings' docs to new 'configuration' section and make minor enhancements in structure and content (#405)
* Update tested dependencies (#377, #383, #384, #386, #389, #390,  #394, #397, #398, #400, #404, #406)
* Debian downstream release metadata (#382)

### Removed
* 'util' crypto module in favor of securesystemslib key interfaces (#402)
* Obsolete coveralls.io API call in Travis test builds (#399)

### Fixed
* Minor docs issues (#396, #385, #395)
* pylint 2.6.0 (#387) and lgtm.com (#379) warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Changelog
 
+## in-toto v1.0.0
+
+### Added
+* '-P/--password' (prompt) cli argument for in-toto-run/in-toto-record (#402)
+* in-toto-run link command timeout setting (#367)
+* API and usage documentation for cryptographic key handling with
+  securesystemslib (#402, #408)
+* Artifact recording exclude pattern documentation (#373, #405)
+* Test key generation mixin (#402)
+* 2021 roadmap document (#381)
+
+### Changed
+* Move 'settings' docs to new 'configuration' section and make minor
+  enhancements in structure and content (#405)
+* Update tested dependencies (#377, #383, #384, #386, #389, #390,  #394, #397,
+  #398, #400, #404, #406, #409, #410, #411)
+* Debian downstream release metadata (#382)
+
+### Removed
+* 'util' crypto module in favor of securesystemslib key interfaces (#402)
+* Obsolete coveralls.io API call in Travis test builds (#399)
+
+### Fixed
+* Minor docs issues (#396, #385, #395)
+* pylint 2.6.0 (#387) and lgtm.com (#379) warnings
+
+
 ## in-toto v0.5.0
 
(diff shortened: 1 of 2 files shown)


SantiagoTorres (approved):
LGTM.

This being such a big milestone, I'd like to have a couple more eyes/approval though

lukpueh:
FYI, the last missing piecefor 1.0.0 was a straightforward and well-documented cryptographic key interface for in-toto users, which we added (and reviewed) in #402, #408.

trishankatdatadog:
Do we have documentation of parameter substitution anywhere, especially the spec?

lukpueh:
> Do we have documentation of parameter substitution anywhere, especially the spec?

Good question. It seems that with none other than the argument description in `verifylib.in_toto_verify`'s docstring which is also available on [read the docs](https://in-toto.readthedocs.io/en/latest/api.html#supply-chain-verification).

https://github.com/in-toto/in-toto/blob/a1dd78b076c7e37ce43369d089afe059ab787738/in_toto/verifylib.py#L1419-L1421

Judging from the rest of the [related code documentation](https://github.com/in-toto/in-toto/blob/a1dd78b076c7e37ce43369d089afe059ab787738/in_toto/verifylib.py#L259-L261) and the fact that it's not part of the in-toto spec, it looks like this is an experimental feature? What do others think?

trishankatdatadog:
> Judging from the rest of the [related code documentation](https://github.com/in-toto/in-toto/blob/a1dd78b076c7e37ce43369d089afe059ab787738/in_toto/verifylib.py#L259-L261) and the fact that it's not part of the in-toto spec, it looks like this is an experimental feature? What do others think?

I dunno, it's being used in production...

adityasaky (approved):

lukpueh:
@adityasaky kindly submitted a ticket to add usage docs for parameter substitution (https://github.com/in-toto/in-toto/issues/414). I am all for it, but I don't think we should block 1.0.0 on it. What do others think?

adityasaky:
> I don't think we should block 1.0.0 on it.

I agree. :)

PR #425, opened 2021-02-11
Allow running in_toto_verify with prerecorded inspection results

marmarek:
that is provided, the better.

**Fixes issue #**:

Allow to verify the layout using pre-recorded inspection results. In
practice, this is very useful with a dummy steps ("/usr/bin/true"
command) when integrating with package managers (apt, dnf, etc) when the
file hashes are already known from repository metadata. This allows to
verify the packages based on links produced by rebuilders, before
actually downloading the packages (do not waste bandwidth for invalid
packages). The package manager will later verify if downloaded
packages actually matches expected hashes.
This is especially useful for DNF (Fedora package manager), where rejecting a package based on its hash (before downloading it), is relatively easy, while rejecting already downloaded package is much harder and prone to accidentally re-use previously rejected package.

Implement this as an extra argument ("prerecorded_inspections") to the in_toto_verify() function, which consists of materials and products hashes for each step. Steps execution included in this dict will be skipped and materials/products hashes from the argument will be used
instead.

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -165,7 +165,7 @@ def load_links_for_layout(layout, link_dir_path):
   return steps_metadata
 
 
-def run_all_inspections(layout):
+def run_all_inspections(layout, prerecorded_inspections=None):
   """
   <Purpose>
     Extracts all inspections from a passed Layout's inspect field and
@@ -178,6 +178,28 @@ def run_all_inspections(layout):
   <Arguments>
     layout:
             A Layout object which is used to extract the Inspections.
+
+    prerecorded_inspections:
+            A dictionary of already performed inspections. Inspections present
+            here will not be executed and provided results will be recorded
+            instead.
+            The dictionary format is as follows:
+
+            {
+                <inspection name> : {
+                    "materials": {
+                        <material 1 filename>: {
+                            "sha256": <material 1 sha256 hash>,
+                            ...
+                        },
+                    "products": {
+                        <product 1 filename>: {
+                            "sha256": <product 1 sha256 hash>,
+                            ...
+                        },
+                    },
+                ...
+            }
 
   <Exceptions>
     Calls function that raises BadReturnValueError if an inspection returned
@@ -194,8 +216,34 @@ def run_all_inspections(layout):
     }
 
   """
+
+  if prerecorded_inspections is None:
+    prerecorded_inspections = {}
+
   inspection_links_dict = {}
   for inspection in layout.inspect:
+    if inspection.name in prerecorded_inspections:
+      LOG.info("Using pre-recorded inspection '{}' results...".format(
+          inspection.name))
+      results = prerecorded_inspections[inspection.name]
+      if "materials" not in results:
+        results["materials"] = {}
+      if "products" not in results:
+        results["products"] = {}
+
+      byproducts = {
+        "return-value": 0,
+        "stdout": "",
+        "stderr": "",
+      }
+
+      link = in_toto.models.link.Link(name=inspection.name,
+          materials=results["materials"], products=results["products"],
+          command=inspection.run, byproducts=byproducts)
+
+      inspection_links_dict[inspection.name] = Metablock(signed=link)
+      continue
+
     LOG.info("Executing command for inspection '{}'...".format(
         inspection.name))
 
@@ -1385,7 +1433,7 @@ def get_summary_link(layout, reduced_chain_link_dict, name):
 
 
 def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
-    substitution_parameters=None, step_name=""):
+    substitution_parameters=None, step_name="", prerecorded_inspections=None):
   """Performs complete in-toto supply chain verification for a final product.
 
   The verification procedure consists of the following activities, performed in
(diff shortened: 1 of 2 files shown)


marmarek:
In fact, having this I think would allow more meaningful layouts for package managers where the actual inspection is not a dummy command, but the actual command building the package. This way, the inspection materials could name the source package, the products could name binary packages and the command could be the actual build command (`dpkg-buildpackage ...`, `rpmbuild ...` etc). While the rebuilder would execute this command to provide a link file as a rebuild proof, the package manager deciding whether to install the package or not would just take hashes already known in the repository metadata (official package builds) and compare that with rebuilders data.
I'm not sure if I haven't missed anything here, but I think at least it gets closer to such state, instead of a [dummy steps in layout](https://github.com/in-toto/apt-transport-in-toto/blob/develop/data/root.layout).

Additionally, this reduce the need to create temporary files/directories for the verification purposes (like [apt-transport-in-toto does](https://github.com/in-toto/apt-transport-in-toto/blob/develop/intoto.py#L638)). It doesn't completely eliminate it yet, because link files still need to be saved there.

This is part of the work to integrate in-toto with DNF package manager. I'm opening the PR early to gather maintainers feedback, but it may make sense to not merge before @fepitre confirms it's enough for DNF integration.

fepitre:
I confirm it works in https://github.com/fepitre/dnf-plugin-in-toto at https://github.com/fepitre/dnf-plugin-in-toto/blob/master/intoto.py#L101. Thank you.

adityasaky:
Hi @marmarek, thank you for the PR! I'm only now able to take a look at it and consider the implications, so I'll get back to you ASAP.

@fepitre, if I'm not mistaken, https://github.com/fepitre/dnf-plugin-in-toto/blob/master/data/root.layout#L30 has the inspection name as `verify-reprobuilds` while the pre-recorded inspection hashes is for `rebuild`, which is actually a step in your layout. Could you confirm if this patch is actually being used in the verification?

fepitre:
> @fepitre, if I'm not mistaken, https://github.com/fepitre/dnf-plugin-in-toto/blob/master/data/root.layout#L30 has the inspection name as `verify-reprobuilds` while the pre-recorded inspection hashes is for `rebuild`, which is actually a step in your layout. Could you confirm if this patch is actually being used in the verification?

I confirm this patch work because it's mandatory. Indeed, we don't have `*.rpm` available locally but only hashes information. So I think my current layout for DNF can be simplified. At first, it was a copy paste of working one https://github.com/in-toto/apt-transport-in-toto/blob/develop/data/root.layout

lukpueh:
Thanks for the contribution, @marmarek and @fepitre! Your patch revives an interesting discussion from https://github.com/in-toto/docs/issues/27, where we looked into ways of matching the final product on the client to the "final step" in the supply chain without running a dummy inspection.

What you propose is similar to one of the solutions there (see https://github.com/in-toto/docs/issues/27#issuecomment-452046247 and https://github.com/in-toto/docs/issues/27#issuecomment-452377383). That is a "null" inspection that can be parametrized on the client. The difference between our and your proposals is that we would tell the client which artifacts to hash for the corresponding inspection link, whereas you directly pass in the pre-recorded artifact hashes. IMO both make sense for different use cases. And either way, it shows the need for command-less inspections that don't shell out to an unrelated command such as `/usr/bin/true`.

cc @JustinCappos

marmarek:
Yes, this indeed is the same topic. Please note there is another angle here too: it not only avoids running the inspection (which is in-toto/docs#27 mostly about), but also let the caller provide the products hashes to verify - instead of in-toto calculating them. In the package installation case this is very useful because you do know expected file hash before downloading the file(s) (and in fact, the current DNF plugin API allows you to reject packages only at this stage - there is no place to reject packages after downloading them).

SantiagoTorres:
I like this! the API decision also makes sense to me :+1:

SantiagoTorres (approved):

in_toto/verifylib.py
```diff
@@ -1423,6 +1471,27 @@ def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
...
+                    },
+                "products": {
+                    <product 1 filename>: {
+                        "sha256": <product 1 sha256 hash>,
+                        ...
+                    },
+                },
+            ...
+        }
+
```
lukpueh, line 1494: Please reformat this [as sphinx literal block](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) so that it renders nicely on readthedocs ...

```suggestion
        The dictionary format is as follows::

            {
                <inspection name> : {
                    "materials": {
                        <material 1 filename>: {
                            "sha256": <material 1 sha256 hash>,
                            ...
                        },
                    "products": {
                        <product 1 filename>: {
                            "sha256": <product 1 sha256 hash>,
                            ...
                        },
                    },
                ...
            }

```

lukpueh: Also, what if we just make `prerecorded_inspections` a dictionary of Metablock link objects? This would allow users to actually pass full prerecorded inspections, instead of only the materials/products aspects (maybe there is a use case for that?), and we could use existing Metablock/Link validate methods to check the inputs, and it would reduce your patch in  `run_all_inspections` to something like...

```python
    if inspection.name in prerecorded_inspections:
      LOG.info("Using pre-recorded inspection '{}' results...".format(
          inspection.name))

      inspection_links_dict[inspection.name] = \
          prerecorded_inspections[inspection.name]
      continue
```

Moreover we could then easily enhance the `in-toto-verify` cli to read prerecorded inspection link files. :)

I admit that it makes it slightly more difficult for the caller. But I think the `in_toto.models.*` provide enough tools to easily generate the corresponding objects.

lukpueh (commented):
Thanks for the very clean PR, @marmarek! We'd love to add this feature. I only have one small suggestion (see below). Let me know what you think.

lukpueh:
Is there still appetite for this feature? If not, I suggest to close the PR.

adityasaky:
Closing due to lack of activity. Please feel free to resubmit if you'd like to see this feature merged!

Issue #454, opened 2021-05-26
Exit code is 0 even on verification failure


omerlh:
Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.

**Description of issue or feature request**:

I'm testing in toto right now and noticed that the exit code is not set correctly on failure:

```
in-toto-verify --layout intoto-poc/scripts/root.layout --layout-key intoto-poc/alice.pub
Run command '['git', 'clone', '<>']' differs from expected command '['git', 'clone', **'<>']'**

echo $?
0
```

**Current behavior**:
Exit code is 0

**Expected behavior**:
Exit code should be non zero

adityasaky:
Hi @omerlh, thanks for opening the issue. Currently, when the command that is run differs from the expected command in the layout, it merely shows a warning--this is not a failure of verification. You can check if verification passed successfully in this instance regardless of commands matching using the `-v` flag, for verbosity.

When verification _does_ fail, the exit code should be non-zero.

adityasaky:
```
saky ~/.../in-toto-demo/final_product > in-toto-verify -l root.layout -k alice.pub
(in-toto-verify) RuleVerificationError: 'DISALLOW *' matched the following artifacts: ['demo-project/foo.py']
Full trace for 'expected_materials' of item 'package':
Available materials (used for queue):
['demo-project/foo.py']
Available products:
['demo-project.tar.gz']
Queue after 'MATCH demo-project/* WITH PRODUCTS FROM update-version':
['demo-project/foo.py']

saky ~/.../in-toto-demo/final_product > echo $?
1
```

This is from in-toto/demo after the supply chain has been tampered with.

trishankatdatadog:
I guess the documentation should clarify this 🙂

adityasaky:
It's part of the spec (emphasis mine):

> Finally, the "expected_command" field contains a string, COMMAND, describing the suggested command to run. It is important to mention that, in a case where a functionary’s key is compromised, this field can easily be forged (e.g., by changing the PATH environment variable in a host) and thus it should not be trusted for security checks. In addition, commands may be executed with different flags at the behest of the functionary's personal preference (e.g., a developer may run a command with --color=full or not). Because of this, during verification, **clients should only show a warning if the expected command field does not match its counterpart in the link metadata**. This warning may help auditors check whether something was out of the norm, but will not make verification fail, as these changes are not necessarily a problem.

I'm following up to see if it's also part of the documentation for the reference implementation. :smile:

[...]

PR #455, opened 2021-05-27
Mention verification does not fail if commands don't align

adityasaky:
**Fixes issue #**: Related to #454 

Adds to the documentation that failure in command alignment does not mean in-toto verification fails.

requirements**:

### DIFF
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -1406,7 +1406,10 @@ def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
   4.  Load link metadata
   5.  Verify link signatures with keys in layout
   6.  Recurse into sublayout verification
-  7.  Soft-verify alignment of reported and expected commands
+  7.  Soft-verify alignment of reported and expected commands. Note:
+      verification does not fail if commands don't match, instead a warning is
+      logged as per the in-toto specification --
+      https://github.com/in-toto/docs/blob/master/in-toto-spec.md#431-steps
   8.  Verify threshold artifact constraints
   9.  Process step product and material rules
   10. Execute inspection commands and generate inspection links


adityasaky:
@trishankatdatadog @omerlh

Please let me know if this makes things clearer. :slightly_smiling_face:

Here is the documentation with the change in this PR: https://in-toto--455.org.readthedocs.build/en/455/api.html#supply-chain-verification

trishankatdatadog:
I think you can be clearer. Maybe like a state transition diagram that depicts which states you go into? Don't know if overkill, but pls think about it

adityasaky:
Sorry, I didn't quite get that. What are the different states? As far as I can tell, this particular change only documents that command non-alignment is a warning, and not a failure of verification. Do you mean we should include a graphic for the full verification workflow?

trishankatdatadog:
Yes, which things soft-fail, which don't, and so on. There are many steps involved in the whole process. It's not clear how things can go wrong when and in which way.

omerlh:
This [comment](https://github.com/in-toto/in-toto/issues/454#issuecomment-849195031) really help me reason about the why - so maybe worth mentioning it.

adityasaky:
@omerlh I've added a reference to the specification which has that quote.

@trishankatdatadog as far as I can tell, command alignment is the only part of the verification workflow where failure results in a warning rather than a full in-toto verification failure. However, I think a graphic could be useful, also to highlight nuances about how failure of sublayouts do not necessarily lead to in-toto verification failure, but rather are subject to threshold constraints and so on. I'm currently leaning towards merging this change and opening a ticket for the workflow graphics because I think it could be a helpful task for someone new looking to better understand the verification workflow. Thoughts?

SantiagoTorres:
LGTM. Thank you! I'll merge this and re-open 454 to drop a comment soon-ish. Thank you for the discussion @omerlh, @adityasaky, @trishankatdatadog !

Issue #391, opened 2020-09-07
More test cases for exclude patterns


shibumi:
**Description of issue or feature request**:

We need to make sure, that our exclude implementation satisfies our desired functionality as specified in https://github.com/itsbenwu/in-toto/blob/a51366432a22a2bb1112b10dd07797232c23757d/doc/source/exclude-patterns.md

The exclude patterns specification mentions a few edge cases we are not testing yet:

* A slash followed by two consecutive asterisks and a slash matches path segments between two directories (e.g., a/**/b matches a/b, a/x/b, a/x/y/b and so on).
* A pattern ending with a slash will match all descendant paths. This is equivalent to {pattern}/**. If a pattern matches a directory, all files and subdirectories are also excluded.
* A hash serves as a comment, but it doesn't apply to this context. The hash can be escaped with a back-slash to match a literal hash (i.e., \#).
* An exclamation mark negates the rest of the pattern. Like the hash, this isn't particularly necessary for the context of exclude patterns. This can be escaped with a back-slash to match a literal exclamation mark (i.e., \!).
* A forward slash indicates a directory separator.
* A single / does not match any file.
* Separators at the beginning or middle (or both) of a pattern are relative to the current directory. Separators at the end of the pattern only match directories. (e.g., /frotz matches frotz/ but not doc/frotz, doc/frotz/ matches the doc/frotz/ directory but not a/doc/frotz/; however frotz/ matches frotz/ and a/frotz/).
* To match an absolute path using a pattern, a single forward slash (/) at the beginning of the pattern is not sufficient. A single forward slash is used for relative roots rather than the "actual" root found in absolute paths. Using double forward slashes will get around this (i.e., //<pattern>).
* Leading double asterisks match any preceding path segments (e.g., **/foo matches file or directory foo, and **/foo/bar matches file or directory bar anywhere that is directly under directory foo).

Furthermore we should make sure, that this functionality is operating system independent.

**Current behavior**:

We are not testing all patterns yet.

**Expected behavior**:

We should have at least one test case for each pattern

adityasaky:
Thanks for setting this up, @shibumi!

Pa3cia07:
hello, is this issue still actual?

shibumi:
@Pa3cia07 I think it is. Are you willing to contribute a patch? :D

Pa3cia07:
@shibumi I would like to try it :D

shibumi:
[...]

PR #487, opened 2022-02-04
Add more test cases for exclude patterns

rszamszur:
that is provided, the better.

**Fixes issue #391**:

requirements**:

Few notes:
> A single / does not match any file.

From my test, it seems `--exclude /` matches everything. Unless you've meant `/**`.

> To match an absolute path using a pattern, a single forward slash (/) at the beginning of the pattern is not sufficient. A single forward slash is used for relative roots rather than the "actual" root found in absolute paths. Using double forward slashes will get around this (i.e., //).

This is true only if provided `--products` and/or `--materials` paths are absolute paths as well.

Lastly, if you'd like me to add more tests cases, or I've missed one let me know.

### DIFF
--- a/tests/test_runlib.py
+++ b/tests/test_runlib.py
@@ -121,7 +121,7 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
     os.mkdir("subdir")
     os.mkdir("subdir/subsubdir")
 
-    self.full_file_path_list = ["foo", "bar", "subdir/foosub1",
+    self.full_file_path_list = ["foo", "bar", "#esc!", "subdir/foosub1",
         "subdir/foosub2", "subdir/subsubdir/foosubsub"]
 
     for path in self.full_file_path_list:
@@ -190,7 +190,7 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
 
   def test_lstrip_paths_valid_prefix_directory(self):
     lstrip_paths = ["subdir/subsubdir/"]
-    expected_artifacts = sorted(["bar", "foo", "subdir/foosub1",
+    expected_artifacts = sorted(["#esc!", "bar", "foo", "subdir/foosub1",
         "subdir/foosub2", "foosubsub"])
     artifacts_dict = record_artifacts_as_dict(["."],
         lstrip_paths=lstrip_paths)
@@ -217,7 +217,7 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
 
   def test_lstrip_paths_invalid_prefix_directory(self):
     lstrip_paths = ["not/a/directory/"]
-    expected_artifacts = sorted(["bar", "foo", "subdir/foosub1",
+    expected_artifacts = sorted(["#esc!", "bar", "foo", "subdir/foosub1",
                                  "subdir/foosub2", "subdir/subsubdir/foosubsub"])
     artifacts_dict = record_artifacts_as_dict(["."],
         lstrip_paths=lstrip_paths)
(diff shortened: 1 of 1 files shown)


rszamszur:
I guess the last pattern format to be tested is:

> To match an absolute path using a pattern, a single forward slash (/) at the beginning of the pattern is not sufficient. A single forward slash is used for relative roots rather than the "actual" root found in absolute paths. Using double forward slashes will get around this (i.e., //<pattern>).

However, I'm confused about how it is supposed to work exactly? The only configuration which works for me is if both `--exclude` and `--product/--materials` are absolute paths with `//`. Also, how this pattern applies to Windows paths?

adityasaky:
Hi @rszamszur, thanks for picking this ticket up! I'll try and review it soon, apologies for the delay.

tests/test_in_toto_run.py
```diff
@@ -129,6 +128,88 @@ def test_main_optional_args(self):
     self.assertListEqual(list(link_metadata.signed.products.keys()),
         [self.test_artifact[len(strip_prefix):]])
 
+  def test_main_exclude_patterns(self):
```
adityasaky: Instead of adding these test cases to test_in_toto_run.py which primarily tests the CLI, we can use them in test_runlib.py. https://github.com/in-toto/in-toto/blob/develop/tests/test_runlib.py#L415

tests/test_in_toto_run.py
```diff
@@ -142,7 +223,6 @@ def test_main_with_metadata_directory(self):
 
     self.assertTrue(os.path.exists(linkpath))
 
-
```
adityasaky: Guessing editor shenanigans were the cause here and elsewhere.

tests/test_in_toto_run.py
```diff
@@ -129,6 +128,88 @@ def test_main_optional_args(self):
...
+    Path("doc/frotz/test_artifact").touch()
+    Path("a").mkdir()
+    Path("a/doc").mkdir()
+    Path("a/doc/frotz").mkdir()
+    Path("a/doc/frotz/test_artifact").touch()
+
+    # Define exclude pattern test cases. Each test case is a tuple with three
+    # values: items, pattern, and expected result.
+    test_cases = [
+      (self.test_artifact, "/**/test_artifact", {}),
```
adityasaky: Can we add a test case for:

```A single / does not match any file.```

**Review by adityasaky — CHANGES_REQUESTED** [review]
Hi @rszamszur, the test cases look good to me. However, instead of testing the CLI, I recommend testing the API. I've left a link below.

File: tests/test_in_toto_run.py (hunk 2 above)
Comment (rszamszur) [reply tests/test_in_toto_run.py]: Actually, this was intentional, PEP8 E303 (Method definitions inside a class are surrounded by a single blank line.)
* https://www.python.org/dev/peps/pep-0008/#blank-lines
* https://www.flake8rules.com/rules/E303.html

I thought while changing this file I might as well improve PEP style guide. But I can revert this, your call.

File: tests/test_in_toto_run.py (hunk 3 above)
Comment (rszamszur) [reply tests/test_in_toto_run.py]: From my tests `/` matched every file. But maybe I did something wrong, or it's a defect.

image

File: tests/test_in_toto_run.py (hunk 1 above)
Comment (rszamszur) [reply tests/test_in_toto_run.py]: Sure thing, already on it!

File: tests/test_in_toto_run.py (hunk 3 above)
Comment (adityasaky) [reply tests/test_in_toto_run.py]:
```suggestion
      ("a", "/", {
        "a/doc/frotz/test_artifact": {"sha256": sha256},
      }),
      (self.test_artifact, "/**/test_artifact", {}),
```

This passes for me. IIRC, pathspec has some of its own edge cases that are likely coming into play here. That's the kind of stuff that led us to working on the reference documentation. Please let me know if I've made a silly mistake here. :)

rszamszur: I have misled myself :man_facepalming: :man_facepalming: :man_facepalming:. 

In the initial exclude test cases implementation I was asserting "opposite" values: 
Files that will be matched by exclude pattern VS files that will remain, hence the confusion. In reality, it works properly and I just was asserting the opposite expected output.

tests/test_in_toto_run.py
rszamszur: @adityasaky do you want me to revert those extra blank lines as it was before, or it can stay that way?

tests/test_runlib.py
```diff
@@ -416,14 +416,61 @@ def test_exclude_patterns(self):
...
+      (
+        ["/subdir/subsubdir"],
+        ["foo", "bar", "#esc!", "subdir/foosub1", "subdir/foosub2"],
+      ),
+      (
+        ["subdir/subsubdir/"],
+        ["foo", "bar", "#esc!", "subdir/foosub1", "subdir/foosub2"],
+      ),
+      (
+        ["\#esc*"],
```
```suggestion
        ["\#esc*"],  # pylint: disable=W1401
```

tests/test_runlib.py
```diff
@@ -416,14 +416,61 @@ def test_exclude_patterns(self):
...
+        ["foo", "bar", "subdir/foosub1", "subdir/foosub2",
+            "subdir/subsubdir/foosubsub"],
+      ),
+      (
+        ["*esc\!"],
```
```suggestion
        ["*esc\!"],  # pylint: disable=W1401
```

tests/test_in_toto_run.py
adityasaky: Yes please, we typically prefer to keep PRs for one task only. And for things like style changes, we prefer to leave them for when there's a rewrite in that particular location anyway.

adityasaky (changes requested):
This is almost good to go, just a minor fix to make the linter happy. Also, can I ask you to squash your commits? We prefer to rebase rather than introduce merge commits in feature branches. :)

tests/test_runlib.py
adityasaky: Alternatively this could be a raw string I suppose.

adityasaky:
Thanks, @rszamszur!

PR #490, opened 2022-03-03
Update package metadata

ofek:

## Background

Hello again! The Python packaging ecosystem has now standardized on the interface for build backends ([PEP 517](https://www.python.org/dev/peps/pep-0517/)/[PEP 660](https://www.python.org/dev/peps/pep-0660/)) and the format for metadata declaration ([PEP 621](https://www.python.org/dev/peps/pep-0621/)/[PEP 631](https://www.python.org/dev/peps/pep-0631/)).

As such, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

## Summary of changes

This implements [PEP 621](https://www.python.org/dev/peps/pep-0621/), obviating the need for `setup.py` and `MANIFEST.in`. Support has not landed in `setuptools`, so builds will now use `hatchling`. It is quite stable and actively maintained, the only reason the version is 0.x is because 1.0.0 will drop support for Python 2. It's also on the major distribution channels such as [conda-forge](https://anaconda.org/conda-forge/hatchling), [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), etc.

I've done this for such projects as [pipx](https://github.com/pypa/pipx/blob/2ecc668acf472ad6956cc682499c077c1130d17e/pyproject.toml), all [Datadog Agent integrations](https://github.com/DataDog/integrations-core/pulls?q=is%3Apr+author%3Aofek+in%3Atitle+Add+pyproject.toml+file), etc.

## Notes

- A RTD config was required https://readthedocs.org/projects/in-toto/builds/16252595/
- In future, we should remove the Python upper bound constraint i.e. change `~3.7` to `>=3.7`. Upper bounds for libraries are heavily discouraged by the community, see https://iscinumpy.dev/post/bound-version-constraints/


trishankatdatadog:
This is great! Could you also please help fix this in https://github.com/theupdateframework/python-tuf?

Cc @jku @joshuagl

jku:
> Could you also please help fix this in https://github.com/theupdateframework/python-tuf?

python-tuf build is fairly modern already. We do still use setuptools... but would likely need good reasons to move to a less  known build tool: I'm not trying to belittle hatchling, I'm just saying we try to be careful about pipeline security. Getting rid of MANIFEST.in might not be a good enough reason to make a major build tool change.

A comment on removing setup.py (in case the project uses dependabot): Dependabot can silently break if you do that: https://github.com/dependabot/dependabot-core/issues/4483. This is why python-tuf currently has an essentially empty setup.py.

> It is quite stable and actively maintained

I suppose the code base may have longer history but I think it's worth mentioning that your git history goes back to December 2021 and your first non-yanked releases happened in 2022.

ofek:
> need good reasons to move

I can briefly tell you why we moved at Datadog!

Our goal is to remove all `setup.py` files since their use is now [heavily discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) and build with a [PEP 517](https://www.python.org/dev/peps/pep-0517/) compatible build system while defining metadata with [PEP 621](https://www.python.org/dev/peps/pep-0621/).

### setuptools

Disadvantages:

- Does not yet support `pyproject.toml`-based metadata declaration nor editable installation standards ([PEP 660](https://www.python.org/dev/peps/pep-0660/))
- Based on a [deprecated](https://www.python.org/dev/peps/pep-0632/) standard library module that it now vendors
- Slow to get things merged since the [code quality issue](https://github.com/pypa/pipx/pull/817#issuecomment-1050255755) is exacerbated by the fact that it is mostly maintained by just [one person](https://github.com/jaraco)

### Flit

[`flit-core`](https://pypi.org/project/flit-core/) is the build backend behind [Flit](https://github.com/pypa/flit).

Disadvantages:

- No mechanism for extending behavior by design, so we could never do things like versioning with Git tags
- Very opinionated about project layouts since it targets beginners e.g. [only supports](https://github.com/pypa/flit/issues/370#issuecomment-973359998) implicit namespace packages, must define certain data like `__version__` inside the root `__init__.py` file, etc.

### Poetry

[`poetry-core`](https://pypi.org/project/poetry-core/) is the build backend behind [Poetry](https://github.com/python-poetry/poetry).

Disadvantages:

- Persistent unwillingness to adopt new standards if they are deemed suboptimal (see comments on [PEP 621](https://discuss.python.org/t/5243/12) and [PEP 665](https://github.com/python-poetry/poetry/issues/4710#issuecomment-973946104))

### Hatchling

- In comparison to `setuptools`, Hatchling is far smaller, making it less susceptible to security issues and more auditable. For example, the logic to build a wheel resides in this [one file](https://github.com/ofek/hatch/blob/master/backend/src/hatchling/builders/wheel.py).
- In comparison to `flit-core`, Hatchling is extensible by design. For example, there is already [a plugin](https://github.com/ofek/hatch-vcs) for [setuptools_scm](https://github.com/pypa/setuptools_scm) (which despite the legacy naming is actually now decoupled from `setuptools`) and [a plugin](https://github.com/ofek/hatch-mypyc) for compiling code with [Mypyc](https://github.com/mypyc/mypyc).
- In comparison to `poetry-core`, Hatchling never strays from community standards and is in fact eager to adopt new ones (it's the only backend that currently supports [PEP 639](https://www.python.org/dev/peps/pep-0639/)).

-----

> This is why python-tuf currently has an essentially empty setup.py.

I can do that if we want!

adityasaky:
I like this. One thing though is that we need to get hatch/hatchling into Arch `community` minimum before we adopt this. I see an orphaned AUR package (https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-hatch) that I'm happy to adopt, but that's not enough for in-toto to have a makedependency. cc @SantiagoTorres @shibumi can you help us out here?

adityasaky (approved):

SantiagoTorres:
Happy to take a look!

pyproject.toml
```diff
@@ -0,0 +1,87 @@
...
+description = "A framework to define and secure the integrity of software supply chains"
+readme = "README.md"
+license = "Apache-2.0"
+requires-python = "~=3.7"
+authors = [
+    { name = "New York University: Secure Systems Lab", email = "in-toto-dev@googlegroups.com" },
+]
+maintainers = [
+    { name = "Justin Cappos", email = "jcappos@nyu.edu" },
+    { name = "Santiago Torres", email = "santiago@nyu.edu" },
```
SantiagoTorres: maybe this would be a good time to update my email :)

adityasaky:
```suggestion
    { name = "Santiago Torres", email = "santiagotorres@purdue.edu" },
```

ofek:
As requested, I updated Santiago's email and added back an empty `setup.py` for Dependabot 🙂

trishankatdatadog:
python-tuf 🥺 

Feeling left out...

ofek:
@trishankatdatadog https://github.com/theupdateframework/python-tuf/pull/1896 😄

ofek:
@SantiagoTorres Did you happen to get a chance to look into adding `hatchling` to Arch Linux?

adityasaky:
We're working on getting the package added to Arch, @ofek. Apologies for the lack of updates! This _should_ be good to go after we know for sure everything works, I reckon.

ofek:
@adityasaky That's great, thanks! Where would that be located?

BTW I just rebased (bumping minimum `hatchling` to latest) and https://github.com/theupdateframework/python-tuf/pull/1896 has been merged 🙂

SantiagoTorres:
Hi, i think stuff is ready to go, but admittedly I don't understand how hatch fits in the pipeline. Would this require in-toto have hatch as a runtime dependency or just a build dependency (i.e., like tox)? If so, can we get away with building without hatch? I just want to make sure I fully understand this dep :)

SantiagoTorres:
Actually, reading with a little bit more patience, I think I understand. things better. I was just worried it came with a command to install conda, which would break a couple of rules on our side... 

[here it is!](https://archlinux.org/packages/community/any/python-hatch/)

SantiagoTorres:
@adityasaky is this all we need? we'll have to update the PKGBUILD for in-toto but that doesn't seem like a lot of trouble...

adityasaky (approved):

adityasaky:
I think this should be good to go! Did you have a chance to try the PKGBUILD changes for in-toto?

ofek:
@SantiagoTorres Hey!

There are 2 errors there:

1. `hatch` the CLI was rewritten and is at v1.0.0rc16 https://pypi.org/project/hatch/1.0.0rc16/
2. `hatchling` the build backend is what is required, not the CLI https://pypi.org/project/hatchling/0.21.1/

ofek:
Hatchling would look similar to:

- https://archlinux.org/packages/community/any/python-poetry-core/
- https://archlinux.org/packages/community/any/python-flit-core/

SantiagoTorres:
huh, @ofek the hatch badge says the latest version is 0.23.1? is that a mistake or should I be packaging 1.0.0? Overall, we avoid packaging rc's

ofek:
PyPI badge shows latest non-RC. No worries though, all you need is `hatchling`.

SantiagoTorres:
putting my arch hat on. Would this versions agree (hatch and hatching) in the future or nah? Just to future-proof things...

ofek:
If you mean do they have the same release cycle then no they do not 👍

SantiagoTorres:
:crying_cat_face:  Ok!

SantiagoTorres:
@ofek are we ready to go [now](https://archlinux.org/packages/community/any/python-hatchling/) ?

ofek:
@SantiagoTorres Nice! Are the deps defined? https://github.com/ofek/hatch/blob/hatchling-v0.21.1/backend/src/hatchling/ouroboros.py#L34-L44

SantiagoTorres:
Ugh, not really. I need to fix these now, as I assume hatch depends on hatchling as well..

Of the ones listed, I think we don't host  editables, do you know what this is used for?  could this one be optional? for some reason it didn't complain when I tested things...

ofek:
That's only used for `pip install -e ...`

SantiagoTorres:
oh, then I think I'll go ahead and ommit it. Cheers!

PR #491, opened 2022-03-11
[W.I.P][ITE-4] Add Github resolver

SolidifiedRay:
*This is a draft PR for the ITE-4 github use case:*
https://github.com/in-toto/ITE/blob/master/ITE/4/README.adoc#use-case-2-github-entities

Github has many abstract entities such as Pull Requests and Commits. This PR helps provide attestationns about these artifacts by getting the hash from the Github entities.

*For more discussion on what data we should be capturing, please refer to this link: https://docs.google.com/document/d/14dZ-KKJcz2Uscygb6n02zcsRHRIGQS6jhWtoHWVqU7M/edit?usp=sharing

### DIFF
--- a/in_toto/resolvers/github.py
+++ b/in_toto/resolvers/github.py
@@ -1,0 +1,143 @@
+"""
+ITE-4:
+Github has many abstract entities such as Pull Requests and Commits. This code
+helps provide attestationns about these artifacts by getting the hash from the
+Github entities.
+"""
+
+from urllib3.util import Url
+from urllib3 import PoolManager
+import json
+import hashlib
+
+
+def _get_resolvable_url(generic_uri):
+  """
+  Convert a generic uri to an API URL that represent a Github entity
+
+  We should consider what generic_url looks like. Here is a draft idea of what
+  it should look like.
+    github:org/repo:pr:number
+    github:org/repo:commit:id
+
+  Args:
+    generic_uri: a generic_uri representing a GitHub entity
+
+  Returns:
+    an API URL representing a GitHub entity
+  """
+  uri_split = generic_uri.split(':')
+
+  if uri_split[2] == 'pr':
+    path = 'repos/{}/pulls/{}'.format(uri_split[1], uri_split[3])
+  elif uri_split[2] == 'commit':
+    path = 'repos/{}/commits/{}'.format(uri_split[1], uri_split[3])
+
+  resolvable_url = Url(scheme='https', host='api.github.com', path=path)
+
+  return resolvable_url
+
+
+def _hash_review_representation(review):
+  '''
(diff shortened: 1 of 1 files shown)


in_toto/resolvers/github.py
```diff
@@ -0,0 +1,37 @@
+from urllib3.util import Url
+from urllib3 import PoolManager
+import json
+import hashlib
+
+def get_resolvable_url(generic_url):
+    #TODO: get resolvable url for commits, release, etc.
```
adityasaky: We should consider what the generic_url looks like. The format is defined by what the resolver expects, essentially, and it can't be just the path of the API. There's some guidance regarding this in ITE-4.

in_toto/resolvers/github.py
```diff
@@ -0,0 +1,37 @@
...
+    representation_object['base'] = response_data['base']['label']
+    representation_object['commits_count'] = response_data['commits']
+    representation_object['merge_commit_sha'] = response_data['merge_commit_sha']
+    representation_object['additions'] = response_data['additions']
+    representation_object['deletions'] = response_data['deletions']
+    representation_object['changed_files'] = response_data['changed_files']
+
+    return representation_object
+
+def get_hash_artifacts(generic_url):
```
adityasaky:
```suggestion
def hash_artifacts(generic_url):
```

in_toto/resolvers/github.py
```diff
@@ -0,0 +1,37 @@
...
+    resolvable_url = Url(scheme='https', host='api.github.com', path=generic_url)
+    return resolvable_url
+
+def get_hashable_representation(generic_url):
+    http = PoolManager()
+    resolvable_url = get_resolvable_url(generic_url)
+    response = http.request('GET', str(resolvable_url), headers={'User-Agent': 'in-toto Reference Implementation'})
+    response_data = json.loads(response.data)
+
+    representation_object = {}
```
adityasaky: We should discuss (perhaps synchronously) the fields recorded here, and how / why they're useful wrt the use cases.

adityasaky: This also brings up the question of enabling _some_ user input in deciding what data must be captured, though that's likely a larger, ITE-4 related discussion. The easy way out may be allowing the hashable_representation to have certain optional fields that may be unused.

SolidifiedRay: A link to the discussion: https://docs.google.com/document/d/14dZ-KKJcz2Uscygb6n02zcsRHRIGQS6jhWtoHWVqU7M/edit#heading=h.agahzfc87zba

adityasaky:
Closing this. We have a new module for resolvers, perhaps we could have this resubmitted using that? https://github.com/in-toto/in-toto/blob/develop/in_toto/resolver/_resolver.py

Issue #445, opened 2021-04-12
Add support for DSSE


adityasaky:
**Description of issue or feature request**:
The new signature wrapper [DSSE](https://github.com/secure-systems-lab/dsse) removes the in-toto's current dependence on canonicalization, and is also open to encodings other than JSON. Moving the signature wrapper definition out of the in-toto specification will also help it focus on its key goals, making it more readable. The switch to the new specification is documented in [ITE-5](https://github.com/in-toto/ITE/pull/13).

**Current behavior**: The reference implementation has no support for DSSE.

**Expected behavior**: The reference implementation must be capable of generating metadata using DSSE as an alternative to the current in-toto document format. In order to do this, the reference implementation must include a DSSE implementation.

adityasaky:
The current document format is shared with our sister project, TUF. The implementation for both projects is provided through [securesystemslib](https://github.com/secure-systems-lab/securesystemslib). The tooling for DSSE implemented can be prototyped here, in in-toto, before deciding how to integrate it with securesystemslib for TUF and other projects to take advantage of.

attreyee-muk:
Hello! I would like to contribute to this project for GSOC'22. Can you please help me to get started with it?

adityasaky:
Hi @a-muk, sure thing!

in-toto is a framework that generates metadata files which can be used to make certain verifiable claims about processes in the software supply chain. This metadata is cryptographically signed and represented using what we call a signature wrapper or envelope. The current signature wrapper is defined in the [in-toto specification](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#42-file-formats-general-principles), with the model defined [here](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/metadata.py#L39) in this implementation. Last year, we decided to retire this signature wrapper in favour of [DSSE](https://github.com/secure-systems-lab/dsse). So, the task detailed here is adding capabilities to this repository to generate in-toto metadata using DSSE as the wrapper rather than `Metablock` I've pointed to above.

Please let me know if you have any more questions. If you'd like to see current in-toto metadata represented using `Metablock`, I recommend playing with the [in-toto demo](https://github.com/in-toto/demo). You can compare their structure with the DSSE specification to see how they differ. Finally, feel free to join #in-toto on the [CNCF slack workspace](https://slack.cncf.io).

attreyee-muk:
Thank you @adityasaky for the overview.

PradyumnaKrishna:
[...]

Issue #272, opened 2019-04-11
Gather infos to show how in-toto follows secure design principles


lukpueh:
**Description of issue or feature request**:
[CII silver-level best practices](https://bestpractices.coreinfrastructure.org) requires projects 
>  to implement secure design principles (from "know_secure_design"), where applicable.

[["known_secure_design"]](https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md#know_secure_design)

**Current behavior**:
Required information missing.

**Expected behavior**:
Gather anecdotal evidence (descriptions, links) that show how the in-toto reference implementation follows above mentioned secure design principles.

lukpueh:
Fixed by pointing out the CS background of the primary developers, the governance process, and the secure design practices outlined in the CNCF/sig-security self-assessment.

PR #503, opened 2022-07-29
Add DSSE support for in-toto

PradyumnaKrishna:
**Fixes issue #445**: Add support for DSSE 

This pull request aims to add DSSE support for in-toto, as DSSE signature wrapper is completed in securesystemslib.
I started with adding DSSE support in `in_toto_run` and created a convertor that converts `Link` or `Layout` object to DSSE Envelope.

requirements**:


in_toto/runlib.py
```diff
@@ -571,31 +574,67 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
       materials=materials_dict, products=products_dict, command=link_cmd_args,
       byproducts=byproducts, environment=environment)
 
-  link_metadata = Metablock(signed=link, compact_json=compact_json)
+  if use_dsse:
+    link_metadata = link.create_envelope()
```
adityasaky: Let's log the selected signature wrapper here.

in_toto/runlib.py
```diff
@@ -571,31 +574,67 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
...
+      metadata_dict = link_metadata.to_dict()
+
+      indent = None if compact_json else 1
+      separators = (',', ':') if compact_json else (',', ': ')
+
+      with open(filename, 'w', encoding='utf8') as fp:
+        json.dump(metadata_dict, fp, indent=indent, separators=separators)
+
+  else:
+    link_metadata = Metablock(signed=link, compact_json=compact_json)
```
adityasaky, line 673: Same here.

in_toto/runlib.py
```diff
@@ -571,31 +574,67 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
...
+      if metadata_directory is not None:
+        filename = os.path.join(metadata_directory, filename)
+
+      LOG.info("Storing link metadata to '{}'...".format(filename))
+      metadata_dict = link_metadata.to_dict()
+
+      indent = None if compact_json else 1
+      separators = (',', ':') if compact_json else (',', ': ')
+
+      with open(filename, 'w', encoding='utf8') as fp:
```
adityasaky: Can this be offloaded to a `dump` method in the DSSE envelope?

in_toto/runlib.py
```diff
@@ -571,31 +574,67 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
...
-      filename = os.path.join(metadata_directory, filename)
+      if metadata_directory is not None:
+        filename = os.path.join(metadata_directory, filename)
 
-    LOG.info("Storing link metadata to '{}'...".format(filename))
-    link_metadata.dump(filename)
+      LOG.info("Storing link metadata to '{}'...".format(filename))
+      link_metadata.dump(filename)
 
   return link_metadata
```
adityasaky: I think this should be straightforward to implement for `in_toto_record_*` and `in_toto_mock` as well!

adityasaky (commented):
This is looking good!

in_toto/runlib.py
PradyumnaKrishna: This is possible but Metadata wrapper for python-tuf uses different naming convention, `to_file` and `from_file`. In my opinion either this should be done in the in-toto itself or create a method that is suitable for both projects.

adityasaky: I don't have a strong opinion vis-à-vis the naming convention, especially since the eventual plan for `Metablock` is to deprecate it. I think it makes sense for the DSSE library to implement it rather than the implementation extending it.

in_toto/runlib.py
```diff
@@ -429,7 +431,8 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
     record_streams=False, signing_key=None, gpg_keyid=None,
     gpg_use_default=False, gpg_home=None, exclude_patterns=None,
     base_path=None, compact_json=False, record_environment=False,
-    normalize_line_endings=False, lstrip_paths=None, metadata_directory=None):
+    normalize_line_endings=False, lstrip_paths=None, metadata_directory=None,
+    use_dsse=False):
```
lukpueh, line 521: Please document `use_dsse` argument in docstring.

lukpueh (commented):
This is great! I appreciate how non-invasive your patch is, but the interface and its implementation becomes more and more convoluted. I think we should seize the opportunity of introducing DSSE -- I assume it will become an API breaking change -- and clean up a little.

I'd like us to take inspiration from the latest developments in `python-tuf` and `securesystemslib` wrt metadata-related abstractions. Most notably (plus some ideas):

- [`securesystemslib.signer.Signer`](https://github.com/secure-systems-lab/securesystemslib/blob/2bec9299a01bac03de9b907b3098a7edcc773eee/securesystemslib/signer.py#L140) -- DSSE envelope already uses this, we could make `Metablock.sign` use it too. I would even suggest to remove `signing_key`, `gpg_keyid`, `gpg_use_default`, `gpg_home` args from `in_toto_run` and have it accept a `Signer` instead.

- [`tuf.api.metadata.Signed`](https://github.com/theupdateframework/python-tuf/blob/3108998f752a8c454b071540c0de11b4f7e359e9/tuf/api/metadata.py) -- Base class for inner tuf metadata, akin to in-toto's `Signable`. We could port it to securesystemslib and use it for any inner tuf or in-toto metadata, regardless of the signature wrapper.

- [`tuf.api.metadata.Metadata`](https://github.com/theupdateframework/python-tuf/blob/3108998f752a8c454b071540c0de11b4f7e359e9/tuf/api/metadata.py#L86) --  Container class for traditional signature wrapper, akin to in-toto's Metablock, but more modern and flexible, e.g. it uses the `Signer` interface and other useful abstractions like `StorageBackendInterface`. It would be nice to port this to securesystemslib and use it for TUF and in-toto.

- [`tuf.api.serialization`](https://github.com/theupdateframework/python-tuf/tree/3108998f752a8c454b071540c0de11b4f7e359e9/tuf/api/serialization) -- Provides abstract interfaces and default implementations to de/serialize different wireline formats for traditional metadata  (see `Metadata[Des|S]erializer`), and to serialize payload (see `SignedSerializer`). I wonder if we can extend/reuse/replicate this module for DSSE.

in_toto/runlib.py
lukpueh: I suggest to hold off porting this feature to `in_toto_record` and `in_toto_mock` until the metadata architecture has crystallized (see https://github.com/in-toto/in-toto/pull/503#pullrequestreview-1058361691)

PradyumnaKrishna:
@lukpueh thank you for providing some thoughts.

Here is my opinion about them, and I think we can continue with the `Signer` implementation and review again.

* **`Signer`**
  * We can make `Metablock.sign` to use this and it will reduce the duplication from the implementation.
  * Accepting `Signer` in in_toto_run wouldn't make a major impact because somewhere we have to parse command line arguments and convert them into required `Signer`.

* **`Signable` and TUF `Metadata`**
  * TUF metadata do not support GPGSignatures, and it would be tricky to add them to it because tuf metadata works on different type of Key implementation.
  * In future, in-toto will deprecate[^1] its current metadata wrapper (old-style envelope) because of it's flaw. Porting the metadata would be unnecessary job.
  * Lets say if we port the metadata wrapper to securesystemslib, then we have to change all the Metablock references to metadata, and that will be a major and breaking change for in-toto. Doing that is similar to integrating DSSE in in-toto.
  * Porting Signed is not effective, because in the end `Signed` is a `Link` or `Layout` and there will be no changes.
  * I believe in abstraction between metadata wrappers, lets think about it more.

* **`Serialization`**
The serialization will be inetegrated in both wrappers (DSSE and Metablock) and will be achieved easily.

I will push a patch soon, maybe tomorrow. After that we can discuss this again.

[^1]: https://github.com/in-toto/ITE/blob/master/ITE/5/README.adoc#backwards-compatibility

in_toto/models/metadata.py
```diff
@@ -137,14 +139,15 @@ def type_(self):
     return self.signed.type_
 
 
-  def sign(self, key):
+  def sign(self, key=None, signer: Signer=None):
```
lukpueh: I appreciate the effort to maintain backwards compatibility, but I'm more in favor of maintaining a clean API.

My preferred solution would be to replace the current `sign(key: dict)` and `sign_gpg(keyid: str)` with a `sign(signer: Signer)` and maybe add a `isinstance(signer, Signer)` check with an error message that points out the breaking change. We can remove/reword the error message after a few releases.

If we don't want to break the API (yet), we can also add a new sign method with a different name and deprecate current `sign` and `sign_gpg` more slowly. 

In https://github.com/theupdateframework/python-tuf/pull/2010#issuecomment-1148363668 we had a similar problem. There we ended up with a mix of above two suggestion, i.e. create a new function with a different name and remove the old ones right away. The reason why we changed the name was that it was not as straight-forward as here to point out to the user that the function was called with the old argument types.

in_toto/models/metadata.py
```diff
@@ -282,3 +291,28 @@ def _validate_signatures(self):
...
+
+
+def dump_metadata(path, metadata, compact_json=False):
+  """A temporary function that dumps metadata json into a file."""
+
+  indent = 1
+  separators = (',', ': ')
+  if compact_json:
+    indent = None
+    separators = (',', ':')
```
lukpueh: Note that `Metablock` also has a `compact_json` attribute, which is ignored here. This should really only be a temporary solution. Adding a TODO comment or creating a GitHub issue might help to not forget.

lukpueh (commented):
Cool stuff! `in_toto_run` looks a lot better now

in_toto/models/metadata.py
lukpueh: cc @SantiagoTorres @adityasaky

lukpueh: Maybe we don't need to worry too much about the signing interface of `Metablock` if we decide to henceforth only support creation of DSSE envelopes.

lukpueh:
Thank you for your assessment, @PradyumnaKrishna! I have some replies inline

> * **`Signer`**
>   
>   * We can make `Metablock.sign` to use this and it will reduce the duplication from the implementation.

See https://github.com/in-toto/in-toto/pull/503#discussion_r937614342. 

>   * Accepting `Signer` in in_toto_run wouldn't make a major impact because somewhere we have to parse command line arguments and convert them into required `Signer`.

True. But it seems like a job that we could factor out to something like a Signer factory that's called in the cli. Seems like it would make the library more ergonomic. But we don't have to do this now.

> * **`Signable` and TUF `Metadata`**
>   
>   * TUF metadata do not support GPGSignatures, and it would be tricky to add them to it because tuf metadata works on different type of Key implementation.

TUF's `Metadata` class already uses securesystemslib's  `Signer` and `Signature` objects, and it makes sense for it to use the `Key` implementation that you ported to securesystemslib instead of it's own. This would allow `Metadata` to support gpg signatures. 

Moreover, there is a feature request to port the entire `Metadata` class to securesystemlib  and use it for both TUF and in-toto (see https://github.com/secure-systems-lab/securesystemslib/issues/272#issuecomment-845256237).

But again, we don't have to do this right away. So I'm fine if we work with Metablock for the time being.

>   * In future, in-toto will deprecate[1](#user-content-fn-1-3bf6e0cfaebc0b6fbeecd633d3d0a082) its current metadata wrapper (old-style envelope) because of it's flaw. Porting the metadata would be unnecessary job.

True. But even if we no longer support creating metadata with the traditional wrapper, we will probably need to support reading it and verifying its signatures for a while. So it might make sense to have a common class in securesystemslib.

>   * Lets say if we port the metadata wrapper to securesystemslib, then we have to change all the Metablock references to metadata, and that will be a major and breaking change for in-toto. Doing that is similar to integrating DSSE in in-toto.

True. And we should carefully consider whether we want a breaking change or not. I think it's okay to break the API, as long as we stay backwards compatible with old metadata, i.e. support reading and verifying the traditional metadata wrapper. 

>   * Porting Signed is not effective, because in the end `Signed` is a `Link` or `Layout` and there will be no changes.

It makes sense if we have a common Metadata class in securesystemslib that contains a generic `Signed` object and provides methods to serialize/deserialize, sign/verify, etc. We could subclass the same Signed class then in in-toto (Link, Layout) and tuf (Root, Targets, Snapshot, Timestamp).

>   * I believe in abstraction between metadata wrappers, lets think about it more.

👍

adityasaky:
Note: we want to be able to plug the same `Link` object into both signature wrappers (and sign them with the same keys) to meet the requirements of the transition phase. https://github.com/in-toto/ITE/blob/master/ITE/5/README.adoc#backwards-compatibility

in_toto/models/metadata.py
```diff
@@ -79,6 +85,44 @@ def __repr__(self):
...
+
+
+  def to_dict(self) -> dict:
+    """Returns the JSON-serializable dictionary representation of self."""
+
+    return {
+      "signatures": self.signatures,
+      "signed": attr.asdict(self.signed)
+    }
+
```
lukpueh, line 281: I assume you didn't want to break the interface (too much), which is fair, but there is a bit of redundant functionality in `Metablock` now. I'm fine leaving as is and decide later if we want to just sit out the DSSE transition phase and remove the whole Metablock class at once, or clean it up in a breaking change before.

Still, would you mind creating an issue for the following items?

- `JSONSerializer` can be used instead of `Metablock.__repr__` and `Metablock.compact_json`
- `Metablock.from_file` and `Metadata.to_file` (both via `SerializationMixin`) can be used instead of `Metadata.load` and `Metablock.dump`
- I don't think we need `attr` anymore on `Metablock`

in_toto/models/metadata.py
```diff
@@ -137,7 +169,32 @@ def type_(self):
...
+      The signature. A securesystemslib.signer.Signature type.
+
+    """
+    signature = signer.sign(self.signed.signable_bytes)
+
+    self.signatures.append(signature.to_dict())
+
+    return signature
+
+  def sign_key(self, key):
```
lukpueh: I am aware that I said this was my preferred solution, but it now looks like this is the only breaking change of your DSSE integration.  Maybe we can leave `Metablock.sign` as it was and find a different name for a signing function that accepts a `Signer`? You then won't be able to call `metablock_or_envelope.sign()` anymore, but it seems easier to also rename `Envelope.sign`, than to adopt a breaking change in `Metablock`.

We only need to find a good name. 😬  What about `create_signature`, or `create_sig`? The former clashes a bit with the old `verify_signature`. But if you choose the latter, you could call your new `verify` function `verify_sig` or so... I'm very open for better naming suggestions.

in_toto/models/common.py
```diff
@@ -61,3 +63,18 @@ def signable_bytes(self):
...
+      payload=self.signable_bytes,
+      payload_type="application/vnd.in-toto+json",
+      signatures=[]
+    )
+
+  @classmethod
+  def from_dict(cls, data: dict):
+    """Creates a Signable from its JSON/dict representation."""
+
+    return cls(**data)
```
lukpueh: `from_dict` should never be called directly on `Signable`, because we don't want pure `Signable` objects. And it does not  work as expected when called on a `Layout` child class, because layout object contain other objects that should be parsed from their dict representation. What we want here is already implemented as `Layout.read` and `Link.read`.

lukpueh: Looks like this is currently used when calling `Envelope.deserialize_payload()` with the default `JSONDeserializer`, which expects the class to be a `JSONSerializable` and thus implement `from_dict`, right?  If we checked types in in-toto this should fail, because neither `Link` nor `Layout` is a `JSONSerializable`. Plus the wrong `Layout.from_dict` behavior described above.

I suggest to make `Link` and `Layout` both inherit from `JSONSerializable` and implement `from_dict` in both classes. They can just call the already existing `read` methods. Or, if we are fine with breaking changes, rename them.

in_toto/models/metadata.py
```diff
@@ -282,3 +382,26 @@ def _validate_signatures(self):
 
     for signature in self.signatures:
       securesystemslib.formats.ANY_SIGNATURE_SCHEMA.check_match(signature)
+
+
+# pylint: disable-next=abstract-method
+class MetadataLoader(SerializationMixin, JSONSerializable):
+  """A Metadata loader to load metadata using suitable method."""
```
lukpueh: With in-toto verification in mind, does it make sense to provide a more general metadata abstraction for traditional `Metablock` and new dsse `Envelope` objects? The following routine is typical for `verifylib` regardless of the metadata wrapper.
```
1. Load metadata from disk
2. Verify signatures
3. Extract payload for further processing
```

in_toto/runlib.py
```diff
@@ -903,37 +930,48 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
...
+    verification_key = GPGKey.from_dict(gpg_pubkey)
 
-  link_metadata.verify_signature(verification_key)
+  if isinstance(link_metadata, Envelope):
+    link_metadata.verify([verification_key], 1)
+    link = link_metadata.deserialize_payload(in_toto.models.link.Link)
+    use_dsse = True
+  else:
+    link_metadata.verify(verification_key)
+    link = link_metadata.signed
```
lukpueh: See comment above above more general metadata abstraction that provides a common `verify` method.

tests/test_runlib.py
```diff
@@ -926,12 +944,37 @@ def test_read_only_metadata_directory(self):
...
+    in_toto_record_stop(
+      self.step_name, [self.test_product], self.key, use_dsse=True)
+
+    link_metadata = Envelope.from_file(self.link_name)
+    link_metadata.verify([SSlibKey.from_securesystemslib_key(self.key)], 1)
+
+    link = link_metadata.deserialize_payload(in_toto.models.link.Link)
+    self.assertEqual(list(link.products.keys()), [self.test_product])
+    os.remove(self.link_name)
+
```
lukpueh, line 1025: I wonder if `record stop` shouldn't just use the same metadata wrapper type as `record start`. I don't think there is a use case to switch.

in_toto/runlib.py
```diff
@@ -571,39 +581,47 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
...
 
   elif gpg_keyid:
     LOG.info("Signing link metadata using passed GPG keyid...")
-    signature = link_metadata.sign_gpg(gpg_keyid, gpg_home=gpg_home)
+    signer = GPGSigner(keyid=gpg_keyid, homedir=gpg_home)
 
   elif gpg_use_default:
     LOG.info("Signing link metadata using default GPG key ...")
-    signature = link_metadata.sign_gpg(gpg_keyid=None, gpg_home=gpg_home)
+    signer = GPGSigner(keyid=None, homedir=gpg_home)
```
lukpueh, line 686: I still think that we should, at some point, remove all the key args from`in_toto_run`, `in_toto_record_start` and `in_toto_record_stop`, and instead have them accept a `Signer`, plus, in the case of `in_toto_record_stop` a `Key` for verification.  The code to create the objects based on the key args could live in `common_args.py`.

Would you mind creating a ticket?

in_toto/models/metadata.py
```diff
@@ -198,6 +256,48 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
...
+      raise SignatureVerificationError("No signature found for key '{}'"
+          .format(key.keyid))
+
+    if securesystemslib.formats.GPG_SIGNATURE_SCHEMA.matches(signature):
+      valid = key.verify(
+        GPGSignature.from_dict(signature), self.signed.signable_bytes)
+
+    elif securesystemslib.formats.SIGNATURE_SCHEMA.matches(signature):
+      valid = key.verify(
+        Signature.from_dict(signature), self.signed.signable_bytes)
```
lukpueh: Ideally, `Metablock.signatures` is already a list of `Signature` objects, so you don't have to case handle here. Can you add it to the issue with `Metablock` enhancements that may or may not be worth providing before we deprecate it?

lukpueh (changes requested):
Stellar work, @PradyumnaKrishna. I'm very impressed how well you navigate through the code base. Please address comments inline and we can discuss how to roll out this feature, and also move on to adding dsse support in `verifylib`.

PradyumnaKrishna:
@lukpueh ~I found that https://github.com/in-toto/in-toto/pull/503/commits/7f3a801bf5f091c630dfe3785176c08abdfcd92f was failing for some testcase which was not expected. It raises `FileNotFoundError` and `StorageError`.~

https://github.com/in-toto/in-toto/pull/503/commits/ba415f86a3ee2104ab243c1b61409bd66628d0c2 This does not raises `ThresholdVerificationError` on some testcases.

Edit: `MetadataLoader` was working fine, A wrong testcase was affecting others.

lukpueh:
> [ba415f8](https://github.com/in-toto/in-toto/commit/ba415f86a3ee2104ab243c1b61409bd66628d0c2) This does not raises `ThresholdVerificationError` on some testcases.
> 

Looks like too much indentation for this block:

https://github.com/in-toto/in-toto/blob/ba415f86a3ee2104ab243c1b61409bd66628d0c2/in_toto/verifylib.py#L542-L555

It used to execute after this loop, now it runs in every iteration:
https://github.com/in-toto/in-toto/blob/ba415f86a3ee2104ab243c1b61409bd66628d0c2/in_toto/verifylib.py#L470

I admit that the function has become long enough for mistakes like this to easily happen. Maybe it makes sense to factor out some things to helper functions. I'd at least try to not make it any longer by adding DSSE support. With a common Metablock/Envelope abstraction that has a verify method, this should be possible, right?

in_toto/models/metadata.py
```diff
@@ -282,3 +340,120 @@ def _validate_signatures(self):
...
     for signature in self.signatures:
       securesystemslib.formats.ANY_SIGNATURE_SCHEMA.check_match(signature)
+
+
+class AnyMetadata(SerializationMixin, JSONSerializable):
+  """A Metadata Wrapper to provide some common methods."""
+
+  def __init__(self, metadata):
+    self.metadata: Union[Envelope, Metablock] = metadata
+    self.dsse = isinstance(metadata, Envelope)
```
lukpueh: I don't think composition is the best suited pattern here, it requires case handling in every function for objects of different classes that could have the same interfaces for serialization, signature creation/verification and payload extraction. 

What if AnyMetadata was a base class for Metablock and Envelope that only implements the `from_dict` factory as you have it now, and defines common interfaces for above tasks? I can see two challenges there. 
1. current signature verification methods expect different arguments for Metablock and Envelope
--> A good solution would be to implement a new additional verify method on Metablock that accepts the same arguments as Envelope's verify method.

2. Envelope is implemented in securesystemslib, so it can't inherit from AnyMetadata in in-toto
--> I can see multiple solutions:
   1. Use protocols to not require subclassing. This would require either dropping 3.7 or adding an additional dependency to pull in the backport
   2. Create a glue Envelope class in in-toto, which could inherit both from AnyMetadata and securesystemslib's Envelope. This solutions seems a bit messy.
   3. Implement AnyMetadata and a Metablock-like class in securesystemslib.  The latter is also requested in [#272](https://github.com/secure-systems-lab/securesystemslib/issues/272). However, this would mean  a breaking change.

What do you think?

in_toto/models/metadata.py
PradyumnaKrishna: `Metablock.signatures` is not a list of `Signature` objects right now. Also, I found a flaw in the verify method added previously, `Signature.from_dict` destroys the dict of signature provided.

The solution to add a method similar to the DSSE having same arguments would be tricky. The one way I can think of is to convert the whole Metablock wrapper to use `Signer`, `Key` and `Signature` as objects.

Right now I removed the method and using the `Metablock` verify signature method for verification.

in_toto/models/metadata.py
PradyumnaKrishna: I found that that it is possible to define a base class `AnyMetadata` for `Envelope` and `Metablock` and use them in it's `from_dict` method. I am surprised to see this behaviour. But if we are going to have a base class `AnyMetadata` then these all metadata need to be present in a single file otherwise they will cause circular import.

It is not possible to do first task without refactoring `Metablock` (see https://github.com/in-toto/in-toto/pull/503#discussion_r954870651) and I am not sure about second task either.

in_toto/models/metadata.py
```diff
@@ -23,26 +23,86 @@
...
+    data = super().deserialize(raw_data)
+    _type = data.get("_type")
+    if _type == "link":
+      return Link.read(data)
+    if _type == "layout":
+      return Layout.read(data)
+
+    raise securesystemslib.exceptions.DeserializationError(
+      f"Invalid payload type {_type}, must be `link` or `layout`"
+    )
```
lukpueh, line 159: Do you think it's worth moving `AnyMetadataDeserializer` and `PayloadDeserializer` to an `in_toto.models.serialization` module? 

You'll have a circular import problem with `default_deserializer` then, but you could import `AnyMetadataDeserializer` inside the method.

in_toto/models/metadata.py
```diff
@@ -23,20 +23,86 @@
...
+class AnyMetadataDeserializer(JSONDeserializer):
+  """Deserialize bytes into Metablock / DSSE Envelope."""
+
+  def deserialize(self, raw_data: bytes) -> Union["Envelope", "Metablock"]:
+    """Deserialize JSON bytes into Metablock or Envelope as per ITE-5. """
+
+    data = super().deserialize(raw_data)
+
+    if "payload" in data:
+      if data.get("payloadType") == "application/vnd.in-toto+json":
```
lukpueh: Can you make `"application/vnd.in-toto+json"` a constant either in the metadata module or on the `Envelope` class and also use that constant in `create_envelope`?

in_toto/models/metadata.py
```diff
@@ -23,20 +23,86 @@
...
+
+    raise securesystemslib.exceptions.DeserializationError(
+      f"Invalid payload type {_type}, must be `link` or `layout`"
+    )
+
+
+class Envelope(AnyMetadata, SSlibEnvelope):
+  """in-toto copy of DSSE Envelope."""
+
+  def get_payload(self, deserializer: Optional[BaseDeserializer] = None
```
lukpueh: Please add docstring.

in_toto/models/metadata.py
```diff
@@ -23,20 +23,86 @@
...
+    if _type == "layout":
+      return Layout.read(data)
+
+    raise securesystemslib.exceptions.DeserializationError(
+      f"Invalid payload type {_type}, must be `link` or `layout`"
+    )
+
+
+class Envelope(AnyMetadata, SSlibEnvelope):
+  """in-toto copy of DSSE Envelope."""
```
lukpueh: Not really a copy. What about
```suggestion
  """DSSE Envelope for in-toto payloads."""
```

in_toto/models/metadata.py
```diff
@@ -23,20 +23,86 @@
...
+
+  def deserialize(self, raw_data: bytes) -> Union[Link, Layout]:
+    data = super().deserialize(raw_data)
+    _type = data.get("_type")
+    if _type == "link":
+      return Link.read(data)
+    if _type == "layout":
+      return Layout.read(data)
+
+    raise securesystemslib.exceptions.DeserializationError(
```
lukpueh: Is there a reason to not do an `if "link", elif "layout", else raise` switch?

in_toto/models/metadata.py
lukpueh: Can you create tickets for these?

lukpueh (changes requested):
Here's a first batch of review comments. All minor stuff really. The PR is in great shape! 💯 

Could you please also look at the [sphinx docs](https://github.com/in-toto/in-toto/tree/develop/doc) and see if we need to explicitly enable docs for new classes (AnyMetadata, Envelope, serializers?).

Please also make sure that important methods inherited from securesystemslib (e.g. from/to_bytes/file are shown in the docstring of the inheriting class).

lukpueh:
> ... there is a bit of redundant functionality in `Metablock` now. I'm fine leaving as is and decide later if we want to just sit out the DSSE transition phase and remove the whole Metablock class at once, or clean it up in a breaking change before.
> 
> Still, would you mind creating an issue for the following items?
> 
> * `JSONSerializer` can be used instead of `Metablock.__repr__` and `Metablock.compact_json`
> * `Metablock.from_file` and `Metadata.to_file` (both via `SerializationMixin`) can be used instead of `Metadata.load` and `Metablock.dump`
> * I don't think we need `attr` anymore on `Metablock`

More things to ticketize:

* `create_sig` and `verify_sigs` can be used instead of `sign`, `sign_gpg`, `verify_signature`.
* Maybe we can find better names for `create_sig` and `verify_sigs` (`sign` and `verify_signature` are already taken)

in_toto/models/metadata.py
PradyumnaKrishna: pylint tells me to do it that way.

doc/source/model.rst
```diff
@@ -10,13 +10,36 @@ and read or write in-toto conformant metadata (see
...
+.. autoclass:: in_toto.models.serialization.AnyMetadataDeserializer
+  :inherited-members:
+  :members:
+
+PayloadDeserializer
+-------------------
+.. autoclass:: in_toto.models.serialization.PayloadDeserializer
+  :inherited-members:
+  :members:
+
```
lukpueh: Minor nit: Can we tidy up a bit and stick the deserializer classes [in a subdocument, as we do in python-tuf](https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.serialization.html)

in_toto/models/metadata.py
```diff
@@ -23,20 +23,57 @@
...
 from in_toto.models.layout import Layout
-from in_toto.exceptions import SignatureVerificationError
+
+
+ENVELOPE_PAYLOAD_TYPE = "application/vnd.in-toto+json"
+
+
+# pylint: disable=import-outside-toplevel
+class AnyMetadata(SerializationMixin):
+  """A Metadata abstraction between DSSE Envelope and Metablock."""
```
lukpueh, line 48: Should we mention how `AnyMetadata` can be used, either in words, or even with a little code snippet, e.g. as outlined here: https://github.com/theupdateframework/python-tuf/issues/2093#issuecomment-1234034698 ?

I think it would also be useful if we defined the methods that Envelope and Metablock have in common, both for future mypy runs and readthedocs.

in_toto/models/serialization.py
```diff
@@ -0,0 +1,64 @@
...
+from securesystemslib.serialization import JSONDeserializer
+
+from in_toto.exceptions import InvalidMetadata
+from in_toto.models.layout import Layout
+from in_toto.models.link import Link
+from in_toto.models.metadata import Envelope, Metablock, ENVELOPE_PAYLOAD_TYPE
+
+
+class PayloadDeserializer(JSONDeserializer):
+  """Deserialize JSON bytes into Link or Layout."""
```
lukpueh: I think you can remove this docstring. The one in deserialize should be enough.

in_toto/models/serialization.py
```diff
@@ -0,0 +1,64 @@
...
+    if _type == "layout":
+      return Layout.read(data)
+
+    raise DeserializationError(
+      f"Invalid payload type {_type}, must be `link` or `layout`"
+    )
+
+
+class AnyMetadataDeserializer(JSONDeserializer):
+  """Deserialize bytes into Metablock / DSSE Envelope."""
```
lukpueh: I think you can remove this docstring. The one in deserialize should be enough.

lukpueh (changes requested):
Thanks for adding docs! Here's a review of just those:

Things that need to be fixed in securesystemslib for docstrings that get pulled in via inheritence:
- `default_serializer` and `default_deserializer` don't need to show up in the docstrings, maybe we can just make them internal, i.e.  `_default_serializer` and `_default_deserializer`, which should automatically exclude them.
- please remove mention of `Serializable`, which no longer exists
- please use double-backticks consistently and full import paths for class names etc., eg. ` ``securesystemslib.serialization.JSONDeserializer`` `

in_toto/models/common.py
```diff
@@ -61,3 +61,15 @@ def signable_bytes(self):
...
+    """Create a DSSE envelope with signable bytes as payload."""
+
+    # pylint: disable-next=import-outside-toplevel
+    from in_toto.models.metadata import Envelope, ENVELOPE_PAYLOAD_TYPE
+
+    return Envelope(
+      payload=self.signable_bytes,
+      payload_type=ENVELOPE_PAYLOAD_TYPE,
+      signatures=[]
+    )
```
lukpueh: Minor nit: IMO, the contained item does not necessarily need to know about the container. Does the method make more sense as static method on AnyMetadata, or classmethod on Envelope?

in_toto/models/metadata.py
```diff
@@ -137,13 +191,38 @@ def type_(self):
...
+    return signature
+
   def sign(self, key):
     """Creates signature over signable with key and adds it to signatures.
 
     Uses the UTF-8 encoded canonical JSON byte representation of the signable
     attribute to create signatures deterministically.
 
-    Attributes:
+    Arguments:
```
lukpueh: Please remove unrelated changes. The PR is big enough. :)

in_toto/models/metadata.py
```diff
@@ -156,6 +235,7 @@ def sign(self, key):
       The signature. Format is securesystemslib.formats.SIGNATURE_SCHEMA.
 
     """
+
```
lukpueh: Please remove unrelated changes.

in_toto/models/metadata.py
```diff
@@ -216,9 +361,9 @@ def verify_signature(self, verification_key):
     Raises:
       securesystemslib.exceptions.FormatError: The passed key is malformed.
 
-      SignatureVerificationError: No signature keyid matches the verification
-          key keyid, or the matching signature is malformed, or the matching
-          signature is invalid.
+      securesystemslib.exceptions.SignatureVerificationError: No signature
+          keyid matches the verification key keyid, or the matching signature
+          is malformed, or the matching signature is invalid.
```
lukpueh: Please remove unrelated changes.

in_toto/models/metadata.py
```diff
@@ -240,8 +385,8 @@ def verify_signature(self, verification_key):
         break
 
     else:
-      raise SignatureVerificationError("No signature found for key '{}'"
-          .format(verification_keyid))
+      raise securesystemslib.exceptions.SignatureVerificationError(
+        "No signature found for key '{}'".format(verification_keyid))
```
lukpueh: Please remove unrelated changes.

in_toto/models/metadata.py
```diff
@@ -255,8 +400,8 @@ def verify_signature(self, verification_key):
       valid = False
 
     if not valid:
-      raise SignatureVerificationError("Invalid signature for keyid '{}'"
-          .format(verification_keyid))
+      raise securesystemslib.exceptions.SignatureVerificationError(
+        "Invalid signature for keyid '{}'".format(verification_keyid))
```
lukpueh: Please remove unrelated changes.

in_toto/models/metadata.py
```diff
@@ -198,6 +278,71 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
...
+      # GPGSignature/ Signature from_dict method destroys the signature dict,
+      # Hence, deepcopy of signature is created.
+      if securesystemslib.formats.GPG_SIGNATURE_SCHEMA.matches(signature):
+        signature = GPGSignature.from_dict(copy.deepcopy(signature))
+
+      elif securesystemslib.formats.SIGNATURE_SCHEMA.matches(signature):
+        signature = Signature.from_dict(copy.deepcopy(signature))
+
+      else:
+        raise securesystemslib.exceptions.SignatureVerificationError
```
lukpueh: We need to be careful here. A single malformed or unknown signature should not fail the entire verification. We only fail if good signatures don't meet the threshold.

lukpueh (changes requested):
More comments on `models` subpackage...

in_toto/in_toto_sign.py
```diff
@@ -144,25 +148,32 @@ def _verify_metadata(metadata, args):
...
   try:
+    keys = []
+
     # Load pubkeys from disk ....
     if args.key is not None:
       pub_key_dict = interface.import_publickeys_from_file(args.key,
           args.key_type)
+      keys = [
+        SSlibKey.from_securesystemslib_key(key)
+        for key in list(pub_key_dict.values())
```
lukpueh: Why cast `dict_values` to `list`? Same comment applies below.

in_toto/in_toto_sign.py
```diff
@@ -144,25 +148,32 @@ def _verify_metadata(metadata, args):
...
+        GPGKey.from_dict(key)
+        for key in list(pub_key_dict.values())
+      ]
 
-
-    for keyid, verification_key in pub_key_dict.items():
-      metadata.verify_signature(verification_key)
-      LOG.info("Signature verification passed for keyid '{}'"
-          .format(keyid))
+    LOG.info("Verifying metadata signatures against keys in `pub_key_dict`...")
```
lukpueh: Minor nit: The caller does not need to know about a variable called `pub_key_dict`. I think it's more useful to print the keyids as we did before. i.e. `pub_key_dict.keys()`

in_toto/verifylib.py
```diff
@@ -222,7 +225,7 @@ def run_all_inspections(layout, persist_inspection_links):
         link.signed.byproducts.get("return-value"),
         inspection.run)
 
-    inspection_links_dict[inspection.name] = link
+    inspection_links_dict[inspection.name] = link.signed
```
lukpueh, line 226: Please adopt change in docstring.

in_toto/verifylib.py
```diff
@@ -1470,13 +1465,22 @@ def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
     useful during recursive sublayout verification.
 
   """
+  keys = []
+  for key in list(layout_key_dict.values()):
+    if securesystemslib.formats.GPG_PUBKEY_SCHEMA.matches(key):
+      keys.append(GPGKey.from_dict(key))
+    elif securesystemslib.formats.KEY_SCHEMA.matches(key):
+      keys.append(SSlibKey.from_securesystemslib_key(key))
```
lukpueh: Minor nit: maybe move to helper function? Same goes for similar lines of code in other places, where traditional public keys/private keys/signatures are translated to Key, Signer, Signature objects.

in_toto/verifylib.py
```diff
@@ -1524,4 +1531,5 @@ def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
   # Return a link file which summarizes the entire software supply chain
   # This is mostly relevant if the currently verified supply chain is embedded
   # in another supply chain
-  return get_summary_link(layout, reduced_chain_link_dict, step_name)
+  return get_summary_link(layout, reduced_chain_link_dict, step_name,
+      use_metablock)
```
lukpueh: I think you can leave `get_summary_link` is it was. Summary links don't even have signatures, so there is no reason to support two different signature wrappers.

lukpueh: We could even only return the payload here. But then we should change `verify_sublayouts` to extract payloads for all items in `chain_link_dict`, not only the ones it recurses into.

in_toto/verifylib.py
lukpueh: Actually, there are a bunch of occurrences of `Metablock` in docstrings of this document, that are no longer accurate.

lukpueh (changes requested):
Here's a review of `verifylib.py` and other module, where I didn't find anything comment-worthy.

Files still to be reviewed are `runlib.py` and tests.

in_toto/models/metadata.py
PradyumnaKrishna: This test case was failing that's why I raised this exception.

https://github.com/in-toto/in-toto/blob/8fac55f5bbcfcf36d9c7ff53dddc6c6b12e483ec/tests/test_verifylib.py#L948-L949

in_toto/models/serialization.py
PradyumnaKrishna: Pylint shows warning.

lukpueh:
> Things that need to be fixed in securesystemslib for docstrings that get pulled in via inheritence:
> 
> * `default_serializer` and `default_deserializer` don't need to show up in the docstrings, maybe we can just make them internal, i.e.  `_default_serializer` and `_default_deserializer`, which should automatically exclude them.
> * please remove mention of `Serializable`, which no longer exists
> * please use double-backticks consistently and full import paths for class names etc., eg. `` ``securesystemslib.serialization.JSONDeserializer`` ``

Thanks for addressing these in https://github.com/in-toto/securesystemslib/pull/13! When re-building the docs, I noticed that a reader of `SerializationMixin.{from, to}_{bytes, file}` won't know about the default de/serializers, which might be useful usage information. 

Should we mention  `cls._default_(de)serializer` in `SerializationMixin.{from, to}_{bytes, file}` and show docstrings of the overridden methods (e.g. `AnyMetadata.cls._default_(de)serializer`) after all? 

Alternatively, we could just mention the default de/serializers (i.e. JSON) to the class docstrings of the classes that inherit `{from, to}_{bytes, file}` (i.e. AnyMetadata, Metablock, Envelope).

Any thoughts?

in_toto/runlib.py
```diff
@@ -903,37 +930,48 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
...
+  link.products = record_artifacts_as_dict(
       product_list, exclude_patterns=exclude_patterns, base_path=base_path,
       follow_symlink_dirs=True, normalize_line_endings=normalize_line_endings,
       lstrip_paths=lstrip_paths)
 
-  link_metadata.signatures = []
+  if use_metablock:
+    LOG.info("Generating link metadata using Metablock...")
+    link_metadata = Metablock(signed=link)
+  else:
```
lukpueh: Minor nit: Assuming that AnyMetadata can only be Metablock or Envelope adds unnecessary coupling. I suggest to check for both and raise an error if it's neither Metablock nor Envelope.  

```suggestion
  if isinstance(link_metadata, Metablock):
    LOG.info("Generating link metadata using Metablock...")
    link_metadata = Metablock(signed=link)
  elif isinstance(link_metadata, Envelope):
```

in_toto/runlib.py
```diff
@@ -571,39 +580,49 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
       materials=materials_dict, products=products_dict, command=link_cmd_args,
       byproducts=byproducts, environment=environment)
 
-  link_metadata = Metablock(signed=link, compact_json=compact_json)
+  if use_dsse:
+    LOG.info("Generating link metadata using DSSE...")
+    link_metadata = link.create_envelope()
+  else:
+    LOG.info("Generating link metadata using Metablock...")
+    link_metadata = Metablock(signed=link, compact_json=compact_json)
```
lukpueh, line 673: Please update "Returns" section in docstring.

in_toto/runlib.py
```diff
@@ -406,11 +411,11 @@ def in_toto_mock(name, link_cmd_args):
 
   """
```
lukpueh, line 495: Please update "Returns" section in docstring.

tests/test_runlib.py
```diff
@@ -704,11 +704,18 @@ def test_in_toto_read_only_metadata_directory(self):
...
+    with self.assertRaises(securesystemslib.exceptions.StorageError):
       in_toto_run(self.step_name, None, None, ["python", "--version"],
           True, self.key, metadata_directory=tmp_dir)
     os.rmdir(tmp_dir)
 
+  def test_in_toto_for_dsse(self):
+    """Test metadata generation using dsse."""
+
+    link_metadata = in_toto_run(self.step_name, None, None,
+        ["python", "--version"], True, self.key, use_dsse=True)
```
lukpueh, line 779: Nit: The test would be stronger if we assert type of returned link_metadata here.

lukpueh (changes requested):
Here's a review of the modules I haven't reviewed so far, that is runlib, test_runlib and test_verifylib.

lukpueh:
Just finished reviewing the full PR. This is amazing work, @PradyumnaKrishna! Let's try to get it merged soon. Here's what's missing:
- Address comments from reviews
- Add more tests
- Ticketize remaining issues (see e.g. [#503 (comment)](https://github.com/in-toto/in-toto/pull/503#issuecomment-1238057980), [#503 (comment)](https://github.com/in-toto/in-toto/pull/503#discussion_r950104376), [#503 (comment)](https://github.com/in-toto/in-toto/pull/503#discussion_r993422133))
- Figure out to what extent this PR breaks the API 
  - Is it only a few exceptions and a few semi-internal functions in verifylib/runlib? 
  - Can we live with those, or should we aim for full backwards compatibility?
  - Will this require a major release?
- Upstream and release in-toto/securesystemslib
  - Before we do that, I'd like to at least address [in-toto/securesystemslib#8](https://github.com/in-toto/securesystemslib/issues/8)
- Update securesystemslib dependency

Are you available to work on any of these things, @PradyumnaKrishna?

PradyumnaKrishna:
> Thanks for addressing these in [in-toto/securesystemslib#13](https://github.com/in-toto/securesystemslib/pull/13)! When re-building the docs, I noticed that a reader of `SerializationMixin.{from, to}_{bytes, file}` won't know about the default de/serializers, which might be useful usage information.
> 
> Should we mention `cls._default_(de)serializer` in `SerializationMixin.{from, to}_{bytes, file}` and show docstrings of the overridden methods (e.g. `AnyMetadata.cls._default_(de)serializer`) after all?
> 
> Alternatively, we could just mention the default de/serializers (i.e. JSON) to the class docstrings of the classes that inherit `{from, to}_{bytes, file}` (i.e. AnyMetadata, Metablock, Envelope).
> 
> Any thoughts?

I guess we can mention those default deserializer and serializer in `AnyMetadata` and also mention those two special classes in the class doc-string of `SerializationMixin`.

Also there is one small issue that I found during testing. [This one line of code](https://github.com/in-toto/in-toto/pull/503/files#diff-40f0ea7898882c994402200f6f9c27e7b8e1934971eda4ba81b853b3d9673466R930) breaks for DSSE because `Envelope` signatures are object and `Metablock` signatures are dict.
The easiest fix would be if-else statement but I would like your opinion.

> Are you available to work on any of these things, @PradyumnaKrishna?

Yes, I am available, I have couple uncommitted changes that I will push this week and address other issues. I will not be available next week but continue my work after that.

lukpueh:
> we can mention those default deserializer and serializer in AnyMetadata

👍 

> ... also mention those two special classes in the class doc-string of SerializationMixin.

Which special classes? 

> The easiest fix would be if-else statement but I would like your opinion.

Good find! I guess if/else is okay.

> Yes, I am available, I have couple uncommitted changes that I will push this week and address other issues. I will not be available next week but continue my work after that.

🚀

PradyumnaKrishna:
> Which special classes?
I mean we can mention `_default_serializer` and `_default_deserializer` in the class doc-string of `SerializationMixin`.

lukpueh:
> I mean we can mention `_default_serializer` and `_default_deserializer` in the class doc-string of `SerializationMixin`.

I think that's not necessary. The `SerializationMixin` class docstring won't show in the in-toto docs. And I was more worried about users of `{AnyMetadata, Envelope, Metablock}.{from, to}_{bytes, file}`, than about users of `SerializationMixin`.

in_toto/runlib.py
PradyumnaKrishna: `AnyMetadataDeserializer` already raises a [InvalidMetadata exception](https://github.com/in-toto/in-toto/pull/503/files#diff-d036bafdd463a770d30e4159e24f9f2c5dce02546cc59cc4f85dfeba1bb12aa8R64) while loading. 

There is signature verification and payload extraction step in between loading of link metadata and generation of new metadata.
So it would be better if I add a check for both metadata after the [`AnyMetadata.from_file()`](https://github.com/in-toto/in-toto/pull/503/files#diff-40f0ea7898882c994402200f6f9c27e7b8e1934971eda4ba81b853b3d9673466R905) command.

in_toto/models/common.py
```diff
@@ -45,7 +46,7 @@ def validate(self):
 
 
 @attr.s(repr=False, init=False)
-class Signable(ValidationMixin):
+class Signable(ValidationMixin, SerializationMixin):
```
lukpueh: `SerializationMixin` also adds `from_bytes`, `from_file` and `to_file` method, which we don't need, and, at least the former, won't work. Right? Maybe it's better to just reimplement `to_bytes` instead of using the mixin.

in_toto/models/metadata.py
lukpueh: ping

in_toto/models/metadata.py
lukpueh: I still think this would be useful, but we can ticketize it.

in_toto/models/common.py
lukpueh: What do you think?

PradyumnaKrishna: classmethod on envelope is a better approach I guess.

in_toto/models/common.py
PradyumnaKrishna: Yeah, I will remove `SerializationMixin`.

in_toto/models/metadata.py
lukpueh: What did you have there before? IMO that particular test should also fail if you `continue` here instead of raising an error. Then the error is raised further below, because the threshold isn't met. Right?

But it looks like the old `verifylib.verify_layout_signatures` would have also failed for a malformed signature. Interestingly, in both cases it depends on where in the signature list the malformed signature is located, and whether enough valid signatures can be matched by keyid and successfully verified before. It also looks like old and new functions would fail for slightly different sequences of keyid-matching, valid, and malformed signatures .

Regardless, it shouldn't matter for `in_toto_verify`, because `Metablock.validate()` is called on load and will error on malformed signatures before we get to verify them.

in_toto/runlib.py
lukpueh: > So it would be better if I add a check for both metadata after the [AnyMetadata.from_file()](https://github.com/in-toto/in-toto/pull/503/files#diff-40f0ea7898882c994402200f6f9c27e7b8e1934971eda4ba81b853b3d9673466R905) command.

If we add the check, then yes, it would be better to fail before we do all that work.

But it's probably okay to just rely on `AnyMetadata.from_file()` returning either `Metablock` or `DSSE`. I was just worried that a change in `AnyMetadataDeserializer` not being adopted here and leading to weird behavior. But you can leave it as is for now.

tests/test_param_substitution.py
```diff
@@ -191,6 +192,20 @@ def test_substitute(self):
...
+    signed_layout = self.layout.create_envelope()
+    signed_layout.create_sig(SSlibSigner(self.alice))
+
+    # we will catch a LinkNotFound error because we don't have (and don't need)
+    # the metadata.
+    with self.assertRaises(in_toto.exceptions.LinkNotFoundError):
+      in_toto_verify(signed_layout, self.alice_pub_dict,
+          substitution_parameters={"EDITOR":"vim"})
+
+    self.assertEqual(self.layout.steps[0].expected_command[0], "vim")
```
lukpueh, line 207: Wow, that's thorough testing. I wouldn't expect this to be affected by the wrapper. But you never know. :)

tests/test_in_toto_record.py
```diff
@@ -227,5 +227,168 @@ def test_missing_unfinished_link(self):
...
+  @classmethod
+  def tearDownClass(self):
+    self.tear_down_test_dir()
+
+
+  def test_start_stop(self):
+    """Test CLI command record start/stop with various arguments. """
+
+    # Start/stop recording using rsa key
+    args = ["--step-name", "test1", "--key", self.rsa_key_path, "--use-dsse"]
```
lukpueh, line 258: You test with `"--use-dsse"` here but not below. Doesn't that make below tests duplicates of `TestInTotoRecordTool`?

lukpueh (commented):
Just took a look at the `test_in_toto_{sign, record, run}.py` additions. I don't think we need to duplicate all these cli tests for dsse, especially since the tests don't assert whether dsse was actually used.

In my mind the goal of the cli tests is to assert that certain combinations of cli options result in a specific exit value. And I think we get good coverage for that without testing all combinations of all options. So for the sake of dsse, it might be enough to test only one or two or a couple of exemplary invocations with `--use-dsse`, or with dsse metadata as inputs.

More thorough tests of combinations of parameters should be performed for the `in_toto_{run, record, verify}` library functions. There we should also assert that the input or output metadata files are actually dsse.

Does this make sense?

in_toto/verifylib.py
```diff
@@ -438,12 +451,8 @@ def verify_link_signature_thresholds(layout, chain_link_dict):
...
       # Verify signature and skip invalidly signed links
       try:
-        if securesystemslib.formats.GPG_PUBKEY_SCHEMA.matches(
-            verification_key):
-          keys = [GPGKey.from_dict(verification_key)]
-        elif securesystemslib.formats.KEY_SCHEMA.matches(verification_key):
-          keys = [SSlibKey.from_securesystemslib_key(verification_key)]
-
+        keys = create_key_list_from_key_dict(
+            {authorized_keyid: verification_key})
```
lukpueh: Just remembered that we have to be careful when using `from_dict` methods (which you call in `create_key_list_from_key_dict`), as they usually have the side-effect of popping elements from the passed dict.

The reason why we can still use `verification_key["keyid"]` below is that `SslibKey.from_dict` only pops some items put not the keyid and `GPGKey.from_dict` does not pop at all yet. The latter may change with https://github.com/in-toto/securesystemslib/issues/7.

We should  hopefully catch the change with our tests, but it might be worth already updating the code to not use key dicts after they have been passed to `create_key_list_from_key_dict` (and add a note about the side-effect to the function docstring).

in_toto/verifylib.py
```diff
@@ -1356,21 +1340,24 @@ def get_summary_link(layout, reduced_chain_link_dict, name):
...
-              <link name> : <Metablock containing a Link object>,
+              <link name> : <Link object>,
               ...
             }
     name:
             The name that the summary link will be associated with.
 
+    use_metablock:
+            Use metablock to generate metadata.
+
```
lukpueh:
```suggestion
```

lukpueh:
Thanks for you optimization of `chain_link_dict` in cbbd01d2c132eb46d2dfe2a836b1d7c1dfcb8cdc, it makes a lot of sense!  

For the sake of not going crazy, can we maybe find different names for the data structure that contains the full metadata objects with Metablock/Envelope wrapper and the one that only include the Link objects? `chain_link_dict` by itself is unintuitive enough (I take the blame). 

Maybe keep using `chain_link_dict` for the new data structure without wrapper, and call the other one something like `chain_metadata`? ... naming is hard :( 
 
If you think a rename is not necessary, I'm also fine with adding a note inside the `in_toto_verify` function, where you call `verify_sublayouts`, that explains how the passed `chain_link_dict` differs from the returned one, with regard to the removed metadata wrapper.

tests/test_in_toto_record.py
PradyumnaKrishna: Oh, I forgot to add it, will fix them.

lukpueh: Before you fix them, please see my comment above: https://github.com/in-toto/in-toto/pull/503#pullrequestreview-1140292691

I'm not sure if we even need that many dsse-flavored cli tests.

PradyumnaKrishna:
> More thorough tests of combinations of parameters should be performed for the `in_toto_{run, record, verify}` library functions. There we should also assert that the input or output metadata files are actually dsse.
> 
> Does this make sense?

I will have to review all the test case one by one, to figure out the coverage for DSSE. I will try to make these tests more optimal for DSSE.

> Maybe keep using chain_link_dict for the new data structure without wrapper, and call the other one something like chain_metadata? ... naming is hard :(

It will reduce confusion between metadata and link/layout objects. I will rename the data structures and docs to reduce this clutter.

I will not be available next week, So I will update this Pull Request after next Monday (24 Oct).

PradyumnaKrishna:
Sorry for the force push,

After carefully reviewing all tests, I removed tests related to
* argument parsing fails
* invalid arguments
* same tests with different key types
* metadata directory

and remains tests related to
* main test
* with optional args (record streams was broken previously)
* one test with gpg

I would like to get your opinion on the optional args test case.
Also, I don't know how to generate metadata files with gpg signatures for in-toto-verify testing.

in_toto/verifylib.py
PradyumnaKrishna: I found `GPGKey.from_dict` and `SSlibKey.from_securesystemslib_key` both doesn't have any side effects currently.

tests/models/test_envelope.py
```diff
@@ -0,0 +1,84 @@
...
+    self.assertEqual(env.get_payload(), layout)
+
+  def test_load_envelope(self):
+    """Test loading and parsing of in-toto envelope."""
+
+    # Demo DSSE Metadata Files
+    demo_dsse_files = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "../demo_dsse_files")
+
+    self.set_up_test_dir()
```
lukpueh: FYI: `set_up_test_dir` needs `tear_down_test_dir` to clean up.

lukpueh: Also, you're only reading `demo_dsse_files`, so you actually don't need to copy them to a test dir, do you?

lukpueh:
Thanks for addressing all my comments and adding tests, @PradyumnaKrishna !  I suggest we leave this PR as is for the time being and focus on upstreaming `in-toto/securesystemslib` next. 

There is some realted work going on upstream (see https://github.com/secure-systems-lab/securesystemslib/pull/456). This also means some conflicts and minor re-designs that will be necessary here. But it looks like things will play together nicely eventually. And I am happy to help coordinate upstreaming.

Regarding your comments from above:

> with optional args (record streams was broken previously)

What exactly was broken?

> I would like to get your opinion on the optional args test case.

Can you elaborate?

> Also, I don't know how to generate metadata files with gpg signatures for in-toto-verify testing.

I looks like you have figured it out in the meanwhile?

PradyumnaKrishna:
> Can you elaborate?

Just want your opinion on these test cases, are they enough, or more than enough.

> What exactly was broken?

I am just mentioning that record streams argument was broken in in-toto-run previously becuase the CanonicalJSON payload deserialization was failing. So, these test cases might be important.

> I looks like you have figured it out in the meanwhile?

Yes, I figured out to the way to load gpg keys from gpg keyring.


> Thanks for addressing all my comments and adding tests, @PradyumnaKrishna ! I suggest we leave this PR as is for the time being and focus on upstreaming `in-toto/securesystemslib` next.

I am ready to help in upstreaming `in-toto/securesystemslib`, but not sure where to begin with.

lukpueh:
> Just want your opinion on these test cases, are they enough, or more than enough.

I think they are fine for now. I'll take another look, once we've upstreamed `in-toto/securesystemslib`.

> I am ready to help in upstreaming `in-toto/securesystemslib`, but not sure where to begin with.

Great! I'll post some thoughts about how to do this in https://github.com/secure-systems-lab/securesystemslib/issues/370

in_toto/verifylib.py
```diff
@@ -430,7 +403,7 @@ def verify_link_signature_thresholds(layout, chain_link_dict):
       main_keys_for_subkeys[sub_keyid] = main_key
 
   # Dict for valid and authorized links of all steps of the layout
-  verfied_chain_link_dict = {}
+  verfied_steps_metadata = {}
```
adityasaky:
```suggestion
  verified_steps_metadata = {}
```

lukpueh:
https://github.com/secure-systems-lab/securesystemslib/pull/487#issuecomment-1432924302 suggests to drop the `securesystemslib.serialization` used here. To move forward with this PR, I see two options;
a. replicate `securesystemslib.serialization` in in-toto,
b. do something simpler.

Summarizing a conversation with @PradyumnaKrishna plus some additional thoughts, I think b. might be good enough:

- Envelope implements load/dump methods which read/write from/to standard DSSE JSON envelope file 
  -> more flexibility about the envelope format is optional as per the DSSE spec and currently not needed
- AnyMetadata implements its own load method, which deserializes JSON and dispatches to Metablock or Envelope as per ITE-5
- Envelope.get_payload implements json deserialization + dispatching to Link or Layout creation 
  -> more flexibility about the payload format is not mandated by ITE-5 and currently not needed

While this solution has a few repeated lines of code, it should overall be less code, and definitely reduces complexity. Optimization towards arbitrary envelope and payload formats is not a priority at the moment.

doc/source/model.rst
```diff
@@ -10,13 +10,42 @@ and read or write in-toto conformant metadata (see
 :doc:`layout-creation-example`).
 
 
+AnyMetadata
+-----------
+.. autoclass:: in_toto.models.metadata.AnyMetadata
```
lukpueh:
```suggestion
Metadata
-----------
.. autoclass:: in_toto.models.metadata.Metadata
```

doc/source/model.rst
```diff
@@ -10,13 +10,42 @@ and read or write in-toto conformant metadata (see
...
+.. autoclass:: in_toto.models.serialization.AnyMetadataDeserializer
+  :inherited-members:
+  :members:
+
+PayloadDeserializer
+*******************
+.. autoclass:: in_toto.models.serialization.PayloadDeserializer
+  :inherited-members:
+  :members:
+
```
lukpueh:
```suggestion
```

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+    """Creates signature over signable with Signer and adds it to signatures.
+
+    Uses the UTF-8 encoded canonical JSON byte representation of the signable
+    attribute to create signatures deterministically.
+
+    Arguments:
+      signer: A "Signer" class instance to create signature.
+
+    Raises:
+      securesystemslib.exceptions.FormatError: Key argument is malformed.
```
lukpueh: There no longer is a key.

in_toto/models/metadata.py
```diff
@@ -216,9 +422,9 @@ def verify_signature(self, verification_key):
     Raises:
       securesystemslib.exceptions.FormatError: The passed key is malformed.
 
-      SignatureVerificationError: No signature keyid matches the verification
-          key keyid, or the matching signature is malformed, or the matching
-          signature is invalid.
+      securesystemslib.exceptions.VerificationError: No signature keyid matches
```
lukpueh: I think the old function should keep the old exception until we deprecate it.

lukpueh (changes requested):
Thanks for implementing my recent suggestion! This is ready for final review, right? I just left a couple of minor comments. Will take a closer look soon.

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+        keys: A list of public keys to verify the signatures.
+        threshold: Number of signatures needed to pass the verification.
+
+    Raises:
+        ValueError: If "threshold" is not valid.
+        VerificationError: If the enclosed signatures do not pass the
+          verification.
+
+    Returns:
+        accepted_keys: A dict of unique public keys.
```
lukpueh: See https://github.com/secure-systems-lab/securesystemslib/pull/487/files#r1113994810

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+    ).encode("utf-8")
+
+    with open(path, "wb") as fp:
+      fp.write(json_bytes)
+
+  def create_signature(self, signer: Signer) -> Signature:
+    """Creates signature over signable with Signer and adds it to signatures.
+
+    Uses the UTF-8 encoded canonical JSON byte representation of the signable
+    attribute to create signatures deterministically.
```
lukpueh: This is only true for Metablock. Either move to Metablock.create_signature, or just remove.

```suggestion
```

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+
+  def verify_signatures(
+    self,
+    keys: List[Key],
+    threshold: int
+  ) -> Dict[str, Key]:
+    """Verifies signature over signable in signatures with a list of keys.
+
+    Uses the UTF-8 encoded canonical JSON byte representation of the signable
+    attribute to verify the signature deterministically.
```
lukpueh: This is only true for Metablock, either move to Metablock.create_signature, or just remove.

```suggestion
```

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+      if data.get("payloadType") == ENVELOPE_PAYLOAD_TYPE:
+        return Envelope.from_dict(data)
+
+    elif "signed" in data:
+      return Metablock.from_dict(data)
+
+    raise InvalidMetadata
+
+  def to_dict(self) -> dict:
+    """Abstractmethod to convert metadata into dict"""
```
lukpueh: Minor nit: This is not really an abstract method. I suggest you just move the Metablock.to_dict docstring her. It seems generic enough.

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+    raise NotImplementedError  # pragma: no cover
+
+  @classmethod
+  def load(cls, path) -> Union["Envelope", "Metablock"]:
```
lukpueh: You should override `load`, otherwise the type annotation (and returns section in docstring) is incorrect.
Here the return value is either Envelope or Metablock, in Envelope.load it is only Envelope, and in Metablock.load only Metablock .

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+    """Loads the JSON string representation of metadata from disk.
+
+    Arguments:
+      path: The path to read the file from.
+
+    Raises:
+      IOError: The file cannot be read.
+      securesystemslib.exceptions.FormatError: Metadata format is invalid.
```
lukpueh, line 77: Could also be `InvalidMetadata` now, right?

in_toto/models/metadata.py
```diff
@@ -23,20 +23,195 @@
...
+        VerificationError: If the enclosed signatures do not pass the
+          verification.
+
+    Returns:
+        accepted_keys: A dict of unique public keys.
+    """
+    raise NotImplementedError  # pragma: no cover
+
+  def get_payload(self):
+    """Returns ``Link`` or ``Layout`` from the metadata wrapper."""
```
lukpueh: Minor nit: We don't use the term "metadata wrapper" anywhere else. Might be enough to just say
```suggestion
    """Returns ``Link`` or ``Layout``."""
```

in_toto/models/metadata.py
```diff
@@ -198,6 +358,52 @@ def sign_gpg(self, gpg_keyid=None, gpg_home=None):
...
+    # checks for threshold value.
+    if threshold <= 0:
+      raise ValueError("Threshold must be greater than 0")
+
+    if len(keys) < threshold:
+      raise ValueError("Number of keys can't be less than threshold")
+
+    for signature in self.signatures:
+
+      # GPGSignature/ Signature from_dict method destroys the signature dict,
```
lukpueh: `GPGSignature` does not, only `Signature` does.

in_toto/models/metadata.py
```diff
@@ -79,38 +254,9 @@ def __repr__(self):
       )
 
 
-  def dump(self, path):
```
lukpueh: This used to consult with `Metablock.compact_json` (see __repr__). Now that we use `Metadata.dump` the attribute is ignored. Should we keep this function, for the sake of backwards compatibility?

in_toto/runlib.py
lukpueh: Ticket exists: #532

in_toto/in_toto_sign.py
```diff
@@ -155,15 +162,13 @@ def _verify_metadata(metadata, args):
...
           args.gpg, args.gpg_home)
 
-
-    for keyid, verification_key in pub_key_dict.items():
-      metadata.verify_signature(verification_key)
-      LOG.info("Signature verification passed for keyid '{}'"
-          .format(keyid))
+    keys = create_key_list_from_key_dict(pub_key_dict)
+    LOG.info("Verifying metadata signatures against keys...")
+    metadata.verify_signatures(keys, len(keys))
```
lukpueh: Prior to this PR, we failed if any of the public keys couldn't verify the signature. 

Now `create_key_list_from_key_dict` might silently drop keys that don't match the expected schemas and verification would still pass, as long as at least one key successfully verifies a signature.

in_toto/verifylib.py
```diff
@@ -1465,18 +1450,22 @@ def in_toto_verify(layout, layout_key_dict, link_dir_path=".",
...
-    useful during recursive sublayout verification.
+    A Link object, which summarizes the materials and products of the overall
+    software supply chain. This is mostly useful during recursive sublayout
+    verification.
 
   """
-  LOG.info("Verifying layout signatures...")
-  verify_layout_signatures(layout, layout_key_dict)
+
+  keys = create_key_list_from_key_dict(layout_key_dict)
```
lukpueh: Same comment as above in `in_toto_sign.py` applies here. By silently dropping keys in `create_key_list_from_key_dict`, the expected threshold can be subverted without noticing.

in_toto/models/metadata.py
lukpueh: Speaking of gpg keys, I don't think that `Metablock.verify_signatures` works with gpg subkeys as `Metablock.verify_signature` does. Can you confirm? 

Maybe we can move this to `_signer.GPGKey.verify`.

lukpueh (changes requested):
Thanks the updates, @PradyumnaKrishna! I left a bunch of comments. Mostly minor things about documentation being too generic or specific for Metadata, Metablock, Envelope.

I also left a few more important comments about signature verification. We need to be very careful to not degrade any security properties.

I'll follow up with another review of tests, and a thorough comparison of `Metablock.verify_signature` and `Metablock.verify_signatures`.

in_toto/models/metadata.py
PradyumnaKrishna: Yes, `verify_signatures` doesn't raises exceptions and only returns a dict of accepted keys, where `verify_signature` raise exception if the verification fails or gpg key is expired. Only case we are missing is of expired `GPGKey`.

in_toto/models/metadata.py
PradyumnaKrishna: Speaking of `compact_json`, I found, it was actually never used.
It was only referenced in `in_toto_run` at https://github.com/in-toto/in-toto/blob/fbc1eb053dd39c6d1fb307a60bd2c781e667c931/in_toto/runlib.py#L658-L662

which again is always `False` because no command line argument is provided to use this option.
https://github.com/in-toto/in-toto/blob/fbc1eb053dd39c6d1fb307a60bd2c781e667c931/in_toto/runlib.py#L512-L516
https://github.com/in-toto/in-toto/blob/fbc1eb053dd39c6d1fb307a60bd2c781e667c931/in_toto/in_toto_run.py#L221-L227

PradyumnaKrishna: The new `Metadata.dump()` method takes care of compact json using optional argument `compact` that I forgot to mention in doc-string.

in_toto/verifylib.py
PradyumnaKrishna: I guess raising exception or logging it might help.

lukpueh:
So it turns out we call `Metadata.verify_signatures(keys:  List[Key], threshold: int)` either as `verify_signatures([key], 1)` or `verify_signatures(keys, len(keys))`. This is equivalent to `verify_signature(key)` and `for key in keys: verify_signature(key)` respectively.

What do you think about using a `Metadata.verify_signature(key: dict)` interface instead?

Then we wouldn't need to add a new Metablock signature verification implementation and wouldn't have to worry if it changes the behavior (e.g. if it still supports gpg subkeys etc.). 

We could also drop the `create_key_list_from_key_dict` helper. It seems easier to just add a compatible method to the in-toto Envelope, e.g.:

```python
class Envelope(BaseEnvelope):
   # ....
   def verify_signature(key: dict): 
      key = Key.from_dict(key["keyid"], key)
      BaseEnvelope.verify(self,  key, 1)
```

in_toto/models/metadata.py
lukpueh: What are your thoughts on this? Again, for the sake of backwards-compatibility, we could keep `SignatureVerificationError` and, for the sake of a consistent interface, change `Envelope.verify_signature` to something like:
```python
# ...
except VerificationError:
    raise SignatureVerificationError
# ...
```

in_toto/verifylib.py
```diff
@@ -39,19 +39,61 @@
...
+from securesystemslib.exceptions import VerificationError
+from securesystemslib.gpg.exceptions import KeyExpirationError
+
 # Inherits from in_toto base logger (c.f. in_toto.log)
 LOG = logging.getLogger(__name__)
 
 RULE_TRACE = {}
 
 
+def verify_metadata_signatures(metadata: Metadata, keys_dict):
```
lukpueh: For the sake of a smaller diff you could just keep the original `verify_layout_signatures` and only update docs and variable names to  `s/metablock/metadata`, right?

lukpueh (commented):
Thanks for the updates! I'm excited about the diff getting smaller and smaller. 💯

in_toto/models/metadata.py
```diff
@@ -123,10 +259,18 @@ def load(path):
       signed = Layout.read(signed_data)
 
     else:
-      raise securesystemslib.exceptions.FormatError("Invalid Metadata format")
+      raise InvalidMetadata("Invalid Metadata format")
```
lukpueh: It looks like we can maintain backwards-compatibility in this PR. Let's the keep the original FormatError here, shall we?

in_toto/models/metadata.py
lukpueh: Well, we do expose both [`Metadata.compact_json`](https://in-toto.readthedocs.io/en/latest/model.html#in_toto.models.metadata.Metablock.compact_json) and [`runlib.in_toto_run`](https://in-toto.readthedocs.io/en/latest/api.html#in_toto.runlib.in_toto_run) as public API, so the change would be an API break, and technically require a major release.

While I'm not opposed to a new major release, it looks like this PR could otherwise be fully backwards-compatible, so it seems a bit silly to do it just because of this feature.

We could:
- remove the `compact` argument from `Metadata.dump`, and
- continue to use the `compact_json` attribute in `Metablock.dump`, and
- just not support this feature in `Envelope.dump` for now.

What do you think? 

(Note that compacting the envelope alone does not get us far anyway, because it ignores the base64-encoded payload, which is likely to contain way more whitespace than the envelope.)

tests/models/test_envelope.py
lukpueh: ping https://github.com/in-toto/in-toto/pull/503/files#r1022835153

lukpueh (changes requested):
I reviewed everything now including tests and this could be the very last request for changes.

The two inline comments (compact_json, InvalidMetadata vs. FormatError) about backwards-compatibility are the important ones.

In addition, I have some comments on return type annotations:

- `Metadata.from_dict` annotates `Envelope | Metablock`, and `Metablock.load` annotates `Metadata`. I think they should annotate the same return type, and either type seems fine to me.

- `Envelope.from_dict` annotates  `securesystemslib.dsse.Envelope`. I think it should be the in-toto `Envelope`.

- `Envelope.load` and `Metablock.load` annotate `Metadata`. I think it should be `Envelope` and `Metablock` respectively.

Alternatively, if it's easier, you are also free to not introduce type annotations at all in this PR. We don't use them anywhere else, and aren't checking them yet.

lukpueh:
@PradyumnaKrishna, I just released your securesystemslib DSSE implementation. Let's update here and get this merged!! ❤️ 🚀

lukpueh (approved):
Thanks for your persistence, @PradyumnaKrishna! 💯

adityasaky:
Fantastic work, @PradyumnaKrishna!

Issue #435, opened 2021-02-26
GitHub Actions: Enable windows builds


lukpueh:
**Description of issue or feature request:**
Configure recently added GitHub Actions workflow (see #432) to run tests on windows too.

**Current behavior:**
tests run on linux and mac

**Expected behavior:**
tests run on linux, mac and windows

Coordinate with https://github.com/secure-systems-lab/securesystemslib/issues/326 and https://github.com/secure-systems-lab/securesystemslib/issues/337.

Uttam-Singhh:
Hello, @lukpueh I would like to work on this can you please guide me!

adityasaky:
Hey @Uttam-Singhh! Let me try and track down some work. I know @SolidifiedRay was working on this and had some failures on Windows. Am I misremembering, Yuanrui? :)

SolidifiedRay:
@adityasaky Yes, there are some test failures on Windows.

adityasaky:
Do you have a branch you could drop here to get @Uttam-Singhh started? :)

SolidifiedRay:
Yes. https://github.com/SolidifiedRay/in-toto/blob/gh-actions-windows/.github/workflows/ci.yml
This file is not very different from the one we are using now. The only difference is that my branch  runs the tests on windows machines and not on Linux or Mac.

Uttam-Singhh:
Am I good to go?

adityasaky:
Yes please, feel free to take a stab at it, while also taking a look at @SolidifiedRay's branch linked above. Thanks!

lukpueh:
I think we can finally close this with https://github.com/in-toto/in-toto/pull/598

PR #513, opened 2022-10-27
Add support for Windows builds

Rishit-dagli:
that is provided, the better.

**Fixes issue #435**

Adds Windows to the list of OSes on which the action runs.

requirements**:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,10 @@ jobs:
       # on Linux/Python3.x only.
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10']
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         toxenv: [py]
         include:
-          # TODO: Change to 3.x once pylint fully supports Python 3.9
-          - python-version: 3.8
+          - python-version: '3.x'
             os: ubuntu-latest
             toxenv: lint
 


.github/workflows/ci.yml
```diff
@@ -14,10 +14,19 @@ jobs:
         toxenv: [py]
         include:
           # TODO: Change to 3.x once pylint fully supports Python 3.9
-          - python-version: 3.8
+          - python-version: '3.8'
             os: ubuntu-latest
             toxenv: lint
-
+          - python-version: '3.7'
+            os: windows-latest
```
adityasaky: Thanks for working on this! Isn't it sufficient to add windows-latest to the os field above? We want lint outside because running it on one environment is fine.

adityasaky (changes requested):

.github/workflows/ci.yml
Rishit-dagli: So for some reason, it seems that Windows builds fail on Python 3.9 but work well with all other Python versions which is why I explicitly specify the versions it works with. I discussed some possible causes in the Contribfest but this was one of the possible solutions to that. What do you think?

adityasaky: Huh odd. @lukpueh what do you think?

.github/workflows/ci.yml
```diff
@@ -49,4 +58,4 @@ jobs:
           python -m pip install tox
 
       - name: Run tox
-        run: tox -e ${{ matrix.toxenv }}
+        run: tox -e ${{ matrix.toxenv }}
```
lukpueh: Any reason for this change?

.github/workflows/ci.yml
lukpueh: That's odd. Do you have an error log? Are you sure that it was because of 3.9? We do have some flakey tests. Maybe it failed on 3.9 by coincidence.

lukpueh (commented):
Thanks, this is great! We can always troubleshoot 3.9 in a follow up. Testing 4 of 5 supported Python versions on Windows is substantially better than not testing on Windows at all. :) 

LGTM, modulo some minor nit about the removed newline at EOF.

.github/workflows/ci.yml
adityasaky: Looks like a coincidence. I had issues on 3.8 and 3.10 just now and re-run of failed jobs cleared it up (https://github.com/adityasaky/in-toto/actions/runs/3858306634/jobs/6576722830). My suggestion is we leave this enabled alongside AppVeyor for the time being and use AV's Windows results over Github Actions when GA fails. Long term, we should study which tests fail on Windows and why they're flakey in that environment specifically.

adityasaky:
I think we should merge this PR. My suggestion is we leave windows-latest enabled alongside AppVeyor for the time being and use AV's Windows results over Github Actions when GA fails. Long term, we should study which tests fail (and if there's a pattern) on Windows and why they're flakey in that environment specifically. We don't need to mark windows on GA as required for branch protection right now.

adityasaky (approved):

lukpueh (approved):

PR #529, opened 2023-01-12
tests: adopt replacement of HAVE_GPG constant

lukpueh:
*prepare for upcoming securesystemslib release (https://github.com/secure-systems-lab/securesystemslib/issues/475)*

---

secure-systems-lab/securesystemslib#437 replaced the HAVE_GPG bool constant with a function with similar effect, to allow lazy evaluation of whether GPG is present.

That is, we only shell out to check if gpg is present, when we call have_gpg() and not when we import the module, where the name is defined, which was the case before.

This commit adopts the rename in in-toto tests. No other code in in-toto is affected by the rename.


adityasaky (approved):

lukpueh:
Test failures are unrelated and documented here: https://github.com/secure-systems-lab/securesystemslib/issues/337

Should we merge regardless? @adityasaky

lukpueh:
> https://github.com/secure-systems-lab/securesystemslib/issues/337

lukpueh:
Re-build passed. 🤷 

I think we can merge, but we should really figure out a fix for https://github.com/secure-systems-lab/securesystemslib/issues/337 at some point. @SolidifiedRay, would you be interested in taking another stab at that issue?

adityasaky:
We should log which tests fail, I'm curious to see if there's a pattern or if the problem is elsewhere.

SolidifiedRay:
> Re-build passed. 🤷
> 
> I think we can merge, but we should really figure out a fix for [secure-systems-lab/securesystemslib#337](https://github.com/secure-systems-lab/securesystemslib/issues/337) at some point. @SolidifiedRay, would you be interested in taking another stab at that issue?

Sure, I can take a look

PR #535, opened 2023-01-18
Allow `command`, `byproducts`, and `environment` to be optionally set…

### LINKED ISSUE #531: allow setting `command` and `byproducts` with `in_toto_record_start` and `in_toto_record_stop`

We're prototyping using the `in-toto` python package to create layouts and links for [hoppr](https://gitlab.com/hoppr/hoppr).

We're wrapping key steps in the processing with `in_toto.runlib.in_toto_record_start()` and `in_toto.runlib.in_toto_record_stop()`.  The link files look good and validate against the layout, but a lot of the fields are empty.
- `command`
- `byproducts`
- `environment`

When using `in_toto.runlib.in_toto_run()` a terminal command is an argument which is set to the `run` field on the link file. The stdout/stderr is set to the `byproducts`.

There is no method for setting these when using `in_toto.runlib.in_toto_record_start()` and `in_toto.runlib.in_toto_record_stop()`. As of `v1.2.0` this values are set empty. Ideally these would be optional to set when calling in_toto.runlib.in_toto_record_stop.

**Current behavior**:

The current behavior sets these items as empty lists/dicts on `in_toto.runlib.in_toto_record_start()`

- see https://github.com/in-toto/in-toto/blob/6bd25002ef2d930b214288c2d8c0ef3f2a917473/in_toto/runlib.py#L715

**Expected behavior**:

Create default arguments for `in_toto.runlib.in_toto_record_stop()` which would not break the existing functionality, but allow users to pass in `command`, `byproducts`, and `environment` link data.

I'm happy to submit a PR if there is interest.

idunbarh:
… when using `in_toto.runlib.in_toto_record_start()` and `in_toto.runlib.in_toto_record_stop()`, fixes #531

Signed-off-by: Ian Dunbar-Hall <ian.dunbar-hall@lmco.com>

that is provided, the better.

**Fixes issue #531 **:

This PR is to allow `command`, `byproducts`, and `environment` to be optionally set with using in_toto.runlib.in_toto_record_start()` and `in_toto.runlib.in_toto_record_stop()` to wrap commands.

Since these new arguments are optional, it should not affect existing users of the python package.

Additionally, I updated the `Link._validate_environment()` to validate that any dictionary keys can only be `variables`, `filesystem`, or `workdir`. Keys not on this list will cause an exception to be raised. 

Let me know if you think proactive enforcement of the spec for the `environment` section of metadata is not a good idea and I'll remove it.

requirements**:

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -31,6 +31,11 @@ import glob
 import logging
 import os
 import itertools
+import io
+import subprocess  # nosec
+import sys
+import tempfile
+import time
 
 from pathspec import PathSpec
 
@@ -43,7 +48,6 @@ from in_toto.models.link import (UNFINISHED_FILENAME_FORMAT, FILENAME_FORMAT,
 import securesystemslib.formats
 import securesystemslib.hash
 import securesystemslib.exceptions
-import securesystemslib.process
 import securesystemslib.gpg
 
 from in_toto.models.metadata import Metablock
@@ -322,6 +326,86 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
 
   return artifacts_dict
 
+def _subprocess_run_duplicate_streams(cmd, timeout):
+  """Helper to run subprocess and both print and capture standards streams.
+
+  Caveat:
+  * Might behave unexpectedly with interactive commands.
+  * Might not duplicate output in real time, if the command buffers it (see
+    e.g. `print("foo")` vs. `print("foo", flush=True)`).
+  * Possible race condition on Windows when removing temporary files.
+
+  """
+  # Use temporary files as targets for child process standard stream redirects
+  # They seem to work better (i.e. do not hang) than pipes, when using
+  # interactive commands like `vi`.
+  stdout_fd, stdout_name = tempfile.mkstemp()
+  stderr_fd, stderr_name = tempfile.mkstemp()
+  try:
+    with io.open(  # pylint: disable=unspecified-encoding
+      stdout_name, "r"
+    ) as stdout_reader, os.fdopen(  # pylint: disable=unspecified-encoding
+      stdout_fd, "w"
+    ) as stdout_writer, io.open(  # pylint: disable=unspecified-encoding
+      stderr_name, "r"
+    ) as stderr_reader, os.fdopen(
+      stderr_fd, "w"
+    ) as stderr_writer:
+
+      # Store stream results in mutable dict to update it inside nested helper
+      _std = {"out": "", "err": ""}
+
+      def _duplicate_streams():
+        """Helper to read from child process standard streams, write their
+        contents to parent process standard streams, and build up return values
+        for outer function.
+        """
+        # Read until EOF but at most `io.DEFAULT_BUFFER_SIZE` bytes per call.
+        # Reading and writing in reasonably sized chunks prevents us from
+        # subverting a timeout, due to being busy for too long or indefinitely.
+        stdout_part = stdout_reader.read(io.DEFAULT_BUFFER_SIZE)
+        stderr_part = stderr_reader.read(io.DEFAULT_BUFFER_SIZE)
+        sys.stdout.write(stdout_part)
+        sys.stderr.write(stderr_part)
+        sys.stdout.flush()
+        sys.stderr.flush()
+        _std["out"] += stdout_part
+        _std["err"] += stderr_part
+
+      # Start child process, writing its standard streams to temporary files
+      proc = subprocess.Popen(  # pylint: disable=consider-using-with  # nosec
+        cmd,
+        stdout=stdout_writer,
+        stderr=stderr_writer,
+        universal_newlines=True,
+      )
+      proc_start_time = time.time()
+
+      # Duplicate streams until the process exits (or times out)
+      while proc.poll() is None:
+        # Time out as Python's `subprocess.run` would do it
+        if (
+          timeout is not None
+          and time.time() > proc_start_time + timeout
+        ):
+          proc.kill()
+          proc.wait()
+          raise subprocess.TimeoutExpired(cmd, timeout)
+
+        _duplicate_streams()
+
+      # Read/write once more to grab everything that the process wrote between
+      # our last read in the loop and exiting, i.e. breaking the loop.
+      _duplicate_streams()
+
+  finally:
+    # The work is done or was interrupted, the temp files can be removed
+    os.remove(stdout_name)
+    os.remove(stderr_name)
+
+  # Return process exit code and captured streams
+  return proc.poll(), _std["out"], _std["err"]
+
 def execute_link(link_cmd_args, record_streams):
   """
   <Purpose>
@@ -343,10 +427,9 @@ def execute_link(link_cmd_args, record_streams):
     OSError:
             The given command is not present or non-executable
 
-    securesystemslib.process.subprocess.TimeoutExpired:
-            The execution of the given command times out. The default timeout
-            is securesystemslib.settings.SUBPROCESS_TIMEOUT.
-
+    subprocess.TimeoutExpired:
+            The execution of the given command times (see
+            in_toto.settings.LINK_CMD_EXEC_TIMEOUT).
 
   <Side Effects>
     Executes passed command in a subprocess and redirects stdout and stderr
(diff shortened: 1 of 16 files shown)


idunbarh:
@adityasaky when you have a moment, can you provide feedback on this PR? This is closes #531.

idunbarh:
@adityasaky looks like there were a couple linting issues. These were corrected and looking for the workflow to be approved a second time.

in_toto/models/link.py
```diff
@@ -181,3 +181,9 @@ def _validate_environment(self):
       raise securesystemslib.exceptions.FormatError(
           "Invalid Link: field `environment` must be of type dict, got: {}"
           .format(type(self.environment)))
+    for key in self.environment.keys():
```
adityasaky: We should remove this, the specification doesn't mandate a format for `environment` and `byproducts`. https://github.com/in-toto/docs/blob/master/in-toto-spec.md#441-environment-information

in_toto/runlib.py
```diff
@@ -708,13 +723,12 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
       lstrip_paths=lstrip_paths)
 
   LOG.info("Creating preliminary link metadata...")
-  environment = {}
   if record_environment:
```
adityasaky: I think _one_ of `environment` and `record_environment` should be used. Any practical danger `record_environment` overwrites the `workdir` passed in explicitly using `environment`?

adityasaky: cc @lukpueh

in_toto/runlib.py
```diff
@@ -784,6 +798,10 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
     base_path (optional): A path relative to which artifacts are recorded.
         Default is the current working directory.
 
+    byproducts (optional): A dictionary that lists byproducts of the link
```
adityasaky: Curious, at some point here, we stop needing in-toto-record, correct? All this information is available and could be plugged into a `Link` object manually as well?

adityasaky (changes requested):
Some mostly nits, some discussion points, but overall this looks good to me. Thanks for the PR!

in_toto/models/link.py
idunbarh: reverted to original and removed this validation.

in_toto/runlib.py
idunbarh: Interested to hear @lukpueh's thoughts.

in the meantime, added argument validation which raises `securesystemslib.exceptions.FormatError` if both are set. Happy to revert depending on the discussion.

```
  if environment and record_environment:
        "The arguments environment and record_environment are mutually"
        " exclusive. Please use one or the other.")
```

in_toto/runlib.py
idunbarh: At least for me, the advantage of `in_toto.runlib.in_toto_record_start()` and `in_toto.runlib.in_toto_record_stop()` is handling creating and reading of the link files and key management. 

Sure ... its _not a ton_ of code, but I'd find myself copying 10's lines of `in_toto.runlib` code into the hoppr code base.

adityasaky: Makes sense, thanks!

in_toto/runlib.py
```diff
@@ -641,8 +641,20 @@ def in_toto_record_start(step_name, material_list, signing_key=None,
     base_path (optional): A path relative to which artifacts are recorded.
         Default is the current working directory.
 
+    command (optional): A list consisting of a command and arguments executed
```
adityasaky: I can see a slight case for this to go into `stop`, while we're pinging @lukpueh. But I don't think it's a big deal and we generally want to lean towards users knowing what command(s) they're running at `start` time. Definitely not something to block this PR!

adityasaky (commented):
LGTM, aside from the open thread re `environment` and `record_environment` being used mutually. While we're at it, should `environment` also be added to `in_toto_run` do you think? It also currently only has `record_environment`.

idunbarh:
> LGTM, aside from the open thread re `environment` and `record_environment` being used mutually. While we're at it, should `environment` also be added to `in_toto_run` do you think? It also currently only has `record_environment`.

Happy to add this to `in_toto_run`.  Would you want the addition of environment available in the CLI too? 

My previous commits only expose this options if you are importing `in_toto.runlib` and not if you are using the CLI.

adityasaky:
> Would you want the addition of environment available in the CLI too?

I'm concerned that a full fledged environment flag would get clunky, would we have users pass in key value pairs? Reminded of docker's CLI for setting environment vars.

in_toto/runlib.py
```diff
@@ -602,8 +602,8 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
 
 def in_toto_record_start(step_name, material_list, signing_key=None,
     gpg_keyid=None, gpg_use_default=False, gpg_home=None,
-    exclude_patterns=None, base_path=None, record_environment=False,
-    normalize_line_endings=False, lstrip_paths=None):
+    exclude_patterns=None, base_path=None, command=None, environment=None,
+    record_environment=False, normalize_line_endings=False, lstrip_paths=None):
```
lukpueh: I think we should move new kwargs `command` and `environment` to the end, to not break existing users, who call this function without keywords.

Same goes for new kwarg in `in_toto_record_stop`.

in_toto/runlib.py
lukpueh: Why not move all new arguments to `stop`? This would make the change quite a bit smaller.

in_toto/runlib.py
lukpueh: IMO a passed `environment` should override the recorded environment. This should allow for the same use cases, and we wouldn't need any mutual exclusion handling.

lukpueh (changes requested):
Thanks for the patch, @idunbarh. It think this looks mostly good, but I'd love if we could extend the API as little as possible to support your use case.

I noted one blocker (order of arguments), and two suggestions (move all to `stop`, override recorded env) inline.

in_toto/runlib.py
idunbarh: Sorry for the delay getting this PR updated.

Updated for override the recorded environment which `environment` is provided to `in_toto.runlib.in_toto_record_stop()` function.

in_toto/runlib.py
idunbarh: I moved all of the args to `in_toto.runlib.in_toto_record_stop()` and placed them at the end of the argument list.

idunbarh:
> Thanks for the patch, @idunbarh. It think this looks mostly good, but I'd love if we could extend the API as little as possible to support your use case.
> 
> I noted one blocker (order of arguments), and two suggestions (move all to `stop`, override recorded env) inline.

@lukpueh great suggestion. Yeah, thats definitely simplifies moving everything to `in_toto.runlib.in_toto_record_stop()`. Updated the PR to have everything in `in_toto.runlib.in_toto_record_stop()` and add a unittest.

in_toto/runlib.py
```diff
@@ -880,6 +881,22 @@ def in_toto_record_stop(step_name, product_list, signing_key=None,
...
+    command (optional): A list consisting of a command and arguments executed
+        between in_toto_record_start() and in_toto_record_stop() to capture
+        the command ran in the resulting link metadata.
+
+    byproducts (optional): A dictionary that lists byproducts of the link
+        command execution. It should have at least the following entries
+        "stdout" (str), "stderr" (str) and "return-value" (int).
+
+    environment (optional): A dictionary to capture information about
+        the environment to be added in the resulting link metadata eg.::
```
lukpueh, line 893:
```suggestion
        the environment to be added in the resulting link metadata eg.::

```
Minor nit: sphincs requires a newline here, otherwise there's this error in the doc build log
```
docstring of in_toto.runlib.in_toto_record_stop:56: ERROR: Unexpected indentation.
```

Interestingly enough, the html build still finishes and the code block is rendered correctly. 🤷

lukpueh (changes requested):
Thanks for the updates! Please fix linting error and address inline comment, and we can merge this!

adityasaky (approved):
Aside from what's already been pointed out, LGTM. @idunbarh, could you rebase / squash and get rid of the merge commit? Thanks!

PR #536, opened 2023-01-19
Implement ITE-4 Generic URI Resolver Interface

alanssitis:
Implementation of [ITE-4](https://github.com/in-toto/ITE/blob/master/ITE/4/README.adoc) resolver interface. Using a python class interface to accomplish the requirements for ITE-4.

requirements**:

### DIFF
--- a/in_toto/resolver/__init__.py
+++ b/in_toto/resolver/__init__.py
@@ -28,10 +28,10 @@ from in_toto.resolver.resolver import (
 )
 
 RESOLVER_FOR_URI_SCHEME.update(
-    {
-      DEFAULT_SCHEME: FileResolver,
-      FileResolver.SCHEME: FileResolver,
-    }
+  {
+    DEFAULT_SCHEME: FileResolver,
+    FileResolver.SCHEME: FileResolver,
+  }
 )
 
 def set_params(**kwargs):
--- a/in_toto/resolver/__init__.py
+++ b/in_toto/resolver/__init__.py
@@ -23,7 +23,6 @@ from in_toto.resolver.file_resolver import FileResolver
 from in_toto.resolver.resolver import (
   DEFAULT_SCHEME,
   RESOLVER_FOR_URI_SCHEME,
-  RESOLVER_PARAMS,
   Resolver,
 )
 
@@ -33,3 +32,9 @@ RESOLVER_FOR_URI_SCHEME.update(
       FileResolver.SCHEME: FileResolver,
     }
 )
+
+def resolver_set_params(**kwargs):
+  """Set the parameters for the resolvers."""
+  FileResolver.follow_symlink_dirs = kwargs.get("follow_symlink_dirs", False)
+  FileResolver.normalize_line_endings = kwargs.get("normalize_line_endings",
+                                                   False)
--- a/in_toto/resolver/__init__.py
+++ b/in_toto/resolver/__init__.py
@@ -33,9 +33,3 @@ RESOLVER_FOR_URI_SCHEME.update(
     FileResolver.SCHEME: FileResolver,
   }
 )
-
-def set_params(**kwargs):
-  """Set the parameters for the resolvers."""
-  FileResolver.follow_symlink_dirs = kwargs.get("follow_symlink_dirs", False)
-  FileResolver.normalize_line_endings = kwargs.get("normalize_line_endings",
-                                                   False)
--- a/in_toto/resolver/__init__.py
+++ b/in_toto/resolver/__init__.py
@@ -1,5 +1,22 @@
+# Copyright New York University and the in-toto contributors
+# SPDX-License-Identifier: Apache-2.0
+
 """
-resolver contains all default artifact resolvers
+<Program Name>
+  __init__.py
+
+<Author>
+  Alan Chung Ma <achungma@purdue.edu>
+
+<Started>
+  February 1, 2023
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Build resolver scheme dictionary and set default resolver.
+
 """
 
 from in_toto.resolver._file_resolver import FileResolver
--- a/in_toto/resolver/__init__.py
+++ b/in_toto/resolver/__init__.py
@@ -24,6 +24,7 @@ from in_toto.resolver.resolver import (
   DEFAULT_SCHEME,
   RESOLVER_FOR_URI_SCHEME,
   Resolver,
+  get_scheme,
 )
 
 RESOLVER_FOR_URI_SCHEME.update(
@@ -33,7 +34,7 @@ RESOLVER_FOR_URI_SCHEME.update(
     }
 )
 
-def resolver_set_params(**kwargs):
+def set_params(**kwargs):
   """Set the parameters for the resolvers."""
   FileResolver.follow_symlink_dirs = kwargs.get("follow_symlink_dirs", False)
   FileResolver.normalize_line_endings = kwargs.get("normalize_line_endings",
--- a/in_toto/resolver/__init__.py
+++ b/in_toto/resolver/__init__.py
@@ -1,0 +1,17 @@
+"""
+resolver contains all default artifact resolvers
+"""
+
+from in_toto.resolver._file_resolver import FileResolver
+from in_toto.resolver._resolver import (
+  DEFAULT_SCHEME,
+  RESOLVER_FOR_URI_SCHEME,
+  Resolver,
+)
+
+RESOLVER_FOR_URI_SCHEME.update(
+    {
+      DEFAULT_SCHEME: FileResolver,
+      FileResolver.SCHEME: FileResolver,
+    }
+)
--- a/in_toto/resolver/__init__.py
+++ b/in_toto/resolver/__init__.py
@@ -19,10 +19,11 @@
 
 """
 
-from in_toto.resolver._file_resolver import FileResolver
-from in_toto.resolver._resolver import (
+from in_toto.resolver.file_resolver import FileResolver
+from in_toto.resolver.resolver import (
   DEFAULT_SCHEME,
   RESOLVER_FOR_URI_SCHEME,
+  RESOLVER_PARAMS,
   Resolver,
 )
 
--- a/in_toto/resolver/_resolver.py
+++ b/in_toto/resolver/_resolver.py
@@ -9,7 +9,7 @@ RESOLVER_FOR_URI_SCHEME = {}
 
 
 def _get_scheme(uri):
-  match = re.fullmatch(r"(\w+\:)?(.*)", uri)
+  match = re.fullmatch(r"(\w+\:)?(.+)", uri)
   if not match:
     raise ValueError(f"Artifact URI '{uri}' could not be parsed")
   groups = match.groups()
@@ -30,15 +30,13 @@ class Resolver(metaclass=ABCMeta):
   """Interface for resolvers"""
 
   @classmethod
-  def apply_exclude_patterns(cls, names, exclude_filter):
+  def apply_exclude_patterns(cls, names, exclude_patterns):
     """Exclude matched patterns from passed names."""
     included = set(names)
 
-    # Assume old way for easier testing
-    if hasattr(exclude_filter, '__iter__'):
-      exclude_filter = PathSpec.from_lines('gitwildmatch', exclude_filter)
+    exclude_patterns = PathSpec.from_lines('gitwildmatch', exclude_patterns)
 
-    for excluded in exclude_filter.match_files(names):
+    for excluded in exclude_patterns.match_files(names):
       included.discard(excluded)
 
     return sorted(included)
--- a/in_toto/resolver/_resolver.py
+++ b/in_toto/resolver/_resolver.py
@@ -19,7 +19,7 @@ def _get_scheme(uri):
 
 def _get_resolver(uri):
   scheme = _get_scheme(uri)
-  
+
   if scheme not in RESOLVER_FOR_URI_SCHEME:
     raise ValueError(f"Unsupported in-toto resolver scheme '{scheme}'")
 
@@ -45,15 +45,17 @@ class Resolver(metaclass=ABCMeta):
 
   @classmethod
   @abstractmethod
-  def resolve_uri(cls, generic_uri, exclude_patterns=None, follow_symlink_dirs=False):
+  def resolve_uri(cls, generic_uri, exclude_patterns=None,
+                  follow_symlink_dirs=False):
     """Normalize and resolve artifact URIs"""
     resolver = _get_resolver(generic_uri)
-    return resolver.resolve_uri(generic_uri, exclude_patterns, follow_symlink_dirs)
+    return resolver.resolve_uri(
+        generic_uri, exclude_patterns, follow_symlink_dirs)
 
   @classmethod
   @abstractmethod
-  def get_hashable_representation(cls, resolved_uri, normalize_line_endings=False):
+  def get_hashable_representation(cls, resolved_uri,
+                                  normalize_line_endings=False):
     """Return dictionary of hash digests"""
     resolver = _get_resolver(resolved_uri)
     return resolver.hash_artifacts(resolved_uri, normalize_line_endings)
-
--- a/in_toto/resolver/_resolver.py
+++ b/in_toto/resolver/_resolver.py
@@ -1,0 +1,59 @@
+"""Resolver interface"""
+
+from abc import ABCMeta, abstractmethod
+from pathspec import PathSpec
+import re
+
+DEFAULT_SCHEME = "default"
+RESOLVER_FOR_URI_SCHEME = {}
+
+
+def _get_scheme(uri):
+  match = re.fullmatch(r"(\w+\:)?(.*)", uri)
+  if not match:
+    raise ValueError(f"Artifact URI '{uri}' could not be parsed")
+  groups = match.groups()
+  if not groups[0]:
+    return DEFAULT_SCHEME
+  return groups[0][:-1]
+
+def _get_resolver(uri):
+  scheme = _get_scheme(uri)
+  
+  if scheme not in RESOLVER_FOR_URI_SCHEME:
+    raise ValueError(f"Unsupported in-toto resolver scheme '{scheme}'")
+
+  return RESOLVER_FOR_URI_SCHEME[scheme]
+
+
+class Resolver(metaclass=ABCMeta):
+  """Interface for resolvers"""
+
+  @classmethod
+  def apply_exclude_patterns(cls, names, exclude_filter):
+    """Exclude matched patterns from passed names."""
+    included = set(names)
+
+    # Assume old way for easier testing
+    if hasattr(exclude_filter, '__iter__'):
+      exclude_filter = PathSpec.from_lines('gitwildmatch', exclude_filter)
+
+    for excluded in exclude_filter.match_files(names):
+      included.discard(excluded)
+
+    return sorted(included)
+
+  @classmethod
+  @abstractmethod
+  def resolve_uri(cls, generic_uri, exclude_patterns=None, follow_symlink_dirs=False):
+    """Normalize and resolve artifact URIs"""
+    resolver = _get_resolver(generic_uri)
+    return resolver.resolve_uri(generic_uri, exclude_patterns, follow_symlink_dirs)
+
+  @classmethod
+  @abstractmethod
+  def get_hashable_representation(cls, resolved_uri, normalize_line_endings=False):
+    """Return dictionary of hash digests"""
+    resolver = _get_resolver(resolved_uri)
+    return resolver.hash_artifacts(resolved_uri, normalize_line_endings)
+
--- a/in_toto/resolver/_resolver.py
+++ b/in_toto/resolver/_resolver.py
@@ -1,4 +1,27 @@
-"""Resolver interface"""
+# Copyright New York University and the in-toto contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+<Program Name>
+  _resolver.py
+
+<Author>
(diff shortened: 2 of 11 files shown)


alanssitis:
@adityasaky here's what I have so far! To "lightly test" it I ran it against the ite-4-demo and it worked. Let me know if this is how you envisioned it.

alanssitis:
@lukpueh could you take a look at this?

lukpueh:
> @lukpueh could you take a look at this?

Happy to! A description of the PR and its status (e.g. why tests are currently failing) would be helpful.

alanssitis:
> > @lukpueh could you take a look at this?
> 
> Happy to! A description of the PR and its status (e.g. why tests are currently failing) would be helpful.

I updated it with a brief description. The tests are probably failing due to a bug in my implementation of the regular file resolver. I was wondering if you could take a glance over the interface and methods I currently have for the other resolvers and the way I have implemented them.

I will need to change how tests are handled and move a few files around.

pyproject.toml
```diff
@@ -40,6 +40,9 @@ dependencies = [
     "pathspec",
     "python-dateutil",
     "securesystemslib[crypto]>=0.18.0",
+    "urllib3",
+    "docker",
+    "podman",
```
lukpueh: I think these should be optional dependencies.

in_toto/resolver/_resolver.py
```diff
@@ -0,0 +1,61 @@
...
+  if not groups[0]:
+    return DEFAULT_SCHEME
+  return groups[0][:-1]
+
+
+class Resolver(metaclass=ABCMeta):
+  """Interface for resolvers"""
+
+  @classmethod
+  def apply_exclude_patterns(cls, names, exclude_filter):
```
lukpueh: This is currently only used by `FileResolver`. Shouldn't exclude patterns also apply to other URIs?

in_toto/resolver/_resolver.py
```diff
@@ -0,0 +1,61 @@
...
+  def hash_artifacts(cls, resolved_uri, hash_algorithms=None, **kwargs):
+    """Return dictionary of hash digests"""
+    scheme = _get_scheme(resolved_uri)
+
+    if scheme not in RESOLVER_FOR_URI_SCHEME:
+      raise ValueError(f"Unsupported in-toto resolver scheme '{scheme}'")
+    resolver = RESOLVER_FOR_URI_SCHEME[scheme]
+
+    return resolver.hash_artifacts(
+        resolved_uri, hash_algorithms=hash_algorithms, **kwargs)
```
lukpueh: There's a bit of code duplication here that you could factor out into a helper.

lukpueh: Also, resolve URI seems more like an implementation detail. Does this have to be part of the interface? It's a no-op for all resolvers of this PR except for the file resolver.

lukpueh: On a more general note: The ITE requires resolvers to implement `get_hashable_representation` AND `hash_artifacts`. Are really both necessary?  If each resolver does implement `get_hashable_representation` then the hashing does not have to be a resolver task. OTOH, if we only ever need the hash, then `get_hashable_representation` could also be an implementation detail of `hash_artifacts`. Looking at the existing resolvers, the latter seems to apply.

lukpueh: cc @adityasaky

in_toto/resolver/_resolver.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if scheme not in RESOLVER_FOR_URI_SCHEME:
+      raise ValueError(f"Unsupported in-toto resolver scheme '{scheme}'")
+    resolver = RESOLVER_FOR_URI_SCHEME[scheme]
+
+    return resolver.resolve_uri(generic_uri, **kwargs)
+
+  @classmethod
+  @abstractmethod
+  def hash_artifacts(cls, resolved_uri, hash_algorithms=None, **kwargs):
```
lukpueh: I don't think optional arguments that differ for different resolvers are very useful here. If I need to know, which resolver I will be using, there is no need to dispatch based on the URI, right? What about something:

```python
resolver = Resolver.from_uri(uri)

if isinstance(resolver, FileResolver):
  resolver.hash_artifacts(uri, normalize_line_endings=True) 

else:
  ...
```

lukpueh (commented):
I think this is a good start for ITE-4!  @alanssitis, would you mind splitting up this PR a bit. For instance, start with just `_resolver.py` an `_file_resolver.py` and make sure all tests pass? Once we have that, we can add other resolvers one at a time.

pyproject.toml
alanssitis: I agree, will remove for now.

in_toto/resolver/_resolver.py
alanssitis: Yes, I placed it in the class so other `Resolver` instances could utilize it. However, should this be moved to a separate file like `utils.py` or made to be a standalone function?

in_toto/resolver/_resolver.py
alanssitis: `resolve_uri` is currently a NOOP in other resolvers since I haven't implemented the function properly. One future application, for example, is having it be able to use a wildcard character (`*`) to get multiple docker containers. It could be squashed into `hash_artifacts` but I thought it might be best to have them be separate.

As for ITE specification, I think we can get away with either one or the other. I went with the latter since that's what we had done earlier, but I could see a case be made for the former since it will ease the burden on the resolvers.

in_toto/resolver/_resolver.py
alanssitis: I was thinking later down the line, we will allow users to write their own resolvers and place them in a directory. They might have different requirements and parameters but that could be handled as ITE-4 matures.

alanssitis: Will remove `kwargs` from interface.

alanssitis:
Got tests back up and running, will update documentation later. Might need a better way to handle Windows paths.

in_toto/resolver/_resolver.py
```diff
@@ -0,0 +1,113 @@
...
+
+    for excluded in exclude_patterns.match_files(names):
+      included.discard(excluded)
+
+    return sorted(included)
+
+  @classmethod
+  @abstractmethod
+  def resolve_uri(cls, generic_uri, exclude_patterns=None,
+                  follow_symlink_dirs=False):
```
lukpueh: Isn't `follow_symlink_dirs` file resolver specific? I don't think this should be part of the abstract interface. Same goes for `normalize_line_endings` below.

lukpueh: `exclude_patterns` is probably fine in the abstract interface, because a list of strings should work for any URI, and it's up to the resolver how to interpret it.

in_toto/runlib.py
```diff
@@ -57,44 +57,26 @@
...
-      normalize_line_endings=False):
-  """Internal helper that takes a filename and hashes the respective file's
-  contents using the passed hash_algorithms and returns a hashdict conformant
-  with securesystemslib.formats.HASHDICT_SCHEMA. """
+def _hash_artifact(hashable_representation, hash_algorithms=None):
   if not hash_algorithms:
     hash_algorithms = ['sha256']
 
+  if not isinstance(hashable_representation, bytes):
+    hashable_representation = hashable_representation.encode('utf-8')
```
lukpueh: Won't you always call this helper with bytes?

in_toto/runlib.py
```diff
@@ -235,10 +212,20 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
+  resolved_artifacts = []
+  for artifact in artifacts:
+    resolved_artifacts.extend(Resolver.resolve_uri(
+      artifact,
+      exclude_patterns=exclude_patterns,
+      follow_symlink_dirs=follow_symlink_dirs))
+
+  if exclude_patterns:
+    resolved_artifacts = Resolver.apply_exclude_patterns(
+        resolved_artifacts, exclude_patterns)
```
lukpueh: The usage pattern of passing exclude patterns to `Resolver.resolve_uri` and calling `Resolver.apply_exclude_patterns`, if exclude patterns exist is not quite ideal. Shouldn't the former be able to handle both?

in_toto/runlib.py
```diff
@@ -235,10 +212,20 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
     securesystemslib.formats.NAMES_SCHEMA.check_match(exclude_patterns)
-    norm_artifacts = _apply_exclude_patterns(norm_artifacts, exclude_patterns)
+
+  # Normalize passed paths
+  resolved_artifacts = []
+  for artifact in artifacts:
+    resolved_artifacts.extend(Resolver.resolve_uri(
+      artifact,
+      exclude_patterns=exclude_patterns,
+      follow_symlink_dirs=follow_symlink_dirs))
```
lukpueh: As said above, it does not seem right to pass `follow_symlink_dirs` here, where we don't yet know the scheme. Only file resolvers will care for this option. Other URIs could need other things.

lukpueh: Same goes for the base path shenanigans we do further above and below, they only make sense for files.

in_toto/runlib.py
```diff
@@ -249,76 +236,11 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
-              normalize_line_endings=normalize_line_endings)
-
-    # Path is no file and no directory
-    else:
-      LOG.info("path: {} does not exist, skipping..".format(artifact))
-
+  for artifact in resolved_artifacts:
+    key = _apply_left_strip(artifact, artifacts_dict, lstrip_paths)
+    artifacts_dict[key] = _hash_artifact(
+        Resolver.get_hashable_representation(artifact, normalize_line_endings))
```
lukpueh: See comment above about `follow_symlink_dirs` and base path. Same goes for `normalize_line_endings`.

in_toto/resolver/_resolver.py
```diff
@@ -0,0 +1,113 @@
...
+    exclude_patterns = PathSpec.from_lines('gitwildmatch', exclude_patterns)
+
+    for excluded in exclude_patterns.match_files(names):
+      included.discard(excluded)
+
+    return sorted(included)
+
+  @classmethod
+  @abstractmethod
+  def resolve_uri(cls, generic_uri, exclude_patterns=None,
```
lukpueh: Minor naming nit: "resolve" can mean many things and we use it a bit inconsistently here. On one hand it is the name of the interface, whose purpose is to "resolve" a URI into a hash or hashable representation, but in this method it means "resolve" a URI into multiple URis.

lukpueh (changes requested):
Thanks for updating the PR, @alanssitis! think the abstraction boundaries still need some refining. Right now we can't really just use the resolver interface for arbitrary URIs, but need to know what URIs we are dealing with, in order to invoke it with the right set of parameters. That defeats the purpose of the pattern.

This is not necessarily a problem of the implementation only but also of the ITE. But I think we can *resolve* it. ;)

in_toto/runlib.py
alanssitis: In this scenario it is. However, I was thinking that perhaps, down the line, we may be resolving URIs that do not have a string representation e.g. tarballs and binaries.

in_toto/resolver/_resolver.py
alanssitis: I thought so too, but I haven't been able to come up with a good way to handle these parameters. Global variables feel inadequate in this scenario, so I'll keep brainstorming other solutions.

in_toto/runlib.py
alanssitis: It should!

in_toto/runlib.py
lukpueh: Sure, but shouldn't the resolver interface method make sure to aways return the same data type?

in_toto/resolver/_resolver.py
adityasaky:
```
resolver = Resolver.from_uri(uri)
```

I think this approach makes sense and is conducive to supporting custom resolvers in future, no?

in_toto/resolver/_resolver.py
adityasaky: I think it's fine to use the method Lukas described above when it comes to custom parameters.

in_toto/runlib.py
adityasaky: > Sure, but shouldn't the resolver interface method make sure to aways return the same data type?

+1, we should standardize this

alanssitis: should we then utilize type hints for this?

in_toto/resolver/_resolver.py
alanssitis: used a global parameter dictionary in `resolver`

in_toto/runlib.py
alanssitis: since there really arent type hints throughout the repo, am just going to place it in the comments

alanssitis:
One thing worth looking into, should we define a new behavior in ITE 4 on how to handle URI's that we don't have a resolver for instead of just erroring out as I am doing right now?

in_toto/runlib.py
```diff
@@ -249,76 +231,11 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
-              normalized_filepath, artifacts_dict, lstrip_paths)
-          artifacts_dict[key] = _hash_artifact(filepath,
-              normalize_line_endings=normalize_line_endings)
-
-    # Path is no file and no directory
-    else:
-      LOG.info("path: {} does not exist, skipping..".format(artifact))
-
+  for artifact in resolved_artifacts:
+    key = _apply_left_strip(artifact, artifacts_dict, lstrip_paths)
```
adityasaky: This should be updated to left strip while ignoring URI scheme if available.

in_toto/resolver/resolver.py
```diff
@@ -0,0 +1,113 @@
...
+
+  @classmethod
+  def apply_exclude_patterns(cls, names, exclude_patterns=None):
+    """Exclude matched patterns from passed names."""
+    if not exclude_patterns:
+      return names
+
+    included = set(names)
+
+    exclude_patterns = PathSpec.from_lines('gitwildmatch', exclude_patterns)
```
adityasaky: Do you foresee any changes to how exclude patterns are provided for non file examples? Do we assume the scheme is part of it?

adityasaky (commented):
I think this is looking good! I have some questions around how we expect general parameters such as exclude patterns and left stripping apply to the new artifact names.

in_toto/runlib.py
```diff
@@ -235,10 +208,19 @@ def record_artifacts_as_dict(artifacts, exclude_patterns=None,
...
     exclude_patterns = in_toto.settings.ARTIFACT_EXCLUDE_PATTERNS
 
-  # Apply exclude patterns on the passed artifact paths if available
   if exclude_patterns:
     securesystemslib.formats.NAMES_SCHEMA.check_match(exclude_patterns)
-    norm_artifacts = _apply_exclude_patterns(norm_artifacts, exclude_patterns)
+
+  # Set resolver configs
+  RESOLVER_PARAMS["follow_symlink_dirs"] = follow_symlink_dirs
+  RESOLVER_PARAMS["normalize_line_endings"] = normalize_line_endings
```
lukpueh: Am not a big fan of patching globals to configure behavior in methods -- I know, the way I use `in_toto.settings` around here suggests something else, but I learned from past mistakes, see e.g. related secure-systems-lab/securesystemslib#219.


Maybe we can instead configure the resolver directly? E.g.
```python
resolver = Resolver.from_uri(uri)

if isinstance(resolver, FileResolver):
  resolver.follow_symlink_dirs = follow_symlink_dirs
  resolver.normalize_line_endings = normalize_line_endings

resolver.resolve_uri_to_uris(uri) 
```

lukpueh: Speaking of FileResolver specific behavior, can't we also hide the base_path case handling inside the file resolver? It doesn't seem very useful to `chdir` if the URIs are not file paths.

lukpueh (changes requested):

in_toto/runlib.py
alanssitis: Do you want to remove the URI scheme before left-stripping?

in_toto/resolver/resolver.py
alanssitis: I think most cases can be covered by the adding non-file names to exclude patterns.

in_toto/runlib.py
alanssitis: I see, I have a decent idea on how I can go about with this.

in_toto/resolver/_resolver.py
adityasaky: I think we should expose both. We have scenarios where `get_hashable_representation` may not be worth the trouble. Consider recording a git or ostree object, for example. We could argue just having their respective hashes suffice. On the other hand, with less definite systems, we probably want to expose both as a means to enable users to examine the data directly when necessary.

in_toto/runlib.py
adityasaky: I think that makes sense. On the other hand, do we support lstripping paths based on its scheme? i.e., lstrip for file paths but not for scheme X which might also match the pattern. :thinking:

adityasaky: For now, let's set lstrip to apply to the parts after the scheme and we can revisit it later I think.

alanssitis: Let's keep URI-specific lstrips as a future issue.

in_toto/resolver/_resolver.py
alanssitis: I implemented a method that would allow this. Developers either implement `get_hashable_representation` or `hash_artifact`.

in_toto/runlib.py
alanssitis: I think `chdir` should stay since it might be used for non-file specific resolvers like git for example.

in_toto/runlib.py
adityasaky: If my thinking is right, we'd currently have to lstrip with the scheme included in the lstrip pattern. This would remove the scheme as well, right? I think we want to handle that at least in part now.

adityasaky:
I think I'm okay with going ahead with this and making changes in separate PRs aside from the lstrip comment. @lukpueh could you take a look as well?

in_toto/runlib.py
alanssitis: I have lstrip working on only the right part of the scheme (keeps scheme intact). What I meant by URI-specific is the lstripping of paths based on scheme.

adityasaky: Could you give me an example?

alanssitis: If you look at the tests `TestApplyLeftStrip` it does it.

If you are asking for examples to URI-specific lstripping, I can foresee scenarios where an artifact's URI has the same/very-similar URI to a file path in a user's `lstrip_paths` but they do not want it to apply to the aforementioned artifact.

in_toto/runlib.py
lukpueh: I'd argue that, if you need to chdir to resolve artifacts, then you are probably using a sort of file resolver, even if you are looking for very specific files like git objects.

in_toto/resolver/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+)
+
+RESOLVER_FOR_URI_SCHEME.update(
+  {
+    DEFAULT_SCHEME: FileResolver,
+    FileResolver.SCHEME: FileResolver,
+  }
+)
+
+def set_params(**kwargs):
```
lukpueh: May I ask, why you favor this approach over [my suggestion](https://github.com/in-toto/in-toto/pull/536#discussion_r1137239643)? 

`set_params` hides from the caller, which resolver they are configuring, but I don't see the advantage over being explicit.

alanssitis: Was discussed in a meeting earlier today.

I was trying to solve a problem that didn't exist and it wouldn't have worked. Will have the configs be set explicitly.

in_toto/runlib.py
alanssitis: Left strip will be handled by the resolver.

in_toto/resolver/_resolver.py
alanssitis: I have consolidated both into a singular api that asks for a `hash_dict`.

lukpueh:
Thanks for the updates! I still wonder why we need all this new API. The `Resolver` interface exposes 7 new methods, plus another public function in the module, but  `runlib.record_artifacts_as_dict` is still 60 lines long and tightly coupled with the FileResolver class. I think the goal for the runlib function is to look something like this:

```python
def record_artifacts_as_dict(uris, **kwargs): # renamed args/kwargs for emphasis
    hashed_artifacts = {}
    for uri in uris:
        # 1. Get resolver for scheme
        resolver = Resolver.for_uri(uri)

        # 2. Optional: case-handle **kwargs by resolver type

        # 3. Get a dictionary of hashed artifacts for the uri
        # Implementation details: store uri, resolve uri to uris, mangle URIs (lstrip, base path, etc.), apply exclude patterns
        hashed_artifacts.update(resolver.hash_artifacts(**kwargs))

    return hashed_artifacts
```

I suggest that we start with an absolute minimum we need to make this work. It's easier to add missing API later than to maintain or remove existing obsolete API.

Regarding performance, I honestly doubt that it is a problem to instantiate multiple resolver instances of the same type, same goes for running chdir multiple times. 

*EDIT: I think we should ignore those performance concerns for the time being and deal with them if they become a problem. Or actually check if those concerns are valid, before we let them inform our decision.*

alanssitis:
@lukpueh In that case, I will simplify the `Resolver` interface and as other resolvers are added we can figure out what needs to stay and what needs to be moved out.

adityasaky:
Closing in favour of #578. Thanks for your work on this @alanssitis!

Issue #534, opened 2023-01-17
Add custom Key, Signer and Signature implementations for in-toto flavored GPG


lukpueh:
securesystemslib [v0.26.0](https://github.com/secure-systems-lab/securesystemslib/releases/tag/v0.26.0) released an improved signer interface for more signing flexibility. Currently, there are several efforts to adopt usage of this interface in in-toto:
- [#503](https://github.com/in-toto/in-toto/pull/503) uses `Signer` and `Key` objects internally
- [#532](https://github.com/in-toto/in-toto/pull/532), [#533](https://github.com/in-toto/in-toto/pull/533) request to support them in-toto CLI/API

securesystemslib also provides a `GPGSigner` and `GPGKey` implementation, however, they use updated metadata formats, fixing the in-toto/tuf specification compatibility issues described in https://github.com/secure-systems-lab/securesystemslib/issues/450 and https://github.com/secure-systems-lab/securesystemslib/issues/448.

To allow in-toto to use the new signer API, but continue support for existing gpg signature and key metadata formats, we can provide custom `Key`, `Signer` and `Signature` implementations, e.g.  `_LegacyGPGKey`, `_LegacyGPGSigner`, `_LegacyGPGSignature` here in in_toto.

These implementation would align mostly with `GPGSigner`, `GPGSignature` and `GPGKey`, as implemented in securesystemslib prior to https://github.com/secure-systems-lab/securesystemslib/pull/486/ and https://github.com/secure-systems-lab/securesystemslib/pull/488.

lukpueh:
Closed by #538

PR #538, opened 2023-01-27
Adds in-toto flavored GPG Key, Signer, and Signature

PradyumnaKrishna:
**Fixes issue #534**: Add custom Key, Signer and Signature implementations for in-toto flavored GPG

Adds in-toto flavored legacy GPGKey, GPGSigner, and GPGSignature to be used with other Signer implementation added with the DSSE.

DSSE won't support this legacy GPG api but it is useful in supporting old signature wrapper with gpg compatibilities, i.e. `Metablock` to generate in-toto Metadata and signing with GPG.

This newer `GPGSigner` and `GPGKey` implementation provided by securesystemslib utilizes updated metadata formats, that is incompatible with current in-toto signature wrappers.

Hence, these `_LegacyGPGSigner` and `_LegacyGPGKey` implementation is provided to continue old wrapper support without any breaking change.

Signed-off-by: Pradyumna Krishna <git@onpy.in>

requirements**:

### DIFF
--- a/in_toto/models/gpg.py
+++ b/in_toto/models/gpg.py
@@ -1,0 +1,241 @@
+# Copyright New York University and the in-toto contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+<Program Name>
+  gpg.py
+
+<Author>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
+
+<Started>
+  Jan 26, 2023
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Provides in-toto flavored GPGSigner, GPGSignature and GPGKey.
+
+"""
+
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional
+
+import securesystemslib.gpg.functions as gpg
+from securesystemslib.signer import Key, Signature, Signer, SecretsHandler
+
+
+class _LegacyGPGSignature(Signature):
+  """A container class containing information about a gpg signature.
+  Besides the signature, it also contains other meta information
+  needed to uniquely identify the key used to generate the signature.
+
+  Attributes:
+    keyid: HEX string used as a unique identifier of the key.
+    signature: HEX string representing the signature.
+    other_headers: HEX representation of additional GPG headers.
+  """
+
+  def __init__(
+    self,
+    keyid: str,
+    signature: str,
+    other_headers: str,
+  ):
+    super().__init__(keyid, signature)
+    self.other_headers = other_headers
+
+  @classmethod
+  def from_dict(cls, signature_dict: Dict) -> "_LegacyGPGSignature":
+    """Creates a ``_LegacyGPGSignature`` object from its JSON/dict
+    representation.
+
+    Arguments:
+      signature_dict: Dict containing valid "keyid", "signature" and
+        "other_fields" fields.
+    Raises:
+      KeyError: If any of the "keyid", "sig" or "other_headers" fields
+        are missing from the signature_dict.
+    Returns:
+      ``_LegacyGPGSignature`` instance.
+    """
+
+    return cls(
+      signature_dict["keyid"],
+      signature_dict["signature"],
+      signature_dict["other_headers"],
+    )
+
+  def to_dict(self) -> Dict:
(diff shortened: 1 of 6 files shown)


in_toto/models/gpg.py
```diff
@@ -0,0 +1,250 @@
+# Copyright New York University and the in-toto contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+<Program Name>
+  gpg.py
+
+<Author>
+  Lukas Puehringer <lukas.puehringer@nyu.edu>
```
lukpueh: I'm flattered, but I think you should be the author. :)

in_toto/models/gpg.py
```diff
@@ -0,0 +1,250 @@
...
+      key_dict["hashes"],
+      key_dict["keyval"],
+      keyid,
+      key_dict.get("creation_time"),
+      key_dict.get("validity_period"),
+      gpg_subkeys,
+    )
+
+  @classmethod
+  def from_legacy_dict(cls, key_dict: Dict[str, Any]):
```
lukpueh: I don't think this class needs separate `from_dict` and `from_legacy_dict`

in_toto/models/gpg.py
```diff
@@ -0,0 +1,250 @@
...
+    instance.
+
+    Arguments:
+      signature: A ``_LegacyGPGSignature`` class instance.
+      payload: The bytes to be verified.
+    Returns:
+      Boolean. True if the signature is valid, False otherwise.
+    """
+
+    return gpg.verify_signature(signature.to_dict(), self.to_dict(), data)
```
lukpueh: Can you take a look at `verify_signature` functions in securesystemslib? They should raise or return None (see e.g. https://github.com/secure-systems-lab/securesystemslib/pull/488)

lukpueh (changes requested):
Thanks for the PR, @PradyumnaKrishna! Here's a first batch of comments. I'll leave more tomorrow.

in_toto/models/gpg.py
PradyumnaKrishna: `from_dict` is a derived method and pylint throws `arguments-differ` if I remove `keyid` argument from it.

lukpueh: Okay, makes sense. `_LegacyGPGKey.from_legacy_dict` will be the pendant to `SSlibKey.from_securesystemslib_key`.

in_toto/models/gpg.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+  def sign(self, payload: bytes) -> _LegacyGPGSignature:
+    """Signs a given payload by the key assigned to the ``_LegacyGPGSigner``
+    instance. Calls the gpg command line utility to sign the passed content
+    with the key identified by the passed keyid from the gpg keyring at the
+    passed homedir.
+
+    The executed base command is defined in
+    securesystemslib.gpg.constants.GPG_SIGN_COMMAND.
```
lukpueh:
```suggestion
```
This constant has changed. I think we can just remove the comment.

tests/models/test_gpg.py
```diff
@@ -0,0 +1,163 @@
...
+    # Create a signature.
+    signer = _LegacyGPGSigner(self.signing_subkey_keyid, self.gnupg_home)
+    signature = signer.sign(self.test_data)
+
+    # Verify signature using expired key.
+    key = _LegacyGPGKey.from_keyring(self.expired_keyid, self.gnupg_home)
+    with self.assertRaises(VerificationError):
+      key.verify_signature(signature, self.test_data)
+
+  def test_gpg_signer_serialization(self):
```
lukpueh: Minor nit: This should be 
```suggestion
  def test_gpg_signature_serialization(self):
```
Right?

lukpueh (changes requested):
This should work. Thanks, @PradyumnaKrishna!!  Let's merge and see how it plays with your dsse PR. 

Though, maybe we can first make the module non-public? What do you think about calling it `_signer`? IMO then it would even be fine to just call the classes GPGSignature, GPGSigner and GPGKey.

Please also address my comments inline and this is good to go.

lukpueh (approved):
Cheers! 🎉

PR #544, opened 2023-02-09
Stop using securesystemslib.process

lukpueh:
* Instead of `securesystemslib.process.run`, use `subprocess.run` directly. The former is only a thin wrapper over the latter, which provides no advantage in Py3-only times.

* Copy `securesystemslib.process.run_duplicate_streams` to a helper in runlib. This function is really only needed internally by in-toto-run tooling, for which it was developped originally ([#160](https://github.com/in-toto/in-toto/pull/160)). It does not have to be public API in securesystemslib. Moreover, the function is not trivial and there have been multiple related bugs, one of them still unresolved (see [sslib#337](https://github.com/secure-systems-lab/securesystemslib/issues/337)).

This patch allows us to deprecate `securesystemslib.process`, and fix or workaround related issue wrt the specific runlib use case (very simple workaround: just disable the feature on windows)

**Note to git historians:**
run_duplicate_streams was removed from in-toto in in-toto/in-toto@29f063a, and now copied back from [`securesystemslib/process.py@v0.26.0`](https://github.com/secure-systems-lab/securesystemslib/blob/v0.26.0/securesystemslib/process.py)

### DIFF
--- a/in_toto/runlib.py
+++ b/in_toto/runlib.py
@@ -31,6 +31,11 @@ import glob
 import logging
 import os
 import itertools
+import io
+import subprocess  # nosec
+import sys
+import tempfile
+import time
 
 from pathspec import PathSpec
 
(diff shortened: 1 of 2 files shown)


lukpueh:
https://github.com/secure-systems-lab/securesystemslib/issues/337 becomes an in-toto issue, once this PR is merged. Let's transfer it to this repo then.

adityasaky (approved):

PR #545, opened 2023-02-14
Changing expiration check to match TUF

ShubhikaBhardwaj:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue**: #417 

**Description of the changes being introduced by the pull request**:
Changing `expire_datetime <= datetime.datetime.now(tz.tzutc())`
to `expire_datetime <= datetime.datetime.now(tz.tzutc())` as discussed at #417 
**Please verify and check that the pull request fulfills the following
requirements**:

- ☑ The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- ☑ Tests have been added for the bug fix or new feature
- ☑ Docs have been added for the bug fix or new feature

PR #546, opened 2023-02-14
Making metadata print pretty

ShubhikaBhardwaj:
that is provided, the better.

**Fixes issue #18 **:

- Showing metadata in a logically structured and human friendly way as well.
- Order fields logically instead of alphabetically
- Strip keys and signatures in all the modules present

requirements**:

- [☑] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [☑] Tests have been added for the bug fix or new feature
- [☑] Docs have been added for the bug fix or new feature

### DIFF
--- a/in_toto/models/metadata.py
+++ b/in_toto/models/metadata.py
@@ -60,6 +60,13 @@ class Metablock(ValidationMixin):
     self.signed = kwargs.get("signed")
     self.compact_json = kwargs.get("compact_json", False)
 
+    if self.signatures is not None:
+      self.signatures=self.signatures.strip()
+    if self.signed is not None:
+      self.signed = self.signed.strip()
+    if self.compact_json is not None:
+      self.compact_json = self.compact_json.strip()
+
     self.validate()
 
 
@@ -75,7 +82,7 @@ class Metablock(ValidationMixin):
         },
         indent=indent,
         separators=separators,
-        sort_keys=True
+        sort_keys=False
       )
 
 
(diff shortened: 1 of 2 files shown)


lukpueh:
Thanks for the patch, @ShubhikaBhardwaj! Is this PR ready for review? If not, please state so somewhere in the PR description or in a comment. You can also [mark the PR as a draft ](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests).

lukpueh:
Hi @ShubhikaBhardwaj, do you intend to continue your work here? Otherwise, I suggest to close the PR.

ShubhikaBhardwaj:
Hi @lukpueh  
This PR is ready for review. Waiting for your feedback

adityasaky:
I took a quick look at this PR. I think there's been a slight misunderstanding of the original issue but it shouldn't be too hard to fix. Your patch, as it's written here, tries to remove signatures and sort metadata keys in the actual metadata itself. This is not the intent of the original ticket, because we actually need that information to _verify_ the metadata. Instead, what the PR should do is present a separate interface for the user to load some in-toto metadata and gain a higher level, human friendly representation of the information in that metadata.

You'll notice that if you update the PR to fix the use of `strip` where it can't be used, tests will still fail because at the end of the day we (and so our tests) do care about the signatures and other key information, just perhaps not when we want to examine the metadata for higher level reasoning. Moving some of the same changes into a separate interface is probably a good start!

Finally, I think your commit from #545 snuck in here as well, but that's a minor nit. :smile:

lukpueh:
Closing due to inactivity...

PR #567, opened 2023-03-22
Update security policy to use new reporting

### LINKED ISSUE #566: No User ID for Vulnerability Report PGP Key in SECURITY.md
*[based on X41 source code audit informational note]*

The SECURITY.md file specifies:

> Optionally, reports that are emailed can be encrypted with PGP. You should use PGP key fingerprint 903B AB73 640E B6D6 5533 EFF3 468F 122C E816 2295.

This PGP key is hard to use because it contains no user ID:
```
$ gpg --recv-keys 903BAB73640EB6D65533EFF3468F122CE8162295
gpg: key 468F122CE8162295: new key but contains no user ID - skipped

$ gpg --encrypt --recipient 903BAB73640EB6D65533EFF3468F122CE8162295
gpg: 903BAB73640EB6D65533EFF3468F122CE8162295: skipped: No public key
gpg: [stdin]: encryption failed: No public key
```

**Solution Advice**
If the key has a user ID, it is recommended to include the public key instead of the fingerprint. This avoids needing a third party to obtain the key as well as avoiding the stripped user ID problem.

If the key has no user ID in its original form, the easiest solution might be to give it a generic name without email address such as "in-toto Security Key".

lukpueh:
- Enabled new GitHub feature (beta) via GitHub Web UI to privately report security issues to all maintainers.
- Updated security policy document to instruct reporters to use the new reporting mechanism instead of email+pgp.

closes #566

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,14 +1,12 @@
-# Security Policy
+# Security Issues and Bugs
 
-Security issues can be reported by emailing torresariass@gmail.com.
+Security issues can be reported to maintainers [privately via
+GitHub](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability):
 
-At a minimum, the report must contain the following:
+- [**Report new
+  vulnerability**](https://github.com/in-toto/in-toto/security/advisories/new)
 
-- Description of the vulnerability.
-- Steps to reproduce the issue.
-
-Optionally, reports that are emailed can be encrypted with PGP. You should use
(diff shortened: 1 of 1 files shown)


adityasaky:
Should we recommend github security reports as a primary mechanism while allowing users to send an encrypted email as a backup? You can't report a vulnerability without a github account otherwise.

lukpueh:
> Should we recommend github security reports as a primary mechanism while allowing users to send an encrypted email as a backup? 

Good point. Done!

adityasaky (approved):

Issue #562, opened 2023-03-17
Add support for ITE-6 Attestations


adityasaky:
Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.

**Description of issue or feature request**:

in-toto-python does not currently support generating [ITE-6 attestations](https://github.com/in-toto/attestation). The task can be broken down into the following steps:
1. Generating in-toto Statements. See: https://github.com/in-toto/attestation/blob/main/spec/v1.0-draft/statement.md for more information.
2. Adding support for specific predicates such as SLSA Provenance and one or more SBOM specs.

The generated metadata can leverage the newly added support for DSSE (#503).

Related: #464, #465, #466

**Current behavior**:

in-toto-python has no support for generating ITE-6 attestations.

**Expected behavior**:

in-toto-python can generate ITE-6 attestations with support for some predicate types.

chasen-bettinger:
@adityasaky Is this ticket ok to pick up?

adityasaky:
Hi @chasen-bettinger! We've currently listed it for GSoC. Do you plan to apply through GSoC or pick it up directly? (Either is fine)

chasen-bettinger:
@adityasaky I presume GSoC is 'Google Summer of Code'? I don't really know much about that program so probably picking it up directly.

adityasaky:
Gotcha! Feel free to jump on the CNCF slack (#in-toto, #in-toto-python) if you have any questions.

PradyumnaKrishna:
in-toto doesn't have any methods to generate `Provenance` predicates, or other predicate types. Does generating such metadata with in-toto and verifying them comes in this project scope?

Also, FYI I have submitted my first draft of proposal on summer of code portal. Please provide me a review.

mmcniece:
👋 hey @chasen-bettinger & @adityasaky -- I'm looking to use the in-toto lib to generate provenance attestations.  Is there a place I can jump in to help get this across the line and merged?

adityasaky:
Have you taken a look at https://github.com/in-toto/in-toto/pull/570? Also cc @marcelamelara

PR #466, opened 2021-08-03
[W.I.P] Add new model: Provenance

Dentrax:
Fixes #464

Signed-off-by: Furkan <furkan.turkal@trendyol.com>

For the `TODO` line:
```python
# TODO: securesystemslib.formats.ANY_NUMBER_SCHEMA.check_match(self.recipe.type)
```

I already created an issue: https://github.com/secure-systems-lab/securesystemslib/issues/367

I created a small function called `_get_attribute` in order to pass `(OPTIONAL)` variables. (I followed the [spec](https://github.com/in-toto/attestation/blob/main/spec/predicates/provenance.md))

I implemented a bare model and tests, but couldn't figure it out how to make it work. It seems most `in_toto_*` entry points uses the `link` type for the running, generating and validating. Shouldn't we make that `run`, `sign`, `verify`, etc. more generic in order to parse and read different predicate types? Should I create a new for parser for this? What is the way out? Thanks! (I just wanted to learn project's internal domain :) )

Feed free to edit! Waiting your feedback!

---

**Description of the changes being introduced by the pull request**:

See the issue.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [X] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

PR #570, opened 2023-03-25
[WIP] Add support for Statements

chasen-bettinger:
**Fixes issue #**:
#562 

- Added the Statement model
- Incorporated the Statement model into the Envelope.
- Skipped a test because I'm awaiting feedback on whether this is the right direction to take this issue (#562). The test is built on the previously-used 'link' model as the payload.

Some notes:
- This is largely a WIP - general feedback on data structures and architecture most appreciated!
- I studied the following information throughout the process of writing this code:
  - https://slsa.dev/attestation-model
  - https://github.com/in-toto/attestation/issues/155
  - https://github.com/in-toto/attestation/pull/151/files
  - https://github.com/in-toto/attestation/blob/main/spec/v1.0/statement.md
  - https://github.com/in-toto/in-toto/pull/503
  - https://github.com/in-toto/attestation/tree/main/spec/predicates
  - https://github.com/secure-systems-lab/dsse
  - https://github.com/in-toto/in-toto/pull/466
- I will add docs once the code is in a good place. 

Some things I still don't understand:
- What are valid `predicate_types` and where do they come from?
- What is the expected output for the software after this change? I went through the demo and saw that `.link` files were being generated. I presumed based on the information on the SLSA website that `Links` are a subset of `Statements` and that once in-toto supported `Statements` (and `Envelopes` WITH `Signatures` and `Statements`), the output would be an `Envelope`? I don't really understand the bigger picture just yet. 
- I don't know what the values for a `predicate` hash should be nor where that comes from. [This resource](https://github.com/in-toto/attestation/pull/151/files) was most helpful in giving me breadcrumbs, but I'm still uncertain.

requirements**:


in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+  Mar 25, 2023
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Provides a class that represents a Statement, the middle layer of an 
+  attestation, binding it to a particular subject and unambiguously 
+  identifying the types of the Predicate as defined in the in-toto 
+  attestion spec.
```
adityasaky:
```suggestion
  attestation spec.
```

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+UNFINISHED_FILENAME_FORMAT = ".{step_name}.{keyid:.8}.link-unfinished"
+UNFINISHED_FILENAME_FORMAT_GLOB = ".{step_name}.{pattern}.link-unfinished"
+
+
+@attr.s(repr=False, init=False)
+class Statement(Signable):
+  """The Statement is the middle layer of the attestation, binding it to a 
+  particular subject and unambiguously identifying the types of the Predicate.
+
+  A Link object is usually contained in a generic Metablock object for signing,
```
adityasaky: I think we can do away with this and perhaps point to https://github.com/in-toto/attestation/blob/main/spec/v1.0/statement.md instead, especially as the descriptions below are from the spec.

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+  predicate = attr.ib()
+  MODEL_NAME="Statement"
+
+
+  def __init__(self, **kwargs):
+    super(Statement, self).__init__()
+
+    self._type = STATEMENT_TYPE
+    self.subject = kwargs.get("subject")
+    self.predicate_type = kwargs.get("predicate_type")
```
adityasaky: The spec uses camelcasing rather than snake casing. We'd probably be reading in `predicateType` rather than `predicate_type`. Is there any automatic conversion I'm unaware of?

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+  identifying the types of the Predicate as defined in the in-toto 
+  attestion spec.
+"""
+
+import attr
+import securesystemslib.formats
+from in_toto.models.common import Signable
+
+STATEMENT_TYPE = "https://in-toto.io/Statement/v1"
+FILENAME_FORMAT = "{step_name}.{keyid:.8}.link"
```
adityasaky: I think these don't belong at the statement layer. We'll have to handle them smartly for the link predicate only.

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+  def _validate_predicate_type(self):
+    """Private method to check that `predicate_type` is a `str`."""
+    if not isinstance(self.predicate_type, str):
+      raise securesystemslib.exceptions.FormatError(
+          "Invalid {}: field `predicate_type` must be of type str, got: {}"
+          .format(self.MODEL_NAME, type(self.predicate_type)))
+
+  def _validate_predicate(self):
+    """Private method to check that `predicate` is a `dict`."""
+    if not isinstance(self.predicate, dict):
```
adityasaky: Perhaps we can skip this check for now. Once we define link predicate, slsa provenance etc, we'd want `self.predicate` to be an instance of one of those models rather than a dict.

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+  def _validate_subject(self):
+    """Private method to check that `subject` is a `list` of `HASHDICTs`."""
+    if not isinstance(self.subject, list):
+      raise securesystemslib.exceptions.FormatError(
+          "Invalid {}: field `subject` must be of type dict, got: {}"
+          .format(self.MODEL_NAME, type(self.subject)))
+
+    for hash in self.subject:
```
adityasaky:
```suggestion
    for s in self.subject:
```

IMO hash isn't quite right. Just an iterator or `entry` or something may be more appropriate.

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+        for subject_key in hash:
```
adityasaky: I think we don't need this loop per se. The check can be simplified to:

```
if "name" in hash and not isinstance(hash["name"], str):
    raise...
```

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+            subject_value = hash[subject_key]
+            if subject_key == "name":
+                if not isinstance(subject_value, str):
+                    raise securesystemslib.exceptions.FormatError(
+                    "Invalid {}: field `name` must be of type str, got: {}"
+                    .format(self.MODEL_NAME, type(subject_value)))
+
+                continue
+
+            securesystemslib.formats.HASHDICT_SCHEMA.check_match(subject_value)
```
adityasaky: With the above change this becomes:

```suggestion
            securesystemslib.formats.HASHDICT_SCHEMA.check_match(hash["digest"])
```

adityasaky: FWIW, this will evolve to support resource descriptors and such anyway where we'd have to validate one of three fields is necessarily present and so on...

in_toto/runlib.py
```diff
@@ -665,9 +666,29 @@ def in_toto_run(name, material_list, product_list, link_cmd_args,
       materials=materials_dict, products=products_dict, command=link_cmd_args,
       byproducts=byproducts, environment=environment)
 
+  LOG.info("Creating statement...")
```
adityasaky: IMO we don't touch runlib at all in this PR. Let this one just add the Statement layer. A future PR can then add support for link predicates (https://github.com/in-toto/attestation/blob/main/spec/predicates/link.md) and add a flag to runlib to generate links using the Statement layer rather than the current format. We've also got to be mindful of backwards compatibility with runlib and verifylib and this change is certainly a breaking change.

in_toto/models/statement.py
```diff
@@ -0,0 +1,153 @@
...
+      raise securesystemslib.exceptions.FormatError(
+          "Invalid {}: field `predicate_type` must be of type str, got: {}"
+          .format(self.MODEL_NAME, type(self.predicate_type)))
+
+  def _validate_predicate(self):
+    """Private method to check that `predicate` is a `dict`."""
+    if not isinstance(self.predicate, dict):
+      raise securesystemslib.exceptions.FormatError(
+          "Invalid {}: field `predicate` must be of type dict, got: {}"
+          .format(self.MODEL_NAME, type(self.predicate)))
```
adityasaky: lint: single new line at the end of a file

tests/models/test_statement.py
```diff
@@ -0,0 +1,146 @@
...
+    test = copy.deepcopy(_get_statement())
+    test.validate()
+
+    # `predicate_type` property is not a string
+    test.predicate_type = False
+    with self.assertRaises(FormatError):
+      test.validate()
+
+if __name__ == "__main__":
+  unittest.main()
```
adityasaky: lint: single new line at the end of a file

in_toto/models/statement.py
adityasaky: Also, indent levels are inconsistent here with the rest of the file.

tests/models/test_statement.py
```diff
@@ -0,0 +1,146 @@
...
+_correct_predicate_type = "http://in-toto.io/attestation/human-review/vcs/v0.1"
+
+_correct_predicate = {
+    "result": "approved",
+    "reviewLink": "https://github.com/in-toto/in-toto/pull/503#pullrequestreview-1341209941",
+    "timestamp": "2023-03-15T11:05:00Z",
+    "reviewer": "https://github.com/lukpueh"
+}
+
+def _get_statement():
```
adityasaky: Can this be moved into a `setUp` method?

tests/models/test_statement.py
```diff
@@ -0,0 +1,146 @@
...
+      test.validate()
+
+    # `digest` property is not a dict
+    incorrect_subject = copy.deepcopy(_correct_subject)
+    incorrect_subject[0]["digest"] = False
+    test.subject = incorrect_subject
+    with self.assertRaises(FormatError):
+      test.validate()
+
+  def test_validate_predicate_type(self):
```
adityasaky: We should punt this until we know what predicate types we support in the implementation.

tests/models/test_statement.py
```diff
@@ -0,0 +1,146 @@
...
+    with self.assertRaises(FormatError):
+      test.validate()
+
+    # `predicate_type` property is not a string
+    test.predicate_type = False
+    with self.assertRaises(FormatError):
+      test.validate()
+
+  def test_validate_predicate(self):
+    """Test `predicate` field. Must be a `dict` if supplied"""
```
adityasaky: Same here, and we definitely don't want to use the vcs type for testing as that's still WIP.

adityasaky (changes requested):
Great start, @chasen-bettinger! I've left some comments but I think this is looking quite good. I'd like to scope this PR to adding support for the statement layer and tackle predicate implementations separately.

Edit: I also recommend switching to a feature branch instead of using `develop`.

marcelamelara:
Hello! I just came across this issue, thanks @chasen-bettinger . Have you seen the [protobuf definition](https://github.com/in-toto/attestation/blob/main/spec/v1.0/statement.proto) for statement? Instead of implementing a separate Python class for Statement, our recommendation (from the attestation side) would be to build applications based on the protos. This way, all ITE-6 compliant in-toto implementations actually use that same pre-defined format. If you're interested, we would very much welcome a contribution related to using the protos in the Python implementation.

adityasaky:
@chasen-bettinger this should unblock you: https://pypi.org/project/in-toto-attestation/

in_toto/models/link.py
```diff
@@ -85,22 +91,58 @@ class Link(Signable):
...
+            self.materials = []
 
+        if hasattr(self, 'command') == False:
+            self.command = []
+
+        if hasattr(self, 'environment') == False:
+            self.environment = {}
+
+        if hasattr(self, 'byproducts') == True:
+            self.byproducts["return-value"] = int(
```
chasen-bettinger: @adityasaky In order to successfully serialize this class into a canonical JSON, I've had to write in this hack. When the class is parsed from a protobuf to a JSON, the value becomes a float, which breaks downstream processes because that is [against the rules of canonical JSON](https://github.com/secure-systems-lab/securesystemslib/blob/main/securesystemslib/formats.py#L684). In the current design, we cannot get around this limitation because it is a central part of the `Signable` class and its `signable_bytes` property. 

As you may have already witnessed, either due to the architecture or my poor Python skills (most likely the latter), this class (and the Statement class) became quite ugly when trying to integrate the protobufs into the existing structure. As the code is written now, I believe there's not a way around using the Signable class as it is integral in generating and validating signatures. I fear a slight refactor might be in the near future, but I wanted your thoughts first.

Any thoughts on the best way to proceed?

cc: @marcelamelara

in_toto/models/link.py
```diff
@@ -85,22 +91,58 @@ class Link(Signable):
...
 
-        self._type = "link"
-        self.name = kwargs.get("name")
-        self.materials = kwargs.get("materials", {})
-        self.products = kwargs.get("products", {})
-        self.byproducts = kwargs.get("byproducts", {})
-        self.command = kwargs.get("command", [])
-        self.environment = kwargs.get("environment", {})
+        link = linkpb.Link()
+        link.name = kwargs.get("name", "")
```
adityasaky: This is a breaking change, we should support the use of link predicates alongside legacy links for now.

in_toto/models/link.py
adityasaky: If I'm not mistaken, the link predicate shouldn't be `Signable` at all. The predicate will be embedded in a Statement which is what we'd actually be signing.

.gitignore
```diff
@@ -61,3 +61,6 @@ target/
 #Ipython Notebook
 .ipynb_checkpoints
 .travis-solo/
+
+# IDE
+.vscode/
```
adityasaky: Can this go in your global .gitignore?

adityasaky (commented):
I have more comments about how we'd architect this that I'll submit separately.

in_toto/models/link.py
marcelamelara: I missed that the link was extending `Signable`. Agree with Aditya that shouldn't be the case. The link proto should be converted into a struct, inserted into the Statement `predicate` field, and the Statement will then be signed. 

I also believe that Statement might not have to extend `Signable` anymore either. The validation of the structure is now implemented in the Statement wrapper class provided by the `in_toto_attestation` package, and we can similarly use protobuf APIs directly to expose the signable bytes for a Statement.

in_toto/models/statement.py
```diff
@@ -0,0 +1,174 @@
...
+
+import in_toto_attestation.v1.statement_pb2 as statementpb
+import in_toto_attestation.v1.resource_descriptor_pb2 as rdpb
+from google.protobuf.json_format import MessageToDict
+from in_toto.models.common import Signable
+
+STATEMENT_TYPE = "https://in-toto.io/Statement/v1"
+
+
+# NOTE: 08/24/23 - This function is an intentional
```
chasen-bettinger, line 36: @adityasaky @marcelamelara Review please

tests/test_in_toto_run.py
chasen-bettinger, line 1: I'll correct these tests once the structure of the implementation details have settled. Was using this as my test bed for validating changes.

in_toto/models/statement.py
lukpueh: I fear there is not much else you can do here, if you want Statement to be a Signable.

FYI: I had a very similar issue in in-toto-golang a while ago and tried the same fix you proposed in https://github.com/secure-systems-lab/securesystemslib/issues/625 but did not use it for the reasons pointed out by @jku. The fix I did end up using there ([case-handling `json.Number`](https://github.com/in-toto/in-toto-golang/commit/814100d385dabf507399ff183eba16e3071db142)) won't work in Python. But maybe that's okay, because the fix is actually quite a hack.

Because of all these problems with canonical json, I think the best way going forward would be to just not support a canonical json representation for Statements, and use it with DSSE Envelopes only.

lukpueh: Seems like others have said this before: https://github.com/in-toto/in-toto/pull/570#discussion_r1213220358

chasen-bettinger: Yea, understood. I'm curious to learn the ramifications of adding a new method to the Envelope class within this library that doesn't require a `Signable` instance to instantiate.

.gitignore
```diff
@@ -60,4 +60,4 @@ target/
 
 #Ipython Notebook
 .ipynb_checkpoints
-.travis-solo/
+.travis-solo/
```
adityasaky, line 63: Can you add this new line back in?

in_toto/common_args.py
```diff
@@ -196,6 +196,13 @@ def parse_password_and_prompt_args(args):
     "help": ("generate metadata using dsse (experimental)."),
 }
 
+STATEMENT_ARGS = ["--use-statement"]
```
adityasaky, line 199: I'm worried `statement` is ambiguous, could you update this to be `attestation` instead?

in_toto/runlib.py
```diff
@@ -588,7 +593,33 @@ def in_toto_run(
 
     if use_dsse:
         LOG.info("Generating link metadata using DSSE...")
-        link_metadata = Envelope.from_signable(link)
+
+        if use_statement == False:
```
adityasaky, line 597: Potential footgun: if I set `use_statement` (or `use_attestation`) to True, I shouldn't be expected to set `use_dsse` to True as well. I see the CLI handles this but this is an issue with direct API use. I suggest we move `use_attestation` out of the `use_dsse` block and only support DSSE there. The `use_dsse` conditional block should only be for legacy links, and something we phase out in a future version as we deprecate legacy links in favour of link predicates.

adityasaky (changes requested):

in_toto/common_args.py
adityasaky: Ideally, `--generate-attestation`.

in_toto/models/metadata.py
```diff
@@ -223,7 +226,8 @@ def __repr__(self):
         separators = (",", ":") if self.compact_json else (",", ": ")
 
         return json.dumps(
-            {"signatures": self.signatures, "signed": attr.asdict(self.signed)},
+            {"signatures": self.signatures,
```
adityasaky, line 229: editor shenanigans?

in_toto/models/metadata.py
```diff
@@ -438,7 +442,8 @@ def _validate_signatures(self):
             )
 
         for signature in self.signatures:
-            securesystemslib.formats.ANY_SIGNATURE_SCHEMA.check_match(signature)
+            securesystemslib.formats.ANY_SIGNATURE_SCHEMA.check_match(
+                signature)
```
adityasaky, line 446: and here too?

PR #583, opened 2023-04-26
Use DSSE for inspection links

adityasaky:
that is provided, the better.

**Fixes issue #**: N/A

Provide the use-dsse option for inspection links.

requirements**:

### DIFF
--- a/in_toto/in_toto_verify.py
+++ b/in_toto/in_toto_verify.py
@@ -32,7 +32,7 @@ import logging
 from in_toto import verifylib
 from in_toto.common_args import (GPG_HOME_ARGS, GPG_HOME_KWARGS, VERBOSE_ARGS,
     VERBOSE_KWARGS, QUIET_ARGS, QUIET_KWARGS, title_case_action_groups,
-    sort_action_groups, OPTS_TITLE)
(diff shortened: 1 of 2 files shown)


lukpueh:
Inspection links aren't signed, why does the signature wrapper matter?

adityasaky:
Primarily for consistency. I noticed the difference on https://github.com/in-toto/demo/pull/49.

lukpueh:
> Primarily for consistency. I noticed the difference on https://github.com/in-toto/demo/pull/49.

Yeah, the inconsistency feels weird. But is it worth making the interface more clunky? What's the expected behaviour if `use_dsse` is true, but `persist_inspection_links` is false?

 Also, DSSE really does not seem useful, if you don't sign the payload. On the contrary, I need a DSSE library (base64 decoder) to make the payload even legible, which is not the case for the traditional signature wrapper.

I think there are other fixes for the inconsistency problem here, e.g.: we could use the same format as the passed layout, or we just dump the payload w/o signature wrapper, ... It all depends a bit on the expected use cases for those inspection links. I strongly suggest we defer this until we really need a fix, and just write a ticket in the meanwhile.

PR #589, opened 2023-05-04
Add tool to check local materials against products of passed link

lukpueh:
Adds new runlib functionary API, to perform basic vetting of materials. More specifically, to check if the **untrusted** materials match the products of a passed **trusted** link metadata file.

**Caveat**
The used link metadata is not validated against a layout. This means, no checking if the link is part of a certain supply chain, or if it was signed by a threshold of authorised functionaries.

--- 

This PR includes: API, CLI, RTD docs, tests and an small unrelated docs fix. Review commit by commit for details.

### DIFF
--- a/in_toto/in_toto_check_materials.py
+++ b/in_toto/in_toto_check_materials.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+
+# Copyright New York University and the in-toto contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""CLI to check local materials.
+"""
+# TODO: Remove with in-toto/in-toto#580
+# pylint: disable=bad-indentation
+
+import argparse
+import logging
+import sys
+
+from in_toto.common_args import (
+  LSTRIP_PATHS_ARGS,
+  LSTRIP_PATHS_KWARGS,
+  VERBOSE_ARGS,
+  VERBOSE_KWARGS,
+  sort_action_groups,
+  title_case_action_groups,
+)
+from in_toto.models.metadata import Metadata
+from in_toto.runlib import in_toto_check_materials as check_materials
+
+LOG = logging.getLogger(__name__)
+
+
+def create_parser():
+    """Create parser."""
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description="Check if local materials match products in passed link.",
+    )
+    required = parser.add_argument_group("required named arguments")
+    required.add_argument(
+        "-l",
+        "--link",
+        type=str,
+        required=True,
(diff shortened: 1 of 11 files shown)


lukpueh:
kudos to @adityasaky for starting this feature in #582 🎉

in_toto/in_toto_check_materials.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+    # raise Exception("test")
+    if only_products or only_materials or differ:
+        if args.verbose:
+            print("Local materials don't match products in passed link.")
+            for name in only_products:
+                print(f"Only in products: {name}")
+
+            for name in only_materials:
+                print(f"Only in materials: {name}")
```
adityasaky: Can we say something other than `materials`? Depending on context of use, it may not fir the `materials` terminology, say at the last step.

adityasaky (approved):
Very neat and clean! Thanks for picking this one up. :smile:

in_toto/in_toto_check_materials.py
lukpueh: Sure. Although, in the last step we'd still be checking the materials of the last step, right? I think the terminology fits the use case that motivated this. But you are right that the tool could be used for something else too. Let me adopt the naming you suggested in the original PR.

adityasaky: Apologies, I was thinking of how we'd align it with final product verification if we wanted to in future.

lukpueh: I renamed in the last commit. Please re-approve.

tests/test_runlib.py
```diff
@@ -1042,5 +1043,85 @@ def test_create_metadata_with_command_byproducts_environment(self):
     self.assertDictEqual(link.signed.environment, environment)
     os.remove(self.link_name)
 
+
+class InTotoMatchProducts(TmpDirMixin, unittest.TestCase):
```
```suggestion
class TestInTotoMatchProducts(TmpDirMixin, unittest.TestCase):
```

nit: consistency

tests/test_in_toto_match_products.py
```diff
@@ -0,0 +1,82 @@
...
+# pylint: disable=bad-indentation
+import unittest
+from pathlib import Path
+from unittest.mock import patch
+
+import in_toto.in_toto_match_products as cli
+from in_toto.models.metadata import Metadata
+
+
+class InTotoMatchProducts(unittest.TestCase):
```
adityasaky:
```suggestion
class TestInTotoMatchProducts(unittest.TestCase):
```

nit: consistency

adityasaky (approved):

adityasaky:
Thanks @lukpueh!

PR #585, opened 2023-04-27
Add OSTree ITE-4 resolver

adityasaky:
Please fill in the fields below to submit a pull request.  The more information
that is provided, the better.

**Fixes issue #**: N/A

**Description of the changes being introduced by the pull request**:

Adds an OSTree specific resolver based off @lukpueh's work in #584.

NOTE: #584 must be merged first.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

Issue #565, opened 2023-03-22
artifact filename mangling might break valid unix file paths


lukpueh:
*[based on X41 source code audit informational note]*

`in_toto.runlib.record_artifacts_as_dict` mangles file names to provide consistency between different operating systems. But replacing a backslash with a slash, can break valid Unix file paths:

https://github.com/in-toto/in-toto/blob/fe5ec9a4fc6eefeae9df2ab424ffeedce190c16d/in_toto/runlib.py#L260-L263

**Solution Advice**

X41 recommends to normalize file paths by d by using `file://` URLs that are handled the same on all supported operating systems. 

[ITE-4](https://github.com/in-toto/ITE/blob/master/ITE/4/README.adoc) defines the necessary specification change, and https://github.com/in-toto/in-toto/pull/536 provides a implementation.

SantiagoTorres:
Fixed through #578

PR #590, opened 2023-05-05
Add dirHash resolver

adityasaky:
that is provided, the better.

**Fixes issue #**: N/A

Adds implementation of a dirHash resolver as described in https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#dirhash.

requirements**:


adityasaky:
Some interesting results here, I suspect it's because of the `locale.setlocale` not matching up. :thinking:

adityasaky:
`locale.setlocale` didn't work for both `LC_ALL` and `LC_COLLATE` on MacOS and Windows. I need to read the docs to identify how to differentiate by platform here (I'm mostly surprised MacOS isn't working) but it'll have to be later. In the meantime, @lukpueh if you have any tips, lmk!

adityasaky:
Perhaps running https://github.com/in-toto/attestation/blob/main/spec/v1.0/digest_set.md#dirhash locally on MacOS would shed some light as well.

lukpueh:
> In the meantime, @lukpueh if you have any tips, lmk!

Unfortunately, I don't.

> Perhaps running https://github.com/in-toto/attestation/blob/main/spec/v1.0/digest_set.md#dirhash locally on MacOS would shed some light as well.

The link 404s

adityasaky:
> https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#dirhash

The link changed to drop the `.0` from the spec version. Here's the new one!

adityasaky:
This is the command:

`find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '`

Could you try it on the test directory I've included in this PR?

lukpueh:
Here's what I get from within `/tests/resolver/dir_resolver`:

```bash
find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
065702044fb68ae689c8cf9ae62d33ea974bc7107dc6c1fff73028f2e7411431
```

But the more interesting bit wrt your failing tests on macOS and Windows  is probably the git clone warning about case-sensitive paths n a case-insensitive fs:

```bash
git clone -b add-dirhash-resolver git@github.com:adityasaky/in-toto.git in-toto-aditya
Cloning into 'in-toto-aditya'...
remote: Enumerating objects: 9504, done.
remote: Counting objects: 100% (730/730), done.
remote: Compressing objects: 100% (286/286), done.
remote: Total 9504 (delta 466), reused 637 (delta 424), pack-reused 8774
Receiving objects: 100% (9504/9504), 2.67 MiB | 3.60 MiB/s, done.
Resolving deltas: 100% (6595/6595), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'tests/resolver/dir_resolver/subdir/A.txt'
  'tests/resolver/dir_resolver/subdir/a.txt'
  'tests/resolver/dir_resolver/subdir/B.txt'
  'tests/resolver/dir_resolver/subdir/b.txt'
```

adityasaky:
Thanks @lukpueh, that was helpful. The macOS failure was indeed because of the case insensitive collision. That leaves Windows to work out.

lukpueh:
> That leaves Windows to work out.

I've often used https://github.com/mxschmitt/action-tmate lately to figure out breaking CI on Windows runners

adityasaky:
Thanks, that was helpful! I've taken the liberty of not allowing users to pass in `normalize_line_endings`, instead setting it to true always. That turned out to be the mismatch.

tests/test_directory_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+<Started>
+  May 05, 2023
+
+<Copyright>
+  See LICENSE for licensing information.
+
+<Purpose>
+  Test dir resolver.
+
+"""
```
lukpueh: nit: I suggest we drop the doc header boilerplate in new modules.

tests/test_directory_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+import unittest
+
+from in_toto.resolver import DirectoryResolver
+
+
+class TestDirectoryResolver(unittest.TestCase):
+    """Test directory resolver implementation."""
+
+    def test_regular_directory(self):
+        resolver = DirectoryResolver(base_path="tests/resolver")
```
lukpueh: https://github.com/in-toto/in-toto/pull/585#discussion_r1188664663

lukpueh: same applies below

lukpueh: `Path(__file__).parent / "tests" / "resolver"` could be a module global.

tests/test_directory_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+        resolver = DirectoryResolver(lstrip_paths=["tests/resolver/"])
+
+        full_uri = "dir:tests/resolver/dir_resolver"
+        stripped_uri = "dir:dir_resolver"
+
+        # Expected hash calculated using variation of:
+        # find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
+        expected_artifact_dict = {
+            stripped_uri: {
+                "sha256": "87454edb176e16e396856e5be6c73ae51659785aa65999f3b51ff06d8fc59b8e"
```
lukpueh: Oh interesting. Just realized that the paths are included when calculating the hash. That is why the hash is different  for different combinations of base_path, lstrip_paths and uri arguments, even if they record the same files.

Is this obvious to the typical `DirectoryResolver` user? I think it should be documented somewhere.

lukpueh: Also, is this expected to be cross-platform compatible? Or what's the output of `find` on Windows? What path separators does it use?

in_toto/resolver/_resolver.py
```diff
@@ -262,3 +268,95 @@ def hash_artifacts(self, uris):
...
+        return {_HASH_ALGORITHM: digest_obj.hexdigest()}
+
+    def hash_artifacts(self, uris):
+        hashes = {}
+
+        for path in uris:
+            path = self._strip_scheme_prefix(path)
+
+            exhaustive_hashes = self.file_resolver.hash_artifacts([path])
+            if len(exhaustive_hashes) == 0:
```
lukpueh: Thy pythonic way would be...
```suggestion
            if not exhaustive_hashes:
```

tests/test_directory_resolver.py
lukpueh: I see you use file resolver to replicate `find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum`. I think this should be briefly mentioned, especially since FileResolver does some very opinionated file name mangling, which very likely differs form above command.

in_toto/resolver/_resolver.py
```diff
@@ -262,3 +268,95 @@ def hash_artifacts(self, uris):
...
+                "Prefix selection has resulted in non unique dictionary key "
+                f"'{path}'"
+            )
+
+        # Prepend passed scheme prefix
+        path = self.SCHEME + ":" + path
+
+        return path
+
+    def _hash(self, exhaustive_hashes):
```
lukpueh: nit: `exhaustive_hashes` is a long and not very descriptive name. Would `file_hashes` also make sense?

in_toto/resolver/_resolver.py
```diff
@@ -262,3 +268,95 @@ def hash_artifacts(self, uris):
...
+
+        digest_obj = digest(_HASH_ALGORITHM)
+        digest_obj.update(text_repr.encode("utf-8"))
+
+        return {_HASH_ALGORITHM: digest_obj.hexdigest()}
+
+    def hash_artifacts(self, uris):
+        hashes = {}
+
+        for path in uris:
```
lukpueh, line 358: What's the expected behaviour if `path` is not a directory?

tests/test_directory_resolver.py
lukpueh: Oh, _lstrip_paths isn't passed on to FileResolver, so this is not considered when calculating the hash. This means base_path is responsible for the different hashes?

in_toto/resolver/_resolver.py
```diff
@@ -262,3 +268,95 @@ def hash_artifacts(self, uris):
...
+    def _strip_scheme_prefix(self, path):
+        """Helper to strip file resolver scheme prefix from path."""
+
+        return path[len(self.SCHEME + ":") :]
+
+    def _mangle(self, path, existing_paths):
+        """Helper for path mangling."""
+
+        # Normalize slashes for cross-platform metadata consistency
+        path = path.replace("\\", "/")
```
lukpueh, line 314: See https://github.com/in-toto/in-toto/issues/565.

in_toto/resolver/_resolver.py
```diff
@@ -262,3 +268,95 @@ def hash_artifacts(self, uris):
...
+
+        # Left-strip names using configured path prefixes
+        for lstrip_path in self._lstrip_paths:
+            if path.startswith(lstrip_path):
+                path = path[len(lstrip_path) :]
+                break
+
+        # Fail if left-stripping above results in duplicates
+        if self._lstrip_paths and path in existing_paths:
```
lukpueh, line 328: See https://github.com/in-toto/in-toto/issues/601.

lukpueh (changes requested):
Cool stuff, @adityasaky! This should work, but I'm a little worried about surprising behaviour, especially since your config parameters may be relevant for DirectoryResolver itself, but also for the FileResolver used under the hood. How this currently works is not obvious, at least to me. More specifically, why are all but one parameters passed on to FileResolver? 

I can see good reasons to do it the way you do it, but also good reasons to do it differently. Here are some thoughts.

- The original purpose of `lstrip_path`s and `base_path` was to not have absolute paths in materials/products dicts. I think they should stay in DirectoryResolver, to affect the returned URIs and not be passed on to FileResolver, where they affect the hash. Also, you may consider supporting only one of them.
- Instead of passing `base_path` to FileResolver, I suggest to create a new FileResolver for each path in the passed URIs, and pass that path as `base_path` for FileResolver. This maps well to the `find <path> ...` command, which prints file names, relative to <path>. This should also make the hash of a directory reproducible, regardless of these arguments.
- Applying `exclude_patterns` on the URIs explicitly passed to `DirectoryResolver` probably does not make sense. Still, if they are only used inside `FileResolver` this should be documented.
- The `follow_symlink_dirs` and `normalize_line_endings` obviously belong to FileResolver. Should both be configurable via DirectoryResolver? Currently, only `follow_symlink_dirs` is.

tests/test_directory_resolver.py
adityasaky: No, normalize_file_endings was responsible for the differences. I've set that to True always now.

in_toto/resolver/_resolver.py
adityasaky: Good call. I think if it's a file it just hashes that one file and moves on. Should we check at this point and `continue` or error out?

adityasaky: I've added a check to `continue` and a test, lmk what you think.

tests/test_directory_resolver.py
lukpueh: I meant different hashes in `test_regular_directory` and `test_regular_directory_with_lstrip`

adityasaky: Correct, in the second example, the hash is recorded from a different directory so the path per file is different.

I can see that being confusing though. Should we drop one of lstrip / base_path? Alternatively, we can pass lstrip to the file resolver too and make it fully consistent.

in_toto/resolver/_resolver.py
adityasaky: My suggestion is the check happens here: https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L187, the snippet in the issue as well.

That way a resolver instance is only responsible for its own internal collisions.

in_toto/resolver/_resolver.py
adityasaky: Do you suggest removing this altogether? The dir resolver always has the prefix so it can be taken to be platform specific always.

tests/test_directory_resolver.py
lukpueh: > Should we drop one of lstrip / base_path? Alternatively, we can pass lstrip to the file resolver too and make it fully consistent.

Did you see my review summary comment above? https://github.com/in-toto/in-toto/pull/590#pullrequestreview-1427805634

in_toto/resolver/_resolver.py
lukpueh: I think cross-platform metadata consistency is still the goal, right? Why not use a standard file/directory path representation as #565 suggests?

If you want something quick, I suggest to leave it as is. But at some point we should address the issue, and we should probably do this in `FileResolver`, which also still uses this hack.

in_toto/resolver/_resolver.py
lukpueh: Agreed! If we check, we should check in both places. #601 also suggests to just warn instead of raising an error. What do you think?

in_toto/resolver/_resolver.py
lukpueh: I think either way is fine. Maybe add a log statement?  Regardless, the base path join you do here seems overly complicated. Did you read my comment about base_path in the review summary https://github.com/in-toto/in-toto/pull/590#pullrequestreview-1427805634?

in_toto/resolver/_resolver.py
adityasaky: I responded there. I don't know that this is in scope of this PR anymore, resolving this thread for now.

in_toto/resolver/_resolver.py
adityasaky: I've removed this entirely. The tests now don't use hard coded paths any more so there aren't any errors either.

in_toto/resolver/_resolver.py
adityasaky: Added a log statement.

tests/test_directory_resolver.py
adityasaky: I've dropped base path for the directory resolver, I pass on the dir path in each instance as base path to the file resolver, then hashing `["."]` to directly mimic the CLI commands. I've retained lstrip that only applies in the dir resolver context. Other params are passed on to the file resolver, noted in the docs.

in_toto/resolver/_resolver.py
```diff
@@ -259,3 +265,110 @@ def hash_artifacts(self, uris):
...
+
+            file_resolver = FileResolver(
+                base_path=path,
+                exclude_patterns=self._exclude_patterns,
+                follow_symlink_dirs=self._follow_symlink_dirs,
+                normalize_line_endings=self._normalize_line_endings,
+            )
+
+            file_hashes = file_resolver.hash_artifacts(["."])
+            if not file_hashes:  # pragma: no cover
```
lukpueh: Why `# pragma: no cover`? Can this not be tested? If you just don't have a test yet, I think it's more transparent to let coveralls flag it.

in_toto/resolver/_resolver.py
```diff
@@ -259,3 +265,110 @@ def hash_artifacts(self, uris):
...
+    """Directory resolver implementation.
+
+    Implementation of the following shell command for some directory URI:
+
+    find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum | sha256sum |
+        cut -f1 -d' '
+
+    This resolver uses `FileResolver` to calculate the hashes of files in the
+    directory. Therefore, `exclude_patterns`, `follow_symlink_dirs`, and
+    `normalize_line_endings` are passed on to the `FileResolver` instance.
```
lukpueh: nit: I think this could be more user friendly. The caller does not really need to know about `FileResolver`. Maybe something along the lines *... `exclude_patterns` are not applied on the directory paths passed to hash_artifacts, but only on the files inside the directory. Similarly, `follow_symlink_dirs` and `normalize_line_endings` are used on subdirectories and files inside ...*

in_toto/resolver/_resolver.py
```diff
@@ -259,3 +265,110 @@ def hash_artifacts(self, uris):
...
+
+    def hash_artifacts(self, uris):
+        hashes = {}
+
+        for path in uris:
+            path = self._strip_scheme_prefix(path)
+
+            if not os.path.isdir(path):
+                logger.info("path: %s is not a directory, skipping...", path)
+                continue
```
lukpueh: I know I said that a warning seems fine. But maybe your [argument](https://github.com/in-toto/in-toto/issues/601#issuecomment-1561890703) about unexpected metadata results being a good reason for erroring out also applies here? It seem like a usage error if you pass a file path to a directory hasher.

in_toto/resolver/_resolver.py
lukpueh: Regardless, is skipping empty directory correct? The `find` command you are trying to replicate does return a hash in an empty directory, that of an empty string (w/o newline). Should your _hash function support that?

in_toto/resolver/_resolver.py
lukpueh: But this means, that artifact paths in link metadata will be different on different platforms.

lukpueh: And it makes lstrip paths, as they are implemented now, platform dependent too.

lukpueh (changes requested):
This looks quite good modulo some minor nits inline. Do you want to move forward without cross-platform consistency? I'm okay with it, but at some point we should figure out what's right here.

tests/test_directory_resolver.py
```diff
@@ -0,0 +1,148 @@
...
+                "sha256": "ecdbcdc6bd5d2966ad1f7595874fcd6f505abd3feaa97e27bae74bcb78c38e54"
+            }
+        }
+
+        artifact_dict = resolver.hash_artifacts([uri])
+        self.assertEqual(artifact_dict, expected_artifact_dict)
+
+    def test_regular_directory_with_lstrip(self):
+        parent_path = f"{Path(__file__).parent}/resolver/"
+        target_path = f"{parent_path}dir_resolver"
```
lukpueh: This test (and the next one) won't work on a platform that requires backslashes as path separators. I think you should:

```suggestion
        lstrip_path = f"{Path(__file__).parent}/resolver/"
        target_path = str(Path(__file__).parent / "resolver" / "dir_resolver")
```

lukpueh: ... of course this requires normalising the path before lstripping.

in_toto/resolver/_resolver.py
adityasaky: Good point! Updated to support recording empty dir with corresponding change to test.

in_toto/resolver/_resolver.py
adityasaky: Updated!

in_toto/resolver/_resolver.py
adityasaky: Good point. I've added them back, shall we pick this up outside the context of this PR across all resolvers?

adityasaky:
FYI: I'll rebase and squash after this is good to go.

lukpueh (approved):
Great work, @adityasaky. LGTM!

PR #598, opened 2023-05-11
ci: remove Appveyor config and docs

lukpueh:
We run Windows builds on GHA since #513, and relatively reliably now. This should become even better with #597.

Let's drop the legacy Windows CI on Appveyor.


adityasaky (approved):

adityasaky:
Is the plan to hold off merging and rebasing after #597 is merged?

PR #602, opened 2023-05-17
Reconfigure pylint and fix or exempt new warnings

lukpueh:
This PR 
- moves pylint config from pylintrc to pyproject.toml, which is state of the art
- adopts a more succinct and restrictive pylint configuration (as used in python-tuf and securesystemslib)
- fixes new warnings, if feasible, i.e. changes are minimal
- otherwise disables warnings as locally as possible, to appease the linter for existing issues, but allow it to flag issues in new code

Although the diff is big the changes are very repetitive and grouped in commits per linter warning.


tests/__init__.py
```diff
@@ -1,11 +1,14 @@
 # Copyright New York University and the in-toto contributors
 # SPDX-License-Identifier: Apache-2.0
+"""Test in-toto. """
```
adityasaky, line 3:
```suggestion
"""Test in-toto."""
```

tests/test_in_toto_match_products.py
```diff
@@ -1,3 +1,5 @@
+"""Test in_toto_match_products.py """
```
adityasaky:
```suggestion
"""Test in_toto_match_products.py."""
```

adityasaky (approved):
LGTM, nits. Should module level docstrings have a trailing space?

tests/__init__.py
lukpueh: Yes, black does this.

tests/test_in_toto_match_products.py
lukpueh: I think `pydocstyle` would check this.

Issue #417, opened 2020-12-08
Change expiration check to match TUF


joshuagl:
**Description of issue or feature request**:

There was a discussion in the issue tracker of TUF reference implementation (https://github.com/theupdateframework/tuf/issues/1231) about how expiration datetime's should be compared.

Having done some research into how other security providers are comparing expiration equivalents (i.e. OpenSSL x509 certificate checking code, and GnuPG expiration checks), and how other TUF implementations are performing the same check (rust-tuf, go-tuf), we came to a consensus that the correct way to implement expiration comparisons is:

> expiration <= now

I'm filing this issue because the discussion in the thread suggested it is preferable to have in-toto and TUF specifications and reference implementations behave in the same way on this.

**Current behavior**:

`expires < now`

**Expected behavior**:

`expires <= now`

joshuagl:
It may make sense to hold on updating the reference implementation until a similar change has been made to the specification, see https://github.com/in-toto/docs/issues/42

adityasaky:
The spec change has now been made.

MashiatK:
hi, could i take up the issue?

adityasaky:
Go ahead!

ashutosh887:
@joshuagl @adityasaky Please let me work on this Issue

lukpueh:
Thanks for volunteering, @ashutosh887! We already have a patch for this in: https://github.com/in-toto/in-toto/pull/545.

If you are looking for a good first issue, I recommend: https://github.com/in-toto/in-toto/issues/548

adityasaky:
Closed by #616

PR #615, opened 2023-07-06
Added Current Datetime Expiry Test Cases

SpencerKlem:
that is provided, the better.

**Fixes issue #545**:

Using a mock of the datetime.now() function, adds test cases for issue introduced #417 and to check for wanted behaviour from raising a layout expiration error

requirements**:

- [✓] The code follows the [Code Style Guidelines](https://github.com/secure-systems-lab/code-style-guidelines#code-style-guidelines)
- [✓] Tests have been added for the bug fix or new feature
- [✓] Docs have been added for the bug fix or new feature

### DIFF
--- a/tests/test_verifylib.py
+++ b/tests/test_verifylib.py
@@ -932,6 +932,7 @@ class TestInTotoVerify(unittest.TestCase, TmpDirMixin):
         # Store various layout paths to be used in tests
         cls.layout_single_signed_path = "single-signed.layout"
         cls.layout_double_signed_path = "double-signed.layout"
+        cls.layout_current_date_expiry_path = "current-date-expiry.layout"
         cls.layout_bad_sig = "bad-sig.layout"
         cls.layout_expired_path = "expired.layout"
         cls.layout_failing_step_rule_path = "failing-step-rule.layout"
@@ -959,6 +960,14 @@ class TestInTotoVerify(unittest.TestCase, TmpDirMixin):
         layout.sign(alice)
         layout.sign(bob)
         layout.dump(cls.layout_double_signed_path)
+
+        # dump layout with current date expiry
+        layout = copy.deepcopy(layout_template)
+        layout.signed.expires = (
+            datetime.today()
+        ).strftime("%Y-%m-%dT%H:%M:%SZ")
+        layout.sign(alice)
+        layout.dump(cls.layout_current_date_expiry_path)
 
         # dump layout with bad signature
         layout = copy.deepcopy(layout_template)
@@ -1022,6 +1031,12 @@ class TestInTotoVerify(unittest.TestCase, TmpDirMixin):
         layout_key_dict = import_publickeys_from_file(
             [self.alice_path, self.bob_path]
         )
+        in_toto_verify(layout, layout_key_dict)
+
+    def test_verify_passing_layout_current_date(self):
+        """Test pass verification with current date expiry layout."""
+        layout = Metablock.load(self.layout_current_date_expiry_path)
+        layout_key_dict = import_publickeys_from_file([self.alice_path])
         in_toto_verify(layout, layout_key_dict)
 
     def test_verify_passing_empty_layout(self):
--- a/tests/test_verifylib.py
+++ b/tests/test_verifylib.py
@@ -81,7 +81,7 @@ from in_toto.verifylib import (
     verify_sublayouts,
     verify_threshold_constraints,
 )
-from common import GPGKeysMixin, TmpDirMixin
+from tests.common import GPGKeysMixin, TmpDirMixin
 
 
 class TestRaiseOnBadRetval(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_verifylib.py
```diff
@@ -82,7 +82,8 @@
     verify_threshold_constraints,
 )
 from tests.common import GPGKeysMixin, TmpDirMixin
-
+import datetime
```
adityasaky, line 85: Let's fix up the datetime imports so it's imported just once.

tests/test_verifylib.py
```diff
@@ -969,10 +979,14 @@ def setUpClass(cls):
         # dump expired layout
         layout = copy.deepcopy(layout_template)
         layout.signed.expires = (
-            datetime.today() + relativedelta(months=-1)
+            datetime.datetime.now() + relativedelta( months=-4)
         ).strftime("%Y-%m-%dT%H:%M:%SZ")
         layout.sign(alice)
         layout.dump(cls.layout_expired_path)
+
```
adityasaky, line 986: nit: let's remove these lines

tests/test_verifylib.py
```diff
@@ -969,10 +979,14 @@ def setUpClass(cls):
         # dump expired layout
         layout = copy.deepcopy(layout_template)
         layout.signed.expires = (
-            datetime.today() + relativedelta(months=-1)
+            datetime.datetime.now() + relativedelta( months=-4)
```
adityasaky, line 982: This change is out of scope for this PR.

adityasaky: And we probably don't need to change this for the the < to <= change made in this PR.

tests/test_verifylib.py
```diff
@@ -960,6 +962,14 @@ def setUpClass(cls):
         layout.sign(bob)
         layout.dump(cls.layout_double_signed_path)
 
+        # dump layout with current date expiry
+        layout = copy.deepcopy(layout_template)
+        layout.signed.expires = (
+            datetime.datetime.now() +
```
adityasaky, line 968: Let's verify what expiry is set.

adityasaky: Here, the layout is set to expire `now()`. Right now, you're testing with `now()` a few seconds later and the expected error is indeed returned. This doesn't, however, test the change in behavior from `<` to `<=`.

tests/test_verifylib.py
```diff
@@ -1023,6 +1037,20 @@ def test_verify_passing_double_signed_layout(self):
             [self.alice_path, self.bob_path]
         )
         in_toto_verify(layout, layout_key_dict)
+    
+    def test_verify_failing_layout_current_datetime(self):
+        with patch('datetime' + '.date') as mock_date:
```
adityasaky, line 1042: Consider what you want to be patching. The expiry check uses `datetime.datetime.now()`.

adityasaky: Also, check that the mock holds through the in_toto_verify call.

tests/test_verifylib.py
```diff
@@ -1023,6 +1037,20 @@ def test_verify_passing_double_signed_layout(self):
...
+            mock_date.now.return_value = datetime.datetime(2017, 6, 5, 4, 3, 2) 
+            mock_date.side_effect = lambda *args, **kw: date(*args, **kw)
+
+            assert datetime.date.now() == datetime.datetime(2017, 6, 5, 4, 3, 2) 
+            assert datetime.date(2009, 6, 8) == date(2009, 6, 8)
+            """Test pass verification with current date expiry layout."""
+            layout = Metablock.load(self.layout_current_datetime_expiry_path)
+            layout_key_dict = import_publickeys_from_file([self.alice_path])
+            with self.assertRaises(LayoutExpiredError):
```
adityasaky, line 1051: This currently passes because the expiry is set to `now()` at set up time and the check happens a few moments later when the test is executed. We want to test for the change in behavior. I recommend mocking the expiry check for two layouts:
1. The layout expiry and the test time are identical, resulting in an expected error (earlier, the two being the same would not cause an error)
2. The layout expiry is a second after the test time (mocked), resulting in no error.

tests/test_verifylib.py
```diff
@@ -1051,6 +1079,7 @@ def test_verify_failing_layout_expired(self):
         with self.assertRaises(LayoutExpiredError):
             in_toto_verify(layout, layout_key_dict)
 
+
```
adityasaky, line 1082: nit: unrelated to PR

tests/test_verifylib.py
```diff
@@ -1819,3 +1848,4 @@ def test_get_summary_link_from_demo_layout(self):
 
 if __name__ == "__main__":
     unittest.main()
+
```
adityasaky, line 1851: nit: unrelated to PR

adityasaky (changes requested):

PR #616, opened 2023-07-12
Update layout expiry behavior

SpencerKlem:
that is provided, the better.

**Fixes issue #545**:

Adding test cases, has bug
requirements**:

### DIFF
--- a/tests/test_verifylib.py
+++ b/tests/test_verifylib.py
@@ -29,7 +29,7 @@ import shlex
 import shutil
 import tempfile
 import unittest
-from datetime import datetime
+from datetime import datetime, timezone
 from unittest.mock import patch
 
 import securesystemslib.exceptions
@@ -932,6 +932,7 @@ class TestInTotoVerify(unittest.TestCase, TmpDirMixin):
         # Store various layout paths to be used in tests
         cls.layout_single_signed_path = "single-signed.layout"
         cls.layout_double_signed_path = "double-signed.layout"
+        cls.layout_fixed_expiry_path = "fixed-datetime-expiry.layout"
         cls.layout_bad_sig = "bad-sig.layout"
         cls.layout_expired_path = "expired.layout"
         cls.layout_failing_step_rule_path = "failing-step-rule.layout"
@@ -959,6 +960,14 @@ class TestInTotoVerify(unittest.TestCase, TmpDirMixin):
         layout.sign(alice)
         layout.sign(bob)
         layout.dump(cls.layout_double_signed_path)
+
+        # dump layout with fixed (and mocked) date expiry
+        layout = copy.deepcopy(layout_template)
+        layout.signed.expires = (
+            datetime(2025, 1, 2, 3, 4, 6, tzinfo=timezone.utc)
+        ).strftime("%Y-%m-%dT%H:%M:%SZ")
+        layout.sign(alice)
+        layout.dump(cls.layout_fixed_expiry_path)
 
         # dump layout with bad signature
         layout = copy.deepcopy(layout_template)
@@ -1030,6 +1039,18 @@ class TestInTotoVerify(unittest.TestCase, TmpDirMixin):
         layout_key_dict = import_publickeys_from_file([self.alice_path])
         in_toto_verify(layout, layout_key_dict)
 
+    def test_verify_passing_layout_expiry(self):
+        """Test pass verification at expire < current timestamp."""
+        with patch("""in_toto.verifylib.datetime""") as mock_date:
+            mock_date.datetime.now.return_value = datetime(
+                2025, 1, 2, 3, 4, 5, tzinfo=timezone.utc
+            )
+            layout = Metablock.load(self.layout_fixed_expiry_path)
+            layout_key_dict = import_publickeys_from_file([self.alice_path])
+            # the expiry is a second later than the mocked datetime.datetime.now(), which
+            # should not raise an exception due to <= behaviour of expiration comparison
+            in_toto_verify(layout, layout_key_dict)
+
     def test_verify_failing_wrong_key(self):
         """Test fail verification with wrong layout key."""
         layout = Metablock.load(self.layout_single_signed_path)
@@ -1050,6 +1071,19 @@ class TestInTotoVerify(unittest.TestCase, TmpDirMixin):
         layout_key_dict = import_publickeys_from_file([self.alice_path])
         with self.assertRaises(LayoutExpiredError):
             in_toto_verify(layout, layout_key_dict)
+
+    def test_verify_failing_layout_at_expire_timestamp(self):
+        """Test fail expiry verification at current==expire timestamp."""
+        with patch("""in_toto.verifylib.datetime""") as mock_date:
+            mock_date.datetime.now.return_value = datetime(
+                2025, 1, 2, 3, 4, 6, tzinfo=timezone.utc
+            )
+            layout = Metablock.load(self.layout_fixed_expiry_path)
+            layout_key_dict = import_publickeys_from_file([self.alice_path])
+            # the expiry is the same as the mocked datetime.datetime.now(), which
+            # should raise an exception due to <= behaviour of expiration comparison
+            with self.assertRaises(LayoutExpiredError):
+                in_toto_verify(layout, layout_key_dict)
 
     def test_verify_failing_link_metadata_files(self):
         """Test fail verification with link metadata files not found."""
--- a/in_toto/verifylib.py
+++ b/in_toto/verifylib.py
@@ -266,7 +266,7 @@ def verify_layout_expiration(layout):
 
     """
     expire_datetime = iso8601.parse_date(layout.expires)
-    if expire_datetime < datetime.datetime.now(tz.tzutc()):
+    if expire_datetime <= datetime.datetime.now(tz.tzutc()):
         raise LayoutExpiredError("Layout expired")
 
 


tests/test_verifylib.py
```diff
@@ -960,6 +960,14 @@ def setUpClass(cls):
         layout.sign(bob)
         layout.dump(cls.layout_double_signed_path)
 
+        # dump layout with current date expiry
+        layout = copy.deepcopy(layout_template)
+        layout.signed.expires = (
+            datetime.now()
```
adityasaky: Here, we have an expiry of 2023-07-...

tests/test_verifylib.py
```diff
@@ -1023,7 +1031,19 @@ def test_verify_passing_double_signed_layout(self):
             [self.alice_path, self.bob_path]
         )
         in_toto_verify(layout, layout_key_dict)
-
+    
+    def test_verify_failing_layout_current_datetime(self):
+        """Test pass verification with current date expiry layout."""
+        with patch("""in_toto.verifylib.datetime.datetime""") as mock_date:
+            with patch("""in_toto.verifylib.iso8601""") as mock_date2:
```
adityasaky: I don't think we'll need to patch this.

tests/test_verifylib.py
```diff
@@ -1023,7 +1031,19 @@ def test_verify_passing_double_signed_layout(self):
...
+                mock_date.now.return_value= datetime(2025, 1, 2, 3, 4, 6)
+                mock_date2.parse_date.return_value = datetime(2025, 1, 2, 3, 4, 6)
+                assert in_toto.verifylib.datetime.datetime.now() == in_toto.verifylib.iso8601.parse_date(datetime.now())
+                layout = Metablock.load(self.layout_current_datetime_expiry_path)
```
adityasaky: I see, so the layout you configure above is not being used.

tests/test_verifylib.py
```diff
@@ -1023,7 +1031,19 @@ def test_verify_passing_double_signed_layout(self):
...
+                layout_key_dict = import_publickeys_from_file([self.alice_path])
+                with self.assertRaises(LayoutExpiredError):
+                    in_toto_verify(layout, layout_key_dict)
```
adityasaky:
```
-> if expire_datetime <= datetime.datetime.now(tz.tzutc()):
(Pdb) expire_datetime
datetime.datetime(2025, 1, 2, 3, 4, 6)
(Pdb) datetime.datetime.now()
datetime.datetime(2025, 1, 2, 3, 4, 6)
(Pdb)
```

A breakpoint in verifylib indicates this works fine **logically**, and the test indeed passes. However, I think the setup can be better.

tests/test_verifylib.py
adityasaky: Notice that you don't need to mock datetime to set an expiry. You can set _any_ value you want here and it does not matter.

tests/test_verifylib.py
adityasaky: Let's configure the layout in setUp for consistency. Pick a date, set it as the expiry. Then mock verifylib's datetime here so that the expiry check uses the timestamp you picked.

tests/test_verifylib.py
```diff
@@ -1081,7 +1101,7 @@ def test_verify_failing_inspection_rules(self):
         with self.assertRaises(RuleVerificationError):
             in_toto_verify(layout, layout_key_dict)
 
-    def test_verify_layout_signatures_fail_with_no_keys(self):
+    def test_signatures_fail_with_no_keys(self):
```
adityasaky: what's this change for, sorry?

adityasaky (changes requested):
I think logically this coming along. Note that you don't have to close this PR to push more changes, you can just push to the same branch. :)

SpencerKlem:
Current issue: verifylib.py is having trouble comparing the two mocked dates as one is timezone aware and the other date is not.

tests/test_verifylib.py
```diff
@@ -1087,7 +1106,7 @@ def test_verify_layout_signatures_fail_with_no_keys(self):
         with self.assertRaises(SignatureVerificationError):
             in_toto_verify(layout_metablock, {})
 
-    def test_verify_layout_signatures_fail_with_malformed_signature(self):
+    def test_signatures_fail_with_no_keys(self):
```
adityasaky: bump on this change, I'm not sure why this test name was updated?

tests/test_verifylib.py
```diff
@@ -1023,7 +1032,17 @@ def test_verify_passing_double_signed_layout(self):
             [self.alice_path, self.bob_path]
         )
         in_toto_verify(layout, layout_key_dict)
-
+    
+    def test_verify_failing_layout_current_datetime(self):
+        """Test pass verification with current date expiry layout."""
+        with patch("""in_toto.verifylib.datetime""") as mock_date:
+            mock_date.datetime.now.return_value= datetime(2025, 1, 2, 3, 4, 6).replace(tzinfo=pytz.utc)
+            assert in_toto.verifylib.datetime.datetime.now() == datetime(2025, 1, 2, 3, 4, 6).replace(tzinfo=pytz.utc)
```
adityasaky: let's get rid of this assertion

tests/test_verifylib.py
```diff
@@ -960,6 +961,14 @@ def setUpClass(cls):
         layout.sign(bob)
         layout.dump(cls.layout_double_signed_path)
 
+        # dump layout with current date expiry
+        layout = copy.deepcopy(layout_template)
+        layout.signed.expires = (
+            datetime(2025, 1, 2, 3, 4, 6).replace(tzinfo=pytz.utc)
```
adityasaky: can't tzinfo be passed in to datetime as a named parameter? :)

tests/test_verifylib.py
```diff
@@ -29,6 +29,7 @@
 import shutil
 import tempfile
 import unittest
+import pytz
```
adityasaky: Instead of using another dependency I suggest using `datetime.timezone.utc` or `dateutil.tz.tzutc`. The latter is also a third-party dependency but we already use it for `relativedelta`.

adityasaky: Also the CI's likely to fail as this isn't installed.

tests/test_verifylib.py
```diff
@@ -1023,7 +1032,17 @@ def test_verify_passing_double_signed_layout(self):
...
-
+    
+    def test_verify_failing_layout_current_datetime(self):
+        """Test pass verification with current date expiry layout."""
+        with patch("""in_toto.verifylib.datetime""") as mock_date:
+            mock_date.datetime.now.return_value= datetime(2025, 1, 2, 3, 4, 6).replace(tzinfo=pytz.utc)
+            assert in_toto.verifylib.datetime.datetime.now() == datetime(2025, 1, 2, 3, 4, 6).replace(tzinfo=pytz.utc)
+            layout = Metablock.load(self.layout_single_signed_path)
+            layout_key_dict = import_publickeys_from_file([self.alice_path])
+            with self.assertRaises(LayoutExpiredError):
```
adityasaky: Neat, I think we can also add another where the time is off by a second and the expiry check passes.

adityasaky (changes requested):

tests/test_verifylib.py
```diff
@@ -82,7 +81,7 @@
     verify_sublayouts,
     verify_threshold_constraints,
 )
-from tests.common import GPGKeysMixin, TmpDirMixin
+from common import GPGKeysMixin, TmpDirMixin
```
adityasaky: This will break the import, no?

tests/test_verifylib.py
```diff
@@ -1032,17 +1040,23 @@ def test_verify_passing_double_signed_layout(self):
...
-    def test_verify_failing_layout_current_datetime(self):
+
+    def test_verify_passing_near_layout_current_datetime(self):
         """Test pass verification with current date expiry layout."""
         with patch("""in_toto.verifylib.datetime""") as mock_date:
-            mock_date.datetime.now.return_value= datetime(2025, 1, 2, 3, 4, 6).replace(tzinfo=pytz.utc)
-            assert in_toto.verifylib.datetime.datetime.now() == datetime(2025, 1, 2, 3, 4, 6).replace(tzinfo=pytz.utc)
+            mock_date.datetime.now.return_value= datetime(2025, 1, 2, 3, 4, 5, tzinfo=timezone.utc)
+            layout = Metablock.load(self.layout_single_signed_path)
+            layout_key_dict = import_publickeys_from_file([self.alice_path])
```
adityasaky: There should be a call to in_toto_verify here that passes.

tests/test_verifylib.py
```diff
@@ -960,6 +962,22 @@ def setUpClass(cls):
...
+        # dump layout with near current date expiry
+        layout = copy.deepcopy(layout_template)
+        layout.signed.expires = (
+           datetime(2025, 1, 2, 3, 4, 6, tzinfo=timezone.utc)
+        ).strftime("%Y-%m-%dT%H:%M:%SZ")
+        layout.sign(alice)
+        layout.dump(cls.layout_near_current_datetime_expiry_path)
+
+        # dump layout with current date expiry
+        layout = copy.deepcopy(layout_template)
```
adityasaky: I think this layout is identical to the one above?

adityasaky: If yes, let's reuse the same layout.

tests/test_verifylib.py
```diff
@@ -1024,6 +1042,23 @@ def test_verify_passing_double_signed_layout(self):
...
+        with patch("""in_toto.verifylib.datetime""") as mock_date:
+            mock_date.datetime.now.return_value= datetime(2025, 1, 2, 3, 4, 5, tzinfo=timezone.utc)
+            layout = Metablock.load(self.layout_single_signed_path)
+            layout_key_dict = import_publickeys_from_file([self.alice_path])
+            in_toto_verify(layout, layout_key_dict)
+
+    def test_verify_failing_layout_current_datetime(self):
+        """Test fail verification with current date expiry layout."""
+        with patch("""in_toto.verifylib.datetime""") as mock_date:
+            mock_date.datetime.now.return_value= datetime(2025, 1, 2, 3, 4, 6, tzinfo=timezone.utc)
```
adityasaky: I think this needs to be 4 instead of 6 as you're changing the current time and not the expire time. Remember that the comparison is `expires <= current` so for that condition to be false, the current time must be at least a second before the expire time.

adityasaky (changes requested):

adityasaky (approved):

lukpueh (approved):
💯

adityasaky:
Great work, @SpencerKlem!

PR #637, opened 2023-10-06
build: split test and lint dependencies and pin

lukpueh:
* Split requirements-tox into requirements-test and requirements-lint, and update tox.ini. There is no need to install lint requirements when running tests, and vice versa.

* Pin test dependencies to detect update issues with Dependabot.

* unrelated: include doc requirements with dev requirements


lukpueh:
As expected, now that we pin test dependencies, we fail on 3.7. Rebasing on #634 will fix this...

adityasaky (approved):

Issue #533, opened 2023-01-17
Support priv_key_uri as CLI argument


lukpueh:
**Description of issue or feature request**:
The new securesystemslib [v0.26.0](https://github.com/secure-systems-lab/securesystemslib/releases/tag/v0.26.0) signer API allows loading arbitrary signing providers from [private key URIs](https://github.com/secure-systems-lab/securesystemslib/blob/v0.26.0/securesystemslib/signer/_signer.py#L87). Supporting a private key URI argument in relevant in-toto CLI allows adding and removing support for arbitrary signing providers in the future, without changing the CLI. Relevant CLI are: [`in-toto-run`](https://github.com/in-toto/in-toto/blob/6bd25002ef2d930b214288c2d8c0ef3f2a917473/in_toto/in_toto_run.py), [`in-toto-record`](https://github.com/in-toto/in-toto/blob/6bd25002ef2d930b214288c2d8c0ef3f2a917473/in_toto/in_toto_record.py).

IMPORTANT: Coordinate with #532!

**Current behavior**:
Relevant CLI don't support private key URI argument.

**Expected behavior**:
Relevant CLI support private key URI argument.

**Optional expected behavior**:
Deprecate existing signing arguments in relevant CLI (e.g. [`--key`](https://github.com/in-toto/in-toto/blob/6bd25002ef2d930b214288c2d8c0ef3f2a917473/in_toto/common_args.py#L72-L80), [`--gpg`](https://github.com/in-toto/in-toto/blob/6bd25002ef2d930b214288c2d8c0ef3f2a917473/in_toto/common_args.py#L120-L130),[`--gpg-home`](https://github.com/in-toto/in-toto/blob/6bd25002ef2d930b214288c2d8c0ef3f2a917473/in_toto/common_args.py#L132-L140)).

lukpueh:
Turns out, "priv_key_uri" does not do what I had in mind when writing this issue. That is, it does not encode all necessary information to load a signer.

Instead, the API, which takes the URI, also requires a secrets handler and a related public key

We could work around this by taking addtional options, which, together with the URI, would allow us to construct the required public key and secrets handler first. For all currently available signers, these options should work:
```
--uri <uri>  # private key uri, can also be (ab)used to infer pubkey
[--secret [<secret>]  # takes None, password, or toggles prompt
[--keyid <keyid>]  # optional non-default keyid
[--scheme <scheme>]  # optional non-default signing scheme
```

[...]

PR #649, opened 2023-11-20
Add --signing-key (pkcs8) arg to in-toto-run

lukpueh:
*EDIT 2023/12/05: rename `--pkcs8` to `--signing-key`, and test with `--use-dsse`*
Fixes part of #533 (missing: --signing-key support in in-toto-record)

----

Unlike the existing `--key`, the here added `--signing-key` supports a standard file format (PKCS8/pem), which is consistent for all supported key types and eliminates the need for an explicitly passed `--key-type`.

This PR adds a warning that `--key` is deprecated and `--signing-key` should be used instead.

In preparation for the removal of  key file loading (plus decryption via passed or prompted password) in securesystemslib,
the functionality is added to in-toto in only a few lines of code. This is possible thanks to the modern securesystemslib CryptoSigner interface.


lukpueh:
Note about CI failures:
- ~with-sslib-main fail is expected, because we use an unmerged sslib feature~
- ~lint fail is unrelated and will be fixed separately~

in_toto/common_args.py
```diff
@@ -163,6 +163,18 @@ def parse_password_and_prompt_args(args):
     ),
 }
 
+PKCS8_ARGS = ["--pkcs8"]
```
lukpueh: `pkcs8` is a bit technical. We could also call the arg:
- private-key
- signing-key
- key-file
- key-path
- ...


Any thoughts?

adityasaky: I think I prefer `--signing-key`.

tests/test_in_toto_run.py
```diff
@@ -364,6 +364,41 @@ def test_main_encrypted_key_but_no_pw(self):
         self.assert_cli_sys_exit(args, 1)
         self.assertFalse(os.path.exists(self.test_link_ed25519_enc))
 
+    def test_pkcs8(self):
+        """Test in-toto-run, sign link with pkcs8 key file for each algo."""
+        pems_dir = Path(__file__).parent / "pems"
+        args = ["-n", "foo", "-x", "--pkcs8"]
```
adityasaky: should this also test with `--use-dsse` for completeness?

adityasaky (approved):
LGTM, minus minor comments.

in_toto/common_args.py
lukpueh: Fair. Let's go with signing-key then.

PR #653, opened 2023-11-23
Use modern Signer API in Metablock.verify_signature

lukpueh:
Replace usage of legacy `securesystemslib.keys.verify_signature` with modern verification code from Signer API.

The patch is a pure refactor and does not change behavior. It is part of a series of patches to prepare for the removal of legacy securesystemslib modules.

### DIFF
--- a/in_toto/models/metadata.py
+++ b/in_toto/models/metadata.py
@@ -23,6 +23,7 @@
 """
 
 import json
+from copy import deepcopy
 from typing import Union
 
 import attr
@@ -31,8 +32,11 @@ import securesystemslib.formats
 import securesystemslib.gpg.functions
 import securesystemslib.keys
 from securesystemslib.dsse import Envelope as SSlibEnvelope
-from securesystemslib.exceptions import VerificationError
-from securesystemslib.signer import Signature, Signer, SSlibKey
+from securesystemslib.exceptions import (
+    UnverifiedSignatureError,
+    VerificationError,
+)
+from securesystemslib.signer import Key, Signature, Signer, SSlibKey
 
 from in_toto.exceptions import InvalidMetadata, SignatureVerificationError
 from in_toto.models._signer import GPGSigner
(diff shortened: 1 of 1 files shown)


in_toto/models/metadata.py
```diff
@@ -395,18 +399,28 @@ def verify_signature(self, verification_key):
...
-        elif securesystemslib.formats.SIGNATURE_SCHEMA.matches(signature):
-            valid = securesystemslib.keys.verify_signature(
-                verification_key, signature, self.signed.signable_bytes
-            )
-
         else:
-            valid = False
+            # Parse key and (below) signature dicts as `Key` and `Signature`
+            # instances to use modern securesystemslib verification code.
+            # Deepcopy to preserve original dicts, which are otherwise destroyed
```
trishankatdatadog, line 411: Hmm, interesting. Do the `from_dict` methods _need_ to be destructive?

trishankatdatadog (approved):
LGTM FWIW

in_toto/models/metadata.py
lukpueh: This aligns with usage in python-tuf's Metadata API, where dicts can be discarded once deserialised.

There it turned out useful to just pop all expected fields and cram whatever's left into  a `unrecognized_fields: Optional[Dict[str, Any]]` attribute, which exists on all classes that have a `from_dict` deserialisation helper.

Unfortunately, this doesn't work in in-toto, where we only have classes for parts of the metadata but keep most contained data (signatures, keys, etc.) as dicts.

That's what I meant with supporting Key and Signature natively being nice. I don't think it's a priority, but I am happy to submit an issue.

lukpueh:
Thanks for the review, @trishankatdatadog! I appreciate you taking the time. :)

in_toto/models/metadata.py
trishankatdatadog: Oh, I think I see, hmm. Still don't understand why we can't hide away destruction (e.g., by copying) on behalf of users, but you're right, this is not a high-priority issue, although we should make an issue just to be safe. Thanks!

PR #655, opened 2023-11-24
Replace deprecated securesystemslib method

lukpueh:
Replace deprecated `SSlibKey.from_securesystemslib_key` method call, with call to generic `Key.from_dict`.

Nice side-effect: The latter will also accept other `Key` implementations supported by the securesystemslib Signer API.

This is part of a series of patches to prepare for removal of legacy securesystemslib interfaces and key file formats.

### DIFF
--- a/in_toto/models/metadata.py
+++ b/in_toto/models/metadata.py
@@ -36,7 +36,7 @@ from securesystemslib.exceptions import (
     UnverifiedSignatureError,
     VerificationError,
 )
-from securesystemslib.signer import Key, Signature, Signer, SSlibKey
+from securesystemslib.signer import Key, Signature, Signer
 
 from in_toto.exceptions import InvalidMetadata, SignatureVerificationError
 from in_toto.models._signer import GPGSigner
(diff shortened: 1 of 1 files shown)


in_toto/models/metadata.py
```diff
@@ -168,7 +168,13 @@ def create_signature(self, signer: Signer) -> Signature:
         return super().sign(signer)
 
     def verify_signature(self, verification_key):
-        key = SSlibKey.from_securesystemslib_key(verification_key)
+        # Deepcopy to preserve `verification_key`, which might still be needed
+        # in calling context and would otherwise be destroyed in `from_dict`.
+        # NOTE: It would be nice to support `Key` natively in in-toto model.
+        key = Key.from_dict(
+            verification_key["keyid"], deepcopy(verification_key)
```
adityasaky, line 175: > NOTE: It would be nice to support `Key` natively in in-toto model.

So verify_signature would accept `Key` rather than the dict? Which would mean we wouldn't have these deepcopy concerns because we wouldn't have `from_dict` at all, right?

adityasaky (approved):

adityasaky:
LGTM (forgot to add that) minus clarification.

in_toto/models/metadata.py
lukpueh: I'll submit a ticket for it.

PR #657, opened 2023-11-27
Drop in-toto-keygen

lukpueh:
As we are switching to a standard key file format (see #649, #651 and #652), we no longer need to maintain a custom command line tool, to generate key files in a proprietary in-toto/securesystemslib format.

Docs on how to migrate existing key or generate new key files with standard tooling (`pyca/cryptography` or `openssl` cli) are available in:
https://github.com/secure-systems-lab/securesystemslib#legacy-key-migration


trishankatdatadog (approved):
🔥

PR #659, opened 2023-11-28
Deprecate Metablock.sign() and fix related docs

lukpueh:
Add deprecation note to `Metablock.sign()` method in preparation for the removal of dict-based private keys in in-toto and securesystemslib.

A more powerful replacement API exists with `Metablock.create_signature()`.


adityasaky (approved):
LGTM!

PR #664, opened 2023-11-29
Replace legacy key gen/load API in layout example

lukpueh:
Update layout creation example to use new securesystemslib Signer API to create in-memory demo keys.

This is part of a series of patches to prepare for removal of legacy securesystemslib interfaces and key file formats.


adityasaky (approved):
LGTM.

PR #665, opened 2023-12-04
Add backwards-compatible SCHEMA replacement

lukpueh:
Adds helper functions to cover all data validation cases in a backwards-compatible manner, where we previously used securesystemslib SCHEMA.

This is part of a series of patches to prepare for removal of legacy securesystemslib interfaces.  (see https://github.com/secure-systems-lab/securesystemslib/issues/183 for specific issues with SCHEMA)

**Details**

Most of the checks are simple type validations, now without the hidden redundancy of SCHEMA. (e.g. `NAME_SCHEMA`, `PATH_SCHEMA`, or `ANY_STRING_SCHEMA` all just checked if the input was a string, now there is only `_check_str`)

A few checks validate more complex data structures, most notably public key dictionaries. While SCHEMA whitelisted allowed fields and their values, we now just try to deserialise using securesystemslib's Signer API, and re-raise a deserialisation error as format error.

This is more Pythonic (if we can parse it as key, it is a key) and scales nicely, because the Signer API allows to [register new public key schemes](https://github.com/secure-systems-lab/securesystemslib/blob/cc0ead6b03ad7e42418a36799daa87a6742e60cf/securesystemslib/signer/__init__.py#L45-L69) in user code.

**Alternative / future work**

This PR is a **hack** with the goal to remove schema in securesystemslibs without breaking the API in in-toto. It also has  the nice side-effect of making validation more explicit and flexible.

A more sustainable solution can be found in the python-tuf Metadata API, where
- most of the data model (including keys and signatures) is represented by classes and validated on deserialization
- API inputs are above all validated to prevent unexpected behavior, or provide better UX
- in many cases, it's okay to just use the exceptions that come from using invalid input data
- type annotations are very useful

However, this would be a more invasive and surely breaking change. So, even though this PR adds a lot of code, I think we should go with it for now, and revise the class model later. 

Reviewing commit by commit should make the review straight-forward.

### DIFF
--- a/in_toto/formats.py
+++ b/in_toto/formats.py
@@ -133,3 +133,13 @@ def _check_public_keys(arg):
     for k, v in arg.items():
         _check_hex(k)
         _check_public_key(v)
+
+
+def _check_signing_key(arg):
+    """Check legacy signing key dict format.
+
+    NOTE: signing key dict is deprecated, this check will be removed with it
+    """
+    _check_public_key(arg)
+    if not arg["keyval"].get("private"):
+        raise _err(arg, "private key data")
--- a/in_toto/formats.py
+++ b/in_toto/formats.py
@@ -16,22 +16,120 @@
   See LICENSE for licensing information.
 
 <Purpose>
-  Format schemas for in-toto metadata, based on securesystemslib.schema.
-
-  The schemas can be verified using the following methods inherited from
-  securesystemslib.schema:
-
-  in_toto.formats.<SCHEMA>.check_match(<object to verify>)
-  in_toto.formats.<SCHEMA>.matches(<object to verify>)
-
-  `check_match` raises a securesystemslib.exceptions.FormatError and `matches`
-  returns False if the verified object does not match the schema (True
-  otherwise).
+  Helpers to validate API inputs and metadata model objects.
 
 """
+from copy import deepcopy
+from re import fullmatch
+
 import securesystemslib.schema as ssl_schema
+from securesystemslib.exceptions import FormatError
+from securesystemslib.signer import Key, Signature
+
+from in_toto.models._signer import GPGKey, GPGSignature
 
 PARAMETER_DICTIONARY_KEY = ssl_schema.RegularExpression(r"[a-zA-Z0-9_-]+")
 PARAMETER_DICTIONARY_SCHEMA = ssl_schema.DictOf(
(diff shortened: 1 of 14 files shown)


in_toto/formats.py
```diff
@@ -16,22 +16,124 @@
...
 
-  The schemas can be verified using the following methods inherited from
-  securesystemslib.schema:
+"""
+from copy import deepcopy
+from re import fullmatch
 
-  in_toto.formats.<SCHEMA>.check_match(<object to verify>)
-  in_toto.formats.<SCHEMA>.matches(<object to verify>)
+from securesystemslib.exceptions import FormatError
```
adityasaky, line 25: Should `FormatError` move into in-toto as well?

adityasaky (approved):
LGTM!

lukpueh:
Speaking of nice side-effect and flexibility. Here's me "registering" sigstore in user code and generating an in-toto link signed with my email address, with 0 changes in the library code. 😎 

```python
from securesystemslib.signer import (
    KEY_FOR_TYPE_AND_SCHEME,
    SIGNER_FOR_URI_SCHEME,
    Signer,
    SigstoreKey,
    SigstoreSigner,
)
from in_toto.runlib import in_toto_run

# All what's needed for in-toto to speak sigstore
KEY_FOR_TYPE_AND_SCHEME[("sigstore-oidc", "Fulcio")] = SigstoreKey

# This is only needed to use generic `Signer.from_priv_key_uri` interface.
# We could also load SigstoreSigner directly.
SIGNER_FOR_URI_SCHEME[SigstoreSigner.SCHEME] = SigstoreSigner

uri, key = SigstoreSigner.import_via_auth()

signer = Signer.from_priv_key_uri(uri, key)

link = in_toto_run("hello-sigstore", [], [], ["echo", "hello", "sigstore"], signer=signer)

# Hack to convert a `Key` to the in-toto -flavoured dict representation  
key_dict = key.to_dict()
key_dict["keyid"] = key.keyid

link.verify_signature(key_dict)
```

in_toto/formats.py
lukpueh: I'd consider that a breaking change. If we do that I'd rather get rid of `FormatError` altogether and use built-in `ValueError` instead.

adityasaky: fair enough!

adityasaky: We should start tracking some "we'd like to bundle this small thing in the next time we make a breaking change release" tasks somewhere.

Issue #694, opened 2024-01-31
Please add a timeout option to in-toto-verify


skinny-b:
**Description of issue or feature request**:
Requesting the option of having `--run-timeout` added to `in-toto-verify`. Within the inspect it makes use of run which has a hardcoded 10 second timeout. Doing something like having `grype` scan a made sbom can timeout
Layout:
```
...
      "inspect": [{
          "name": "sbom_scan",
          "expected_materials": [ ... ],
          "expected_products": [ ... ],
          "run": [
             "grype", "spdx.bom.json", "-f", "low"
          ]
        }
...
```

**Current behavior**:
```
$ in-toto-verify -l ./the.layout --layout-keys ./owner.pub
'-k', '--layout-keys' is deprecated, use '--verification-keys' instead.
(in-toto-verify) TimeoutExpired: Command '['grype', 'spdx.bom.json', '-f', 'low']' timed out after 9.999979632906616 seconds
```

This command updates a database before running a scan so it will take longer than 10 seconds (most of the time)

**Expected behavior**:
Command will run and either pass(return code 0)/fail(return code 1) based upon the inspection result.

adityasaky:
I think adding this is fine. @skinny-b do you want to take a stab at a PR? We already have an arg for in-toto-run that controls this.

lukpueh:
Agreed, this is definitely worth adding.

skinny-b:
I'd like to take a crack at it.

lukpueh:
Added in #695

PR #695, opened 2024-02-02
Adding argument to in-toto-verify for --run-timeout, to allow inspect…

skinny-b:
…ion a passed in timeout. #694

that is provided, the better.

**Fixes issue #694 **:

requirements**:

### DIFF
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -223,6 +223,20 @@ RUN_TIMEOUT_KWARGS = {
     ),
 }
 
+INSPECT_TIMEOUT_ARGS = ["--inspection-timeout"]
+INSPECT_TIMEOUT_KWARGS = {
+    "type": int,
+    "dest": "inspect_timeout",
+    "default": LINK_CMD_EXEC_TIMEOUT,
+    "help": (
+        "integer that represents the max timeout in seconds for the "
+        "   in-toto-verify command for inspect subprocess."
+        "   Default is '{timeout}' seconds.".format(
+            timeout=LINK_CMD_EXEC_TIMEOUT
+        )
+    ),
+}
+
 
 def title_case_action_groups(parser):
     """Capitalize the first character of all words in the title of each action
--- a/in_toto/common_args.py
+++ b/in_toto/common_args.py
@@ -223,20 +223,6 @@ RUN_TIMEOUT_KWARGS = {
     ),
 }
 
-INSPECT_TIMEOUT_ARGS = ["--inspection-timeout"]
-INSPECT_TIMEOUT_KWARGS = {
-    "type": int,
-    "dest": "inspect_timeout",
-    "default": LINK_CMD_EXEC_TIMEOUT,
-    "help": (
-        "integer that represents the max timeout in seconds for the "
-        "   in-toto-verify command for inspect subprocess."
-        "   Default is '{timeout}' seconds.".format(
-            timeout=LINK_CMD_EXEC_TIMEOUT
-        )
-    ),
-}
-
 
 def title_case_action_groups(parser):
     """Capitalize the first character of all words in the title of each action
--- a/in_toto/in_toto_verify.py
+++ b/in_toto/in_toto_verify.py
@@ -43,11 +43,11 @@ from in_toto import (
 from in_toto.common_args import (
     GPG_HOME_ARGS,
     GPG_HOME_KWARGS,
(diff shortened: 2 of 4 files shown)


skinny-b:
`tox` passed for myself. Built the wheel locally and used it in the same scenario with I had filed issue #694 for.

```
$ in-toto-verify -l ./the.layout --layout-keys ./owner.pub --run-timeout 12
'-k', '--layout-keys' is deprecated, use '--verification-keys' instead.
(in-toto-verify) TimeoutExpired: Command '['grype', 'spdx.bom.json', '-f', 'low']' timed out after 11.999943041 seconds
```

With a higher timeout (my case should fail):
```
$ in-toto-verify -l ./the.layout --layout-keys ./owner.pub --run-timeout 600
'-k', '--layout-keys' is deprecated, use '--verification-keys' instead.
(in-toto-verify) BadReturnValueError: Got non-zero return value '1' from command '['grype', 'spdx.bom.json', '-f', 'low']'.
```

skinny-b:
Marking this with tests added, though I'm not adding but made a fix as the functionality is covered under run and this is just passing an argument through it. `TestLinkCmdExecTimeout` covers the main feature.

in_toto/in_toto_verify.py
```diff
@@ -217,6 +219,7 @@ def create_parser():
     )
 
     parser.add_argument(*GPG_HOME_ARGS, **GPG_HOME_KWARGS)
+    parser.add_argument(*RUN_TIMEOUT_ARGS, **RUN_TIMEOUT_KWARGS)
```
lukpueh: minor nit: the help message says
> integer that represents the max timeout in seconds for the for the **in-toto-run command**

This isn't accurate in this context. An easy fix would be to say something like

> step/inspection subprocess timeout in seconds.

A more user friendly approach would be to not re-use the common arg, but create a new one with a dedicated help message. I'm fine either way. Can you update?

lukpueh (changes requested):
Thanks for the clean patch. Can you update the cli help message?  Otherwise this looks great.

in_toto/in_toto_verify.py
skinny-b: Yes I didn't think about the help when I checked. Would you recommend I change the argument so it's not `--run-timeout` as well to just `--timeout` or `--verify-timeout`? It would be more in depth change but my thought was the familiarity with the run command.

lukpueh: Yeah, using a custom argument is probably the right thing to do, and it doesn't make the change really any bigger. I think I'd call it `--inspection-timeout`, because it's the timeout for a single inspection subprocess (and not for the entire in-toto-verify command).

in_toto/common_args.py
```diff
@@ -223,6 +223,20 @@ def parse_password_and_prompt_args(args):
...
+    "default": LINK_CMD_EXEC_TIMEOUT,
+    "help": (
+        "integer that represents the max timeout in seconds for the "
+        "   in-toto-verify command for inspect subprocess."
+        "   Default is '{timeout}' seconds.".format(
+            timeout=LINK_CMD_EXEC_TIMEOUT
+        )
+    ),
+}
+
```
lukpueh: IMO you can move this directly to `in_toto_verify.py`, it doesn't look like it will be re-used by another command (same is true for RUN_TIMEOUT, but we don't have to fix that here).

in_toto/verifylib.py
```diff
@@ -1487,6 +1506,7 @@ def in_toto_verify(
     substitution_parameters=None,
     step_name="",
     persist_inspection_links=True,
+    run_timeout=in_toto.settings.LINK_CMD_EXEC_TIMEOUT,
```
lukpueh: I suggest to update the name here too (and in the docstring).

in_toto/verifylib.py
```diff
@@ -1329,7 +1341,9 @@ def reduce_chain_links(chain_link_dict):
     return reduced_chain_link_dict
 
 
-def verify_sublayouts(layout, steps_metadata, superlayout_link_dir_path):
+def verify_sublayouts(
+    layout, steps_metadata, superlayout_link_dir_path, run_timeout
```
lukpueh: I suggest to update the name here too (and in the docstring).

lukpueh (changes requested):

lukpueh (approved):
LGTM. Thanks!

PR #701, opened 2024-02-14
Dependabot: use groups

lukpueh:
Group weekly dependency updates to reduce maintenance burden.

Groups are:
  * critical python build/release deps
  * python test and lint deps (only pinned for test repro)
  * all other python dependencies
  * all github action dependencies

This was shamelessly copied from @jku's
https://github.com/theupdateframework/python-tuf/pull/2530. Cheers!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,7 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
-  allow:
-  - dependency-type: direct
-  - dependency-type: indirect
+
 - package-ecosystem: "github-actions"
   directory: "/"
   schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,31 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
-
+  groups:
+    build-and-release-dependencies:
+      # Python dependencies known to be critical to our build/release security
+      patterns:
+        - "build"
+        - "hatchling"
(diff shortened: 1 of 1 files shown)


lukpueh:
@adityasaky, would you kindly approve?

trishankatdatadog (approved):

adityasaky:
image
@trishankatdatadog @lukpueh was this an update feature branch using rebase click on the web UI?

.github/dependabot.yml
```diff
@@ -5,11 +5,31 @@ updates:
...
+  groups:
+    build-and-release-dependencies:
+      # Python dependencies known to be critical to our build/release security
+      patterns:
+        - "build"
+        - "hatchling"
+    test-and-lint-dependencies:
+      # Python dependencies that are only pinned to ensure test reproducibility
+      patterns:
+        - "bandit"
```
adityasaky, line 17: Is this the best way to do this? I haven't seen this feature before but can we do it by updated path perhaps? That way we don't have to keep this list updated if the corresponding requirements.txt files change...

jku: it's entirely possible (maybe by configuring multiple `package-ecosystem: pip`s, each one pointing to a different subdir with different requirements files) but I figured that it would look so strange (and dependabot is so vaguely documented) that I couldn't be bothered.

trishankatdatadog:
> @trishankatdatadog @lukpueh was this an update feature branch using rebase click on the web UI?

yes, and it was me

adityasaky (approved):

PR #727, opened 2024-04-03
Revert "Adopt mandatory Signer.public_key property"

lukpueh:
Reverts in-toto/in-toto#724

I merged this prematurely. Let me revert and re-open as draft.


lukpueh:
The reason to not include #724 in the next in-toto release is that it will also pin securesystemslib to its current minor version, that is before Signer got the public_key property, so the removed public_key property check in in-toto should probably remain (in reality this doesn't matter much, but still).

PR #728, opened 2024-04-03
Adopt mandatory Signer.public_key property

lukpueh:
* blocks on next securesystemslib release

----

secure-systems-lab/securesystemslib#756 added a `public_key` property to the Signer interface.

This PR updates in-toto accordingly:

* Add dummy public_key property to GPG pseudo-signer GPGSigner only implements parts of the Signer interface for compatibility-reasons. This is okay, because it is only used internally.

* Remove obsolete public_key property check and related test in runlib.


lukpueh:
superseded by #740

PR #739, opened 2024-04-30
Remove deprecated CLI/API arguments and legacy securesystemslib integration

### LINKED ISSUE #697: Drop legacy securesystemslib integration
Remove legacy securesystemslib integration and related in-toto interfaces deprecated in v2.2.0 (see [changelog](https://github.com/in-toto/in-toto/blob/v2.2.0/CHANGELOG.md#v220) for details).

Related work exists in https://github.com/in-toto/in-toto/compare/develop...lukpueh:in-toto:rm-legacy-keys2

lukpueh:
Fixes #697

This PR removes deprecated CLI and API arguments related to signature creation and verification. See deprecation and replacement note [v2.2.0 release notes](https://github.com/in-toto/in-toto/blob/v2.2.0/CHANGELOG.md#v220) for details.

Docs and tests are updated accordingly.

I suggest to review commit by commit, and maybe skim test changes (coverage remains the same.

Nice side-effects are:
- legacy securesystemslib integration is removed
- `pynacl` optional dependency is removed
- test speed increases significantly (~10s instead of ~40s)


lukpueh:
test failures in sslib main are expected an will be addressed in a separate PR

doc/source/api.rst
```diff
@@ -6,6 +6,19 @@ can use to generate, consume, modify and verify in-toto metadata, as a more
 feature-rich, programmable alternative to the :doc:`command line tools
 <command-line-tools/index>`.
 
+.. note::
+
+   **Cryptographic Signatures**
+
+   in-toto metadata is signed with cryptographic keys using the in-house crypto
```
adityasaky: nit: maybe we should drop the term "in-house" given in-toto is at the CNCF and securesystemslib is not, I'm not sure how different people may parse it.

doc/source/api.rst
```diff
@@ -6,6 +6,19 @@ can use to generate, consume, modify and verify in-toto metadata, as a more
 feature-rich, programmable alternative to the :doc:`command line tools
 <command-line-tools/index>`.
 
+.. note::
```
adityasaky, line 9: General question: is this expected to bump up the major version? I think it is, right, since we're dropping params from the API too?

tests/test_in_toto_verify.py
```diff
@@ -572,6 +156,37 @@ def test_main_multiple_keys(self):
...
+            "--layout",
+            "demo.layout",
+            "--link-dir",
+            "bad-link-dir",
+            "--verification-keys",
+        ] + self.public_key_paths
+        self.assert_cli_sys_exit(args, 1)
+
 
 class TestInTotoVerifySubjectPublicKeyInfoKeysAndUseDSSE(
```
adityasaky, line 191: Given our "defaults" have effectively changed, maybe this test could be renamed. I was a little thrown by the removal of `TestInTotoVerifyTool` above. But also, it may not be an issue once this PR lands and it's a lot clearer what the tests are doing in the file...

adityasaky (commented):
I didn't take the most detailed pass at this, but this looks good to me, minus some questions about versioning. If this is going to be a "breaking change" release (edit: I see from the last release notes that we call out a major release), maybe we can preserve some of the CLI arg names. --layout-keys, to me, is a lot clearer than --verification-keys, for example.

doc/source/api.rst
lukpueh: Yes, absolutely, this will require a major version bump.

tests/test_in_toto_verify.py
lukpueh: 🤷 I don't like the name either. But I don't mind it enough to rename.

lukpueh:
> If this is going to be a "breaking change" release, maybe we can preserve some of the CLI arg names. --layout-keys, to me, is a lot clearer than --verification-keys, for example.

Yes, we could. Though, it would surely upset users, who "migrated" from `--layout-keys` to `--verification-keys`, after the last release. Now the'll have to half-migrate back.

IMO having a new different successor arg name, instead of repurposing the existing one, is a more transparent API break.

adityasaky:
Seems reasonable! I'll take another pass later today, sorry for the multiple go-s at this. :)

doc/source/api.rst
lukpueh: Fixed in  9ce441a

in_toto/common_args.py
```diff
@@ -169,10 +135,9 @@ def parse_password_and_prompt_args(args):
     "metavar": "<path>",
     "dest": "signing_key",
     "help": (
-        "replacement for '--key' using a standard PKCS8/PEM format. Key type is"
-        " detected automatically and need not be specified with '--key-type'."
-        " Use '--password [<password>]' to pass a decryption password"
-        " or toggle a prompt, if the key es encrypted."
+        "signing key in a standard PKCS8/PEM format. Supported keytypes are"
+        " rsa, ed25519, ecdsa (nistp256). Use '--password [<password>]' to pass"
+        " a decryption password or toggle a prompt, if the key es encrypted."
```
adityasaky:
```suggestion
        " a decryption password or toggle a prompt, if the key is encrypted."
```

adityasaky (approved):

lukpueh:
> test failures in sslib main are expected an will be addressed in a separate PR

this is still true. merging...

Issue #414, opened 2020-11-16
Improve documentation of parameter substitution


adityasaky:
Please fill in the fields below to submit an issue or feature request. The more information that is provided, the better.

**Description of issue or feature request**:

in-toto's reference implementation provides a mechanism to substitute values in artifact rules (in steps and inspections), the expected command field in steps, and the "run" field in inspections. The documentation for this feature needs to be improved - currently, it's only documented [here](https://in-toto.readthedocs.io/en/latest/api.html#in_toto.verifylib.in_toto_verify). The improved documentation should ideally also include some examples of parameter substitution, perhaps as they're deployed in production.

**Current behavior**:

The feature isn't documented in detail.

**Expected behavior**:

Improved documentation, examples to explain use-cases.

lukpueh:
Thanks for documenting this request, @adityasaky.

I suggest we add a new "Usage Examples" top-level section to [in-toto.readthedocs.io](https://in-toto.readthedocs.io/en/latest/) to which we move the already existing ["Layout Creation Example" page](https://in-toto.readthedocs.io/en/latest/layout-creation-example.html) plus more usage examples, one of which could show how run `in_toto_verify` with substitution parameters, maybe using production in-toto metadata from Datadog. What do you think, @trishankatdatadog?

MashiatK:
hi, could i work on the issue?

adityasaky:
Go ahead!

PR #748, opened 2024-05-15
Added Usage Examples as new subtoctree

DarikshaAnsari:
**Fixes issue #414 **:

Added Examples section: Introduced a new section titled "Useage Examples" to showcase practical use cases of in-toto, including parameter substitution and layout creation.

Parameter Substitution: Expanded the subsection on parameter substitution to provide detailed examples of its usage in steps and inspections.

Layout Creation Example: Added a new subsection demonstrating how to create a supply chain layout using in-toto's model classes and convenience methods.


DarikshaAnsari:
@adityasaky @lukpueh PTAL !!

lukpueh:
Cool, thanks! And btw, here's a preview of the RTD build: https://in-toto--748.org.readthedocs.build/en/748/examples.html#

lukpueh (changes requested):
Wow, this looks very nice! I have a few high-level comments:

- Please  reverse the order of "layout creation example" and "parameter substitution". IMO the basic example should come before the more specific example.
- The "Example Usage" under "Using Substitution in Inspections" actually shows Steps. This needs to be changed.
- I think it would be useful to mention all fields in the layout that support substitution (see `verifylib.substitute_parameters`). We don't need to show an example for each. 
- It might be helpful to point out how substitution parameters can be passed to the [verification API](https://in-toto.readthedocs.io/en/latest/api.html#in_toto.verifylib.in_toto_verify)

lukpueh:
> * It might be helpful to point out how substitution parameters can be passed to the [verification API](https://in-toto.readthedocs.io/en/latest/api.html#in_toto.verifylib.in_toto_verify)

ping

lukpueh:
Did you want to finish this PR, @DarikshaAnsari?

DarikshaAnsari:
Sorry @lukpueh I am working on umbrella issues right now I will look at this later.

PR #835, opened 2025-03-11
Apply assorted ruff rules

DimitriPapadopoulos:

Apply ruff rules that seem to improve readability or maintainability. That's debatable of course, which is why I have pushed a commit per rule, so that rules can be disabled.

I did not try to enforce ruff rules, since pylint is currently the linter of choice.

requirements**:

### DIFF
--- a/tests/test_runlib.py
+++ b/tests/test_runlib.py
@@ -547,7 +547,7 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
                 ["foo", "bar", "#esc!", "subdir/foosub1", "subdir/foosub2"],
             ),
             (
-                ["\#esc*"],  # pylint: disable=W1401
+                ["\\#esc*"],
                 [
                     "foo",
                     "bar",
@@ -557,7 +557,7 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
                 ],
             ),
             (
-                ["*esc\!"],  # pylint: disable=W1401
+                ["*esc\\!"],
                 [
                     "foo",
                     "bar",
--- a/tests/test_runlib.py
+++ b/tests/test_runlib.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# coding=utf-8
 
 # Copyright New York University and the in-toto contributors
 # SPDX-License-Identifier: Apache-2.0
--- a/tests/test_runlib.py
+++ b/tests/test_runlib.py
@@ -333,7 +333,7 @@ class TestRecordArtifactsAsDict(unittest.TestCase, TmpDirMixin):
         """If the platform is Windows, raises an error that asks the user if
         developer mode is activated."""
         if os.name == "nt":
-            raise IOError(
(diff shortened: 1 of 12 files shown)


tests/test_runlib.py
```diff
@@ -571,7 +570,7 @@ def test_exclude_patterns(self):
                 ],
             ),
             (
-                ["*esc\!"],  # pylint: disable=W1401
+                ["*esc\\!"],
```
lukpueh, line 559: The anomalous backslash may be intentional. 🤷 

-> https://github.com/in-toto/in-toto/pull/487#discussion_r809668288

lukpueh: @rszamszur, @adityasaky, do you remember the your motivations from 3 years ago? 😅

lukpueh (commented):
Thanks for the patch! Please make sure the current linter still passes and we can merge this.

Or, if you want to switch to ruff altogether, I'd be also supportive.

tests/test_runlib.py
DimitriPapadopoulos: The intent/motivation is irrelevant, this is not a valid escape sequence in Python. In the absence of a valid escape sequence, Python interprets `\` as `\\`, see [2.4.1.1. Escape sequences](https://docs.python.org/3/reference/lexical_analysis.html#escape-sequences):
> Unlike Standard C, all unrecognized escape sequences are left in the string unchanged, i.e., _the backslash is left in the result_. (This behavior is useful when debugging: if an escape sequence is mistyped, the resulting output is more easily recognized as broken.)

The alternative is raw strings:
```suggestion
                [r"*esc\!"],
```

lukpueh: Okay, thanks for checking. Let's leave it as you proposed then. The change doesn't seem to break the tests.

DimitriPapadopoulos:
I can switch to ruff in a different PR.

I will have to select relevant [Pylint](https://docs.astral.sh/ruff/rules/#pylint-pl) rules in ruff (`PLC`, `PLE`, `PLR`, `PLW`) and enforce them. I'm not sure there's a one-to-one mapping between current Pylint rules and their counterparts in ruff.

lukpueh:
> I can switch to ruff in a different PR.

Thanks! 🙏 

> 
> I will have to select relevant [Pylint](https://docs.astral.sh/ruff/rules/#pylint-pl) rules in ruff (`PLC`, `PLE`, `PLR`, `PLW`) and enforce them. I'm not sure there's a one-to-one mapping between current Pylint rules and their counterparts in ruff.

It doesn't have to be one-to-one. In in-toto's sister project tuf, we started the switch with just a small set of ruff rules and then incrementally enabled more, see https://github.com/theupdateframework/python-tuf/issues/2550 and https://github.com/theupdateframework/python-tuf/issues/2567.

DimitriPapadopoulos:
Please bear with me while I run black to fix the CI errors.

DimitriPapadopoulos:
Ready for merging.

lukpueh (approved):

PR #845, opened 2025-03-14
Enforce ruff/flake8-datetimez rules (DTZ)

DimitriPapadopoulos:

```
DTZ002 `datetime.datetime.today()` used
```

Not sure about the fix: `datetime.today()` → `datetime.now(tz=datetime.timezone.utc)`. Do we want naive "local time"? Should we change to UTC, or perhaps better local time with time zone?

requirements**:

### DIFF
--- a/in_toto/models/layout.py
+++ b/in_toto/models/layout.py
@@ -171,7 +171,7 @@ class Layout(Signable):
         _check_int(years)
 
         self.expires = (
-            datetime.today()
+            datetime.today()  # noqa: DTZ002
             + relativedelta(days=days, months=months, years=years)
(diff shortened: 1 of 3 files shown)


lukpueh:
> Not sure about the fix: datetime.today() → datetime.now(tz=datetime.timezone.utc). Do we want naive "local time"? Should we change to UTC, or perhaps better local time with time zone?

I lean towards not changing API behaviour.

DimitriPapadopoulos:
OK. So keep tha naive local time and add `noqa`.

lukpueh (approved):

PR #861, opened 2025-03-19
Port securesystemslib.hash file hashing

lukpueh:
securesystemslib.hash is a small wrapper around hashlib, which serves two main purposes:
* provide helper function to hash a file, while optionally normalizing file endings.
* translate custom hash algorithm name "blake2b-256" to "blake2b" with (digest_size=32).

In preparation for the removal of securesystemslib.hash, this patch ports the former only (file hashing with normalization). The latter was never used in in-toto.

related https://github.com/secure-systems-lab/securesystemslib/issues/943

### DIFF
--- a/in_toto/resolver/_resolver.py
+++ b/in_toto/resolver/_resolver.py
@@ -1,6 +1,7 @@
 """Resolver interface and implementations for files, OSTree, and directory
 artifacts."""
 
+import hashlib
 import locale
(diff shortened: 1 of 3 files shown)


lukpueh:
ping @adityasaky, @SantiagoTorres

could we get a review here?

adityasaky (approved):

Issue #866, opened 2025-03-21
Return code not passed through with run


skinny-b:
When adding `in-toto-run` to a pipeline if the command fails, it's important to fail the step from a pipeline perspective. An easy way to see this is that `in-toto-run` with wrapping the command and absorbing the return code returns 0 when the command can fail. Easy to produce this:
```
$ grep -c NotHere README.md
0
$ echo $?
1
$ in-toto-run -n test --signing-key gitlab-private.pem -- grep -c NotHere README.md
$ echo $?
0
$ in-toto-run --version
in-toto-run 3.0.0
```

if it matter, I'm running on MacOS 15.3.2 (24D81), but I'm seeing similar with a gitlab pipeline that the runner is using the `python:3.12` container image.

adilalperenciftci:
This appears to be fixed on the current `develop` branch by commit 0b3d78e8 (`in-toto#866 adding in Metablock passthrough for "return-code" to in_toto_run`).

Current `in_toto/in_toto_run.py` exits with the recorded `byproducts["return-value"]` after metadata generation, and `tests/test_in_toto_run.py::TestInTotoRunToolWithDSSE::test_main_failed_command` covers a wrapped command returning 2 while asserting that link metadata still exists.

I did not open a PR because the reported behavior is already implemented. The issue may be closable if no additional compatibility behavior is expected.

I could not execute the focused test locally because this checkout's test environment is not installed (`pytest` is unavailable), so this note is based on the current implementation and committed regression test rather than a fresh runtime result.

skinny-b:
@adilalperenciftci I believe you are correct. When I modified the code I validated it within the same system that I had reported the behavior on, though I saw it on a few systems.  I think this was left open by mistake and I'll mark closed.

PR #868, opened 2025-03-27
#866: Add command failure return-value to `in-toto-run` exit

skinny-b:
**Fixes issue #866 **:

Adding checks to verify in_toto_run has a link returned and pull the "return-value" from byproducts to exit with that code. Some commands will fail and currently it is hidden when wrapped by "in-toto-run". This change allows the failed command return code to exit with that code and still produce the link file.

The reason I opted for the link file to be generated, there are cases when a pipeline might be ok with failures but it would still be necessary to validate that a step was executed.

requirements**:

The doc would be now correct as the command states:
```
$ in-toto-run --help
usage: in-toto-run <named arguments> [optional arguments] \
	 -- <command> [args]
...
stdout and stderr) to a link metadata file, which is signed with the passed
key. It returns a non-zero value on failure and zero otherwise.
```

### DIFF
--- a/in_toto/in_toto_run.py
+++ b/in_toto/in_toto_run.py
@@ -317,7 +317,7 @@ def main():
         link
         and getattr(link, "signed", None)
         and link.signed.byproducts.get("return-value")
-        and link.signed.byproducts.get("return-value") != 0
     ):
         sys.exit(link.signed.byproducts.get("return-value"))
+
     sys.exit(0)
--- a/in_toto/in_toto_run.py
+++ b/in_toto/in_toto_run.py
@@ -275,6 +275,7 @@ def main():
             " Please specify (or use the --no-command option)"
         )
 
+    link = None
     try:
         # We load the key here because it might prompt the user for a password in
         # case the key is encrypted. Something that should not happen in the lib.
@@ -290,7 +291,7 @@ def main():
                 args.signing_key, password
             )
 
-        runlib.in_toto_run(
+        link = runlib.in_toto_run(
             args.step_name,
             args.materials,
             args.products,
@@ -312,4 +313,11 @@ def main():
         LOG.error("(in-toto-run) %s: %s", type(e).__name__, e)
         sys.exit(1)
 
+    if (
+        link
+        and getattr(link, "signed", None)
+        and link.signed.byproducts.get("return-value")
+        and link.signed.byproducts.get("return-value") != 0
+    ):
+        sys.exit(link.signed.byproducts.get("return-value"))
     sys.exit(0)
--- a/in_toto/in_toto_run.py
+++ b/in_toto/in_toto_run.py
@@ -275,7 +275,7 @@ def main():
             " Please specify (or use the --no-command option)"
         )
 
-    link = None
+    payload = None
     try:
         # We load the key here because it might prompt the user for a password in
         # case the key is encrypted. Something that should not happen in the lib.
@@ -308,16 +308,18 @@ def main():
             timeout=args.run_timeout,
             signer=signer,
         )
-
+        # FIXME the link has just been serialized and the payload
+        # is being deserialized in order to get the 'return-value'.
+        payload = link.get_payload()
     except Exception as e:  # noqa: BLE001
         LOG.error("(in-toto-run) %s: %s", type(e).__name__, e)
         sys.exit(1)
 
     if (
-        link
-        and getattr(link, "signed", None)
-        and link.signed.byproducts.get("return-value")
+        payload
+        and payload.type_ == "link"
+        and payload.byproducts.get("return-value")
     ):
-        sys.exit(link.signed.byproducts.get("return-value"))
+        sys.exit(payload.byproducts.get("return-value"))
 
     sys.exit(0)
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
 
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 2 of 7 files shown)


skinny-b:
Manually testing this:
```
$ in-toto-run -n test --signing-key gitlab-private.pem -- grep -c NotHere README.md
$ echo $?
1
$ grep "return-value" test.*link
   "return-value": 1,
```
Tox also passed for me
```
$ tox
...
py311: commands[1]> coverage report -m --fail-under 99
Name                                Stmts   Miss Branch BrPart  Cover   Missing
in_toto/__init__.py                     2      0      0      0   100%
in_toto/common_args.py                 45      0      8      0   100%
in_toto/exceptions.py                  10      0      0      0   100%
in_toto/formats.py                     70      0     24      0   100%
in_toto/in_toto_match_products.py      36      8     12      2    67%   82-90, 95
in_toto/in_toto_mock.py                29      0      0      0   100%
in_toto/in_toto_record.py              65      0     14      0   100%
in_toto/in_toto_run.py                 69      0     16      0   100%
in_toto/in_toto_sign.py               114      0     36      0   100%
in_toto/in_toto_verify.py              51      2      8      1    95%   214-215
in_toto/log.py                         26      0      4      0   100%
in_toto/models/__init__.py              0      0      0      0   100%
in_toto/models/_signer.py              81      0     10      1    99%   255->257
in_toto/models/common.py               16      0      4      0   100%
in_toto/models/layout.py              221      2     58      0    99%   206, 255
in_toto/models/link.py                 55      0     16      0   100%
in_toto/models/metadata.py            135      2     32      2    98%   58->64, 216, 295
in_toto/resolver/__init__.py            1      0      0      0   100%
in_toto/resolver/_resolver.py         169      1     70      0    99%   40
in_toto/rulelib.py                     79      1     30      1    98%   203
in_toto/runlib.py                     254      1    108      1    99%   992
in_toto/settings.py                     4      0      0      0   100%
in_toto/verifylib.py                  337      0    132      0   100%
TOTAL                                1869     17    582      8    99%
  lint: OK (5.15=setup[4.67]+cmd[0.46,0.02] seconds)
  py38: SKIP (0.67 seconds)
  py39: OK (14.24=setup[3.78]+cmd[10.24,0.22] seconds)
  py310: OK (14.12=setup[4.15]+cmd[9.77,0.19] seconds)
  py311: OK (14.06=setup[4.11]+cmd[9.77,0.18] seconds)
  congratulations :) (48.28 seconds)
```

in_toto/in_toto_run.py
```diff
@@ -312,4 +313,11 @@ def main():
         LOG.error("(in-toto-run) %s: %s", type(e).__name__, e)
         sys.exit(1)
 
+    if (
+        link
+        and getattr(link, "signed", None)
```
lukpueh: I think, if in_toto_run didn't not return a Metadata object, which has a signed field that is a Link object (which has a byproducts field that is an opaque dictionary) then it's a bug. So we shouldn't need to check for this here.

in_toto/in_toto_run.py
```diff
@@ -312,4 +313,11 @@ def main():
...
+        and link.signed.byproducts.get("return-value")
+        and link.signed.byproducts.get("return-value") != 0
```
lukpueh: Is there a reason, you're not just returning this 0 but the hardcoded one below?

lukpueh (commented):
Thanks for the clean PR! ❤️  I think it's a great addition. Any objections to merge this? cc @SantiagoTorres, @adityasaky? 

ps: strictly speaking this probably requires a major version bump

in_toto/in_toto_run.py
skinny-b: I added this check because `test_pkcs8_signing_key` failed without it:
```
  File ".../tests/test_in_toto_run.py", line 473, in test_pkcs8_signing_key
    self.assert_cli_sys_exit(args + [str(pem_path)], 0)
  File ".../tests/common.py", line 147, in assert_cli_sys_exit
    self.cli_main_func()  # pylint: disable=not-callable
    ^^^^^^^^^^^^^^^^^^^^
  File ".../in_toto/in_toto_run.py", line 316, in main
    if link.signed.byproducts.get("return-value"):
       ^^^^^^^^^^^
AttributeError: 'Envelope' object has no attribute 'signed'
```

in_toto/in_toto_run.py
skinny-b: I'll remove that, no reason.

in_toto/in_toto_run.py
lukpueh: Ah right, forgot about DSSE Envelopes. We need to handle those too. You can call `Envelope.get_payload` to get the contained Link. This is a bit silly though, because it JSON-deserializes the Link, which was serialized just before in in_toto_run. 

An alternative would be to change in_toto_run, e.g. to return the Link (or its byproducts, or just the return value) in addition to the Metadata object. Although changing this return value is probably a lot more disruptive to in-toto users than changing just the exit code of the cli command.

Another alternative would be to include the Link object in the Envelope object, along with its serialized data. But if that gets used elsewhere, it runs the risk of getting out of sync.

Given the options, I think, I prefer the first solution. Maybe together with a big FIXME comment (or GitHub issue), that points out the wasted deserialization work. What do others think?

skinny-b: I'll get to work on the `Envelope.get_payload`, I need to get some time to free up in order to work on it.

adityasaky:
> Thanks for the clean PR! ❤️ I think it's a great addition. Any objections to merge this? cc @SantiagoTorres, @adityasaky?

None from me, sorry for the slow response!

lukpueh (approved):
Great work! Thanks

PR #897, opened 2025-11-05
Refactor code to make it easier to read

immanuwell:
@lukpueh @DimitriPapadopoulos fyi

1. changed one difficult to read if-condition to the condition with the same effect, but much easier to read
2. changed one string with .format() to f-string

There were no new features added by my 2 commits, just refactoring

requirements**:

- [x] Tests have been added for the bug fix or new feature (there were no new features added by my 2 commits, just refactoring)
- [x] Docs have been added for the bug fix or new feature (there were no new features added by my 2 commits, just refactoring)

### DIFF
--- a/in_toto/in_toto_record.py
+++ b/in_toto/in_toto_record.py
@@ -228,7 +228,7 @@ def main():
     LOG.setLevelVerboseOrQuiet(args.verbose, args.quiet)
 
     # Use exactly one of gpg or pkcs8 signing key
-    if sum([bool(args.gpg), bool(args.signing_key)]) != 1:
+    if bool(args.gpg) == bool(args.signing_key):
         parser.print_usage()
         parser.error(
             "Specify either '--signing-key <path>' or '--gpg [<keyid>]'"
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
Much better indeed!

in_toto/in_toto_record.py
```diff
@@ -228,7 +228,7 @@ def main():
     LOG.setLevelVerboseOrQuiet(args.verbose, args.quiet)
 
     # Use exactly one of gpg or pkcs8 signing key
-    if sum([bool(args.gpg), bool(args.signing_key)]) != 1:
+    if bool(args.gpg) == bool(args.signing_key):
```
DimitriPapadopoulos, line 231: It's not entirely straightforward to understand what this is about:
* We don't want `(False, False)` (none of the options used) or `(True, True)` (both of the options used).
* We want exactly one of the options: `(True, False)` or `(False, True)`.

I still find the new code more readable, especially with the comment on top of it.

JustinCappos: There are probably even cleaner ways to write this that are more verbose.  I would tend to favor explicitly checking for False, False (and erroring) and True, True (and erroring separately).  However, I agree this is an improvement.

